PR #1, opened 2019-07-09
Feature/add config

sslaws:
Add mapped environment config to WebClient as a basis for future apps


Apps/WebClient/src/WebClient.csproj
```diff
@@ -7,7 +7,6 @@
     <IsPackable>false</IsPackable>
 
     <!-- SonarQube needs this -->
-    <ProjectGuid>{4EF62115-3012-4911-B7AA-3B798231ADC4}</ProjectGuid>
```
FuriousLlama: Sonarqube will not analize it without this

Apps/WebClient/src/Properties/launchSettings.json
```diff
@@ -0,0 +1,16 @@
+{
```
FuriousLlama: Looks like this file is only needed by the Visual Studio. Not sure if we should keep it in the repo.

PR #3, opened 2019-07-10
Brad

bradhead:
have run sonarqube.  Adds localisation framework and localized immz page as well as lang toggle.


FuriousLlama:
SonarQube report https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=webclientbrad

Apps/WebClient/src/ClientApp/i18n.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (matched && matched.length > 1) {
+
+        const locale = matched[1];
+        messages[locale] = locales(key);
+      }
+    })
+    return messages;
+  }
+
+  export default new VueI18n({
```
FuriousLlama: We can probably make this a class

bradhead: future sprint we should yes.

PR #7, opened 2019-07-11
Remove sonar secret

sslaws:


bradhead:
good idea to remove secret!

PR #8, opened 2019-07-11
Brad

bradhead:
got immunization table sorting and row alternatiing colours working along with localization of the header


bradhead:
cancelled pull request

PR #11, opened 2019-07-12
Added task and script for functional tests

TiagoGraf89:


TiagoGraf89:
Stephen can you review this?
To be honest I don't think it is going to work but I don't have permissions to change the build yaml without a PR, is there a way i could change it on the fly just for test purposes?

sslaws:
I’ll take a look at the PR, right now GitHub is setup for PRs on any change.  I’ll see if I can configure it on paths.

PR #16, opened 2019-07-13
Added logoff vue

TiagoGraf89:
Work on AB#7141


Apps/WebClient/src/ClientApp/components/logout/logout.vue.html
```diff
@@ -0,0 +1,14 @@
+﻿<template>
+    <div class="row justify-content-center h-100 pt-5">
+        <div class="col-lg-6 col-md-6 pt-2">
+            <div class="shadow-lg p-3 mb-5 bg-white rounded border">
+                <h3>
+                    <strong>You signed out of your account</strong>
```
FuriousLlama: These need to be i18n

bradhead: Setup log off section in locales/en.json

PR #20, opened 2019-07-15
7168

TiagoGraf89:
Added unit tests for auth controller
Refactored AuthController into AuthService
Added unit tests for AuthService


TiagoGraf89:
https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=7168-WebClient

Apps/WebClient/src/Controllers/AuthController.cs
```diff
@@ -1,57 +1,60 @@
...
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using HealthGateway.WebClient.Services;
+
+namespace HealthGateway.WebClient.Controllers
+{
+    /// <summary>
+    /// The Authentication and Authorization controller.
+    /// </summary>
```
FuriousLlama: Kudos for the documentation. Maybe we should add this to our style for all public classes and methods

Apps/WebClient/src/Controllers/AuthController.cs
```diff
@@ -1,57 +1,60 @@
...
-        public bool Auth { get; set; }
-        public string Token { get; set; }
-        public string User { get; set; }
-
-        public LogonResult()
-        {
-        }
-    }
-}
+using System.Threading.Tasks;
```
FuriousLlama: Sonar is reporting a major issue with includes like this (which I use as well):
https://sonarqube-gateway.pathfinder.gov.bc.ca/component_measures?id=7168-WebClient&metric=coverage&selected=7168-WebClient%3Asrc%2FControllers%2FAuthController.cs&view=list

Perhaps we should start doing the using declarations inside the namespace to avoid conflicts between namespaces

https://stackoverflow.com/questions/125319/should-using-directives-be-inside-or-outside-the-namespace

PR #30, opened 2019-07-19
Feature AB#7130

FuriousLlama:
Added tests for the error pages. Added snowplow to the header at the… layout level

[AB#7130](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/7130)

SonarQ: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=7130-webclient


Apps/WebClient/src/ClientApp/test/navbar.header.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    localVue.use(BootstrapVue)
+
+    const wrapper = shallowMount(HeaderComponent, { localVue });
+    expect(wrapper.isVueInstance()).toBeTruthy()
+  })
+})
+
+describe('HeaderComponent', () => {
+  test('updates displayed language', () => {
+    // TODO: the following is not working. Figure out how to call a method at wrapper.vm
```
TiagoGraf89: I'll take a look into this

Apps/WebClient/src/Views/Shared/_Layout.cshtml
```diff
@@ -10,6 +10,29 @@
...
         <link rel="stylesheet" href="~/dist/site.css" asp-append-version="true" />
     </environment>
+
+    <!-- Snowplow starts plowing - Standalone vA.2.10.2 -->
+    <script type="text/javascript">
+        ;(function(p,l,o,w,i,n,g){if(!p[i]){p.GlobalSnowplowNamespace=p.GlobalSnowplowNamespace||[];
+        p.GlobalSnowplowNamespace.push(i);p[i]=function(){(p[i].q=p[i].q||[]).push(arguments)
+        };p[i].q=p[i].q||[];n=l.createElement(o);g=l.getElementsByTagName(o)[0];n.async=1;
+        n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,"script","https://sp-js.apps.gov.bc.ca/MDWay3UqFnIiGVLIo7aoMi4xMC4y.js","snowplow"));
+        var collector = 'spm.gov.bc.ca';
```
TiagoGraf89: I think I remember something about making this configurable, did we decide to keep it static?

Apps/WebClient/src/ClientApp/test/error.test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+        expect(wrapper.find("h1").text()).toBe(error.code)
+        expect(wrapper.find("h2").text()).toBe(error.name)
+        expect(wrapper.find("p").text()).toBe(error.message)
+    })
+})
+
+describe('NotFoundView', () => {
+    test('renders properties correctly', () => {
+        const mountWrapper = mount(NotFoundView);
```
TiagoGraf89: it doesn't matter for this smaller tests but ideally the wrapper should be declared/initiated outside this test but inside the describe, so we can reuse it through more than 1 test, you could even break each expect into one specific test, eg: 

describe('ErrorComponent', () => {
  let error: PageError = new PageError('777', 'ERROR_NAME', 'ERROR_MESSAGE')
  const wrapper = shallowMount(ErrorComponent, { propsData: { error } });

  test('renders header 1 correctly', () => {
    expect(wrapper.find("h1").text()).toBe(error.code)
  })

  test('renders header 2 correctly', () => {
    expect(wrapper.find("h2").text()).toBe(error.name)
  })

  test('renders paragraph correctly', () => {
    expect(wrapper.find("p").text()).toBe(error.name)
  })
})

FuriousLlama: That looks way cleaner! Do you know if any state declared on the described will be cleaned after each test?

Apps/WebClient/src/Views/Shared/_Layout.cshtml
FuriousLlama: for now. We can instroduce a razor variable once we get there. But until we can check if we are actually hitting anything I dont think its necessary

PR #32, opened 2019-07-22
Add Unit Tests AB#7127

TiagoGraf89:


Apps/WebClient/src/ClientApp/test/home.test.ts
```diff
@@ -1,10 +1,16 @@
...
+  const wrapper = shallowMount(HomeComponent, { localVue });
+
   test('is a Vue instance', () => {
-    const localVue = createLocalVue()
-    const wrapper = shallowMount(HomeComponent, { localVue })
-    expect(wrapper.isVueInstance()).toBeTruthy()
+    expect(wrapper.isVueInstance()).toBeTruthy();
+  })
+
+  const expectedH1Text = "Health Gateway";
```
FuriousLlama: nit: having the expect value inside the test itself unless is reused somewhere else

PR #37, opened 2019-07-23
Features 7240

sslaws:
Initial work on [AB#7240](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/7240)
new API call at /api/GetAuthenticationData which will return an AuthData object
Refactor the authentication code to remove a lot of code and use config
Remove old imms client code.


Apps/WebClient/src/Server/Services/AuthService.cs
```diff
@@ -39,12 +39,21 @@ public AuthService(ILogger<AuthService> logger, IHttpContextAccessor httpContext
...
-            this.logger.LogTrace("Getting Bearer token");
-            var user = this.httpContextAccessor.HttpContext.User.FindFirst(ClaimTypes.NameIdentifier).Value;
-            var token = await this.httpContextAccessor.HttpContext.GetTokenAsync("access_token").ConfigureAwait(true);
-            return new Models.AuthData { IsAuthenticated = true, Token = token, User = user };
+            Models.AuthData authData = new Models.AuthData();
+            ClaimsPrincipal user = this.httpContextAccessor.HttpContext.User;
+            authData.IsAuthenticated = user.Identity.IsAuthenticated;
+            if (authData.IsAuthenticated)
+            {
+                this.logger.LogDebug("Getting Authentication data");
```
FuriousLlama: We could instead use the constructor for User and pass that one into the constructor of AuthData.
```
User user = new User { 
       Id= user.FindFirstValue("preferred_username"), 
       Name = user.FindFirstValue("name"),
       Email = user.FindFirstValue(ClaimTypes.Email)
}
Models.AuthData authData = new Models.AuthData(user, user.Identity.IsAuthenticated);
```

That way the AuthData object is guaranteed to have a User if it exists, otherwise will null

Apps/WebClient/src/Server/Startup.cs
```diff
@@ -15,84 +17,54 @@ namespace HealthGateway.WebClient
...
-                                    this.configuration["OIDC:IDPHintKey"], ctx.Properties.Items[this.configuration["OIDC:IDPHintKey"]]);
+                                    this.configuration["KeyCloak:IDPHintKey"], ctx.Properties.Items[this.configuration["KeyCloak:IDPHintKey"]]);
                          }
                          else
-                            if (!string.IsNullOrEmpty(this.configuration["OIDC:IDPHint"]))
-                            {
-                                 this.logger.LogDebug("Adding IDP Hint on Redirect to provider");
-                                 ctx.ProtocolMessage.SetParameter(this.configuration["OIDC:IDPHintKey"], this.configuration["OIDC:IDPHint"]);
-                            }
+                            if (!string.IsNullOrEmpty(this.configuration["KeyCloak:IDPHint"]))
```
FuriousLlama: nit, this looks odd

Apps/WebClient/src/Server/Services/AuthService.cs
sslaws: resolved

Apps/WebClient/src/Server/Startup.cs
sslaws: fixed

PR #41, opened 2019-07-24
added login choice and vuex store

bradhead:
https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=WebClient-7238


Apps/WebClient/src/ClientApp/app/router.ts
```diff
@@ -1,63 +1,107 @@
...
+            {
+                path: '/immunizations',
+                component: ImmunizationsComponent,
+                meta: { requiresAuth: true, roles: ['user'] }
+            }
         ]
     },
+    {
+        path: '/auth',
+        component: TempAuthComponent,
```
FuriousLlama: Is this one still needed?

Apps/WebClient/src/ClientApp/app/router.ts
```diff
@@ -1,63 +1,107 @@
...
+router.beforeEach(async (to, from, next) => {
     if (to.meta.requiresAuth) {
-        //const auth = store.state.security.auth
-        const auth = { authenticated: false };//store.state.security.auth
+        let isAuthenticated = store.getters.isAuthenticated;
         console.log('Requires auth!');
-        if (!auth.authenticated) {
+        if (!isAuthenticated) {
             //security.init(next, to.meta.roles)
             next({ name: 'unauthenticated' })
```
FuriousLlama: if not authenticated should skip this one

Apps/WebClient/src/ClientApp/app/store/modules/auth/2auth.js
```diff
@@ -0,0 +1,73 @@
+import axios from 'axios';
```
FuriousLlama: not sure 2auth.js is needed anymore.

Apps/WebClient/src/ClientApp/app/views/auth.vue
```diff
@@ -0,0 +1,36 @@
+<template>
```
FuriousLlama: Looks like this file is for testing only. We can keep it to make sure the authentication is working propertly.

Apps/WebClient/src/ClientApp/app/store/modules/auth/mutations.ts
```diff
@@ -0,0 +1,31 @@
+import { MutationTree } from 'vuex';
+import { AuthState, Authentication } from '@/models/authState';
+import { stat } from 'fs';
+
+export const mutations: MutationTree<AuthState> = {
+    authenticationRequest(state: AuthState) {
+        state.error = false;
+        state.statusMessage = 'loading';
+        state.requestedRoute = '';
```
FuriousLlama: Right now the authentication request will set the requestedRoute to empty. Instead what if we pass the requested route as a parameter to the authenticationRequest mutation?
```
authenticationRequest(state: AuthState, requestedRpite: string) {
      ...
}
```

Apps/WebClient/src/ClientApp/app/store/modules/auth/mutations.ts
```diff
@@ -0,0 +1,31 @@
...
+        state.authentication = payload;
+        state.statusMessage = 'success';
+        state.requestedRoute = '';
+    },
+    authenticationError(state: AuthState, errorMessage: string) {
+        state.error = true;
+        state.authentication = undefined;
+        state.statusMessage = errorMessage;
+    },
+    logout(state: AuthState) {
```
FuriousLlama: Looks like the mutations work as a state diagram for the AuthState. What if we also add logoutRequested, logoutComplete as mutations that set the entire state as the logout flow happens?

Apps/WebClient/src/ClientApp/app/router.ts
```diff
@@ -1,63 +1,107 @@
...
             console.log('Not authenticated');
+            console.log(to.path);
+            // commit the to.path to vuex here
+            store.commit('requestedRoute', to.path);
```
FuriousLlama: Idea, instead of doing a state commit here, what if we pass the to.path as a query string to the Login Component. That way we avoid changing the state until the components stats the login flow

Apps/WebClient/src/ClientApp/app/router.ts
bradhead: wasn't Stephen still using it for debugging?

Apps/WebClient/src/ClientApp/app/store/modules/auth/mutations.ts
bradhead: yes, we could do that in one step for sure.

Apps/WebClient/src/ClientApp/app/router.ts
bradhead: let's do this later.. I'll get a successfully PR build going now wit the fixed msising npm package

PR #49, opened 2019-07-27
Add Swagger documentation

sslaws:
Task: [AB#7241](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/7241)

Adds versioned swagger documentation at /swagger
Fixed most warnings in classes updated as part of the work


Apps/WebClient/src/Server/Startup.cs
```diff
@@ -1,25 +1,49 @@
...
     using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Logging;
+    using Microsoft.Extensions.Options;
+    using Swashbuckle.AspNetCore.SwaggerGen;
+    using Swashbuckle.AspNetCore.SwaggerUI;
 
+    /// <summary>
+    /// Configures the application during startup.
+    /// </summary>
     public class Startup
```
FuriousLlama: SonarQ its flagging some lines on this method:
https://sonarqube-gateway.pathfinder.gov.bc.ca/project/issues?id=webclient_dev&open=AWwmE1CeZJzGytDdbsgL&resolved=false&severities=CRITICAL&types=CODE_SMELL

Are those addressable now?

sslaws: I don't think we can refactor this yet but once we look at doing auth in Vue most of it should go away.

PR #56, opened 2019-07-31
Fixing nits for login functional testing AB#7147

TiagoGraf89:
Add id to menu login link
Fixed warnings in login.vue (using b-row and b-col instead of tr/td)


TiagoGraf89:
Fixed the SiteMinder cached login by expiring the SMSESSION cookie

bradhead:
Two minor things.. set cookie to expire in the past by a few seconds.. i.e. set to now less a a bit, or alternatively, delete the cookie by creating a blank SMSESSION cookie... might be more robust, but let's leave as is for now.
as in... 
eraseCookie('SMSESSION');

function eraseCookie(c_name) {
  createCookie(cookie_name,"",-1);
}

PR #138, opened 2019-10-10
Patient service authorization AB#7521

FuriousLlama:
Added authorization to the Patient service. 
The medication service now passes a authorization header when contacting the patient server


sslaws:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=patientservice_pr_138

healthgatewaybc:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=webclient_pr_138

Apps/Medication/src/Controllers/MedicationController.cs
```diff
@@ -81,7 +81,8 @@ public MedicationController(IMedicationService medicationService, IHttpContextAc
         [Authorize]
         public async Task<List<MedicationStatement>> GetMedications(string hdid)
         {
-            string phn = await this.patientService.GetPatientPHNAsync(hdid).ConfigureAwait(true);
+            string jwtString = this.Request.Headers["Authorization"][0];
```
TiagoGraf89: can you use this.Request.Header.Authorization instead of this? would rather use the strongly typed property for it but not sure if its accessible

FuriousLlama: The problem with that one is that it requires a Class that takes two parameters, Bearer and Token. If we do that we have to split the incoming authorization token before being passed and re build with the same parameters here. IMO unnecessary

sslaws:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_138

sslaws:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=patientservice_pr_138

healthgatewaybc:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=webclient_pr_138

PR #168, opened 2019-10-25
Build files for Database library

FuriousLlama:
Added Azure build files. Added stylecop exclusion to the migration files


healthgatewaybc:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_168

Apps/Database/src/Migrations/20191024233502_InitialCreate.cs
```diff
@@ -457,3 +460,7 @@ protected override void Down(MigrationBuilder migrationBuilder)
         }
     }
 }
+
+
+#pragma warning restore SA1200 // Using directive should appear within a namespace declaration
```
bradhead: will we ever need to regen these initial create files and so if so will we lose these code-smell-avoiding pragmas?

FuriousLlama: Correct. Fortunately it looks like we only need to add // <auto-generated/> to the first line of the files

healthgatewaybc:
SonarQube Report: https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_168

PR #201, opened 2019-11-06
Add UserProfile controller and service

TiagoGraf89:


Apps/WebClient/src/Server/Controllers/UserProfileController.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// </summary>
+        /// <returns>The http status.</returns>
+        /// <param name="userProfile">The user profile model.</param>
+        /// <response code="200">The user profile record was saved.</response>
+        /// <response code="400">The user profile was already inserted.</response>
+        /// <response code="401">The client must authenticate itself to get the requested response.</response>
+        /// <response code="403">The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401, the client's identity is known to the server.</response>
+        [HttpPost]
+        [Authorize(Policy = "PatientOnly")]
+        public async Task<IActionResult> InsertUserProfile([FromBody] UserProfile userProfile)
```
FuriousLlama: We should use PUT instead of POST since this URL will be inserting the record instead of updating.
What would be the url for this call? I believe it should be something like v1/.../UserProfile/{HDID}

TiagoGraf89: I think thats the other way around, PUT means update and POST means insert "Use PUT when you want to modify a singular resource which is already a part of resources collection. PUT replaces the resource in its entirety. Use PATCH if request updates part of the resource. Use POST when you want to add a child resource under resources collection."

TiagoGraf89: The URL is v1/api/UserProfile/
There's no need to pass the hdid since we already pass it in the body

sslaws: Take a look at this: https://restfulapi.net/rest-put-vs-post/  POST for create.

TiagoGraf89: Added the hdid to the endpoint, url should be v1/api/UserProfile/{hdid}

Apps/WebClient/src/Server/Models/UserProfile.cs
```diff
@@ -0,0 +1,38 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------
+namespace HealthGateway.WebClient.Models
+{
+    /// <summary>
+    /// The user profile model.
+    /// </summary>
+    public class UserProfile
```
sslaws: We should move this to database models as its ultimately going to be stored in the DB.

TiagoGraf89: Moved to database

healthgatewaybc:
Database Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=databaselibrary_pr_201)

healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_201)

healthgatewaybc:
WebClient Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=webclient_pr_201)

PR #231, opened 2019-11-22
Features/7713

FuriousLlama:
Initial work for the registration status
[AB#7713](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/7713)

### DIFF
--- a/Apps/WebClient/src/ClientApp/app/boot.ts
+++ b/Apps/WebClient/src/ClientApp/app/boot.ts
@@ -33,7 +33,6 @@ import SERVICE_IDENTIFIER, {
 import container from "@/inversify.config";
 import { ExternalConfiguration } from "@/models/configData";
 import User from "@/models/user";
-import { RegistrationStatus } from "./constants/registrationStatus";
 
 Vue.use(BootstrapVue);
 Vue.use(VueRouter);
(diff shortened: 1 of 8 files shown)


healthgatewaybc:
WebClient Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=webclient_pr_231)

Tools/Dev/Postgres/docker-compose.yml
```diff
@@ -10,4 +10,7 @@ services:
     restart: unless-stopped
     volumes:
       - ./backups:/backups
-      - ./pgdata:/var/lib/postgresql/data
```
sslaws: Why did this get changed?  If you want the external folder to be mapped into data then we can make that change but the internal path should remain at /var/lib/postgresql otherwise we won't be externalizing the data.

FuriousLlama: Without this change postgress wont start due to permission issues.

Apps/WebClient/src/ClientApp/app/boot.ts
```diff
@@ -33,6 +33,7 @@ import SERVICE_IDENTIFIER, {
 import container from "@/inversify.config";
 import { ExternalConfiguration } from "@/models/configData";
 import User from "@/models/user";
+import { RegistrationStatus } from "./constants/registrationStatus";
```
TiagoGraf89: @ instead of .

FuriousLlama: fixed

PR #234, opened 2019-11-26
Features/7665b

bradhead:
protects Hangfire and sets default path to root of the service. Yet to do is to check for user in role, doing that today.


Apps/JobScheduler/src/Authorization/AuthorizationDashboardFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Authorization hook for Hangfire.
+        /// </summary>
+        /// <param name="context">An instance of <see cref="DashboardContext"/> class.</param>
+        /// <returns>true, if authorized.</returns>
+        public bool Authorize(DashboardContext context)
+        {
+            var httpContext = context.GetHttpContext();
+            if (!httpContext.User.Identity.IsAuthenticated)
+            {
+                httpContext.Response.Redirect("/login");
```
sslaws: This should be configurable

Apps/JobScheduler/src/Startup.cs
```diff
@@ -130,7 +149,58 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 },
             });
 
-            app.UseMvc();
+            // This is needed if running behind a reverse proxy
+            // like ngrok which is great for testing while developing
+            app.UseForwardedHeaders(new ForwardedHeadersOptions
+            {
+                RequireHeaderSymmetry = false,
```
sslaws: We have this in the Common startup... might need to add the header symmetry

Apps/JobScheduler/src/Startup.cs
```diff
@@ -130,7 +149,58 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
+                auth.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                auth.DefaultChallengeScheme = OpenIdConnectDefaults.AuthenticationScheme;
+            })
+            .AddCookie(options =>
+            {
+                options.Cookie.Name = "HealthGateway_JobScheduler";
+            })
+            .AddOpenIdConnect(options =>
+            {
+                options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
sslaws: You can simplify this logic by binding to configuration and making all of the configurable with defaults.              
  this.configuration.GetSection("OpenIdConnect").Bind(options);

Apps/JobScheduler/src/Properties/launchSettings.json
```diff
@@ -4,7 +4,7 @@
     "windowsAuthentication": false,
     "anonymousAuthentication": true,
     "iisExpress": {
-      "applicationUrl": "http://localhost:5005"
+      "applicationUrl": "https://localhost:5005"
```
FuriousLlama: I dont think this will work on localhost since https requires a certificate

Apps/JobScheduler/src/Startup.cs
FuriousLlama: +1

Apps/JobScheduler/src/Authorization/AuthorizationDashboardFilter.cs
bradhead: why do we want that? it is just a hook to get to a challenge response so we can prompt the user to authenticate. i should rename it authChallenge, or something as that is what is doing. If we configured it i would have to also configure the name of the route for the handling method in the home controller.

Apps/JobScheduler/src/Startup.cs
bradhead: i will look at what we did in common.

Apps/JobScheduler/src/Startup.cs
bradhead: yes, i know, but i hat debugging the magical binder. i like to literally see what is being set. so it looks deliberate.  Bind() requires the user to go on a goose chase to see what defaults are set, etc... i am happy to use the Bind() but just not a fan of it.

Apps/JobScheduler/src/Properties/launchSettings.json
bradhead: i had setup   local cert when debuggin issues on Friday, it hinted about cookies not exhanged over HTTPS but in the end it was the order of setting up cors and auth for Hangfire that caused the grief. I have since set it back

Apps/JobScheduler/src/Authorization/AuthorizationDashboardFilter.cs
sslaws: Routes follow general conventions in how the routing works.  this is just a random string placed in a Filter.  If anything, I'd rather see a public const in the controller so that anyone changing the controller route would realize that a filter depends on it.

Apps/JobScheduler/src/Startup.cs
sslaws: I'm fine although we can chat when I'm back :)  I find one break point and I can confirm all of the bindings.

Apps/JobScheduler/src/Properties/launchSettings.json
```diff
@@ -11,7 +11,7 @@
     "Healthgateway.JobScheduler": {
       "commandName": "Project",
       "launchBrowser": true,
-      "launchUrl": "hangfire",
```
sslaws: Why was this changed?

bradhead: hangfire is now the root route for the site.

PR #239, opened 2019-11-27
Add edits for protective word AB#7728

sslaws:
Add edits
Add an ignore for certs so that we don't check them in
job scheduler change was made by Brad and already exists


healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_239)

PR #282, opened 2019-12-19
Fix hot reloading

TiagoGraf89:


Apps/WebClient/src/Server/Startup.cs
```diff
@@ -145,10 +146,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseSpa(spa =>
             {
                 spa.Options.SourcePath = "dist";
-
                 if (env.IsDevelopment())
-                {
-                    spa.UseProxyToSpaDevelopmentServer("http://localhost:5000");
+                {
+                    spa.UseReactDevelopmentServer("dev");
```
FuriousLlama: Is this tied to the package.json line "dev"?

TiagoGraf89: Yep!

PR #403, opened 2020-03-28
Update DBProfileDelegate.cs

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8108](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8108)
- [ ] Enhancement
- [x] Bug
- [ ] Documentation

## Description:
Fixes the count of logged in users today in admin module.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required

### Steps to Reproduce:
Login to admin module.
Logged in users today always show 0.

### UI Changes
NO

### DIFF
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -158,10 +158,10 @@ namespace HealthGateway.Database.Delegates
         }
 
         /// <inheritdoc />
-        public int GetTodayLoggedinUsersCount()
+        public int GetLoggedInUsersCount(DateTime startDate)
         {
             int result = this.dbContext.UserProfile
-                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= DateTime.UtcNow.Date);
+                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= startDate);
             return result;
         }
     }
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -162,7 +162,8 @@ namespace HealthGateway.Database.Delegates
         {
             DateTime startDate = TimeZoneInfo.ConvertTimeFromUtc(DateTime.UtcNow.Date, timezone);
             int result = this.dbContext.UserProfile
-                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= startDate);
+                .Count(u => u.LastLoginDateTime.HasValue &&
+                TimeZoneInfo.ConvertTimeFromUtc(u.LastLoginDateTime.Value, timezone) >= startDate);
             return result;
         }
     }
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -161,7 +161,7 @@ namespace HealthGateway.Database.Delegates
         public int GetLoggedInUsersCount()
         {
             int result = this.dbContext.UserProfile
-                .Count(u => u.LastLoginDateTime >= DateTime.Today.ToUniversalTime());
+                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= DateTime.UtcNow.Date);
             return result;
         }
     }
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -158,8 +158,9 @@ namespace HealthGateway.Database.Delegates
         }
 
         /// <inheritdoc />
-        public int GetLoggedInUsersCount(DateTime startDate)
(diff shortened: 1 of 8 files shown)


healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_403)

Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -161,7 +161,7 @@ public int GeUnregisteredInvitedUsersCount()
         public int GetLoggedInUsersCount()
```
FuriousLlama: Name is a tad misleading. How about something like GetTodayLogedinUserCount

Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -161,7 +161,7 @@ public int GeUnregisteredInvitedUsersCount()
...
         {
             int result = this.dbContext.UserProfile
-                .Count(u => u.LastLoginDateTime >= DateTime.Today.ToUniversalTime());
+                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= DateTime.UtcNow.Date);
```
FuriousLlama: Why greater or equal? Also, the ufc offset could count users from the previous night as logged in today and the ones in the afternoon as counted for tomorrow.

tiagograf-AOT: If the last login is recorded as utc there shouldnt be an issue, query should be:

tiagograf-AOT: LastLoginDateTime (2020-03-27 13:00) >= Current Date (2020-03-27 00:00) both converted to utc

healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_403)

Apps/Database/src/Delegates/DBProfileDelegate.cs
FuriousLlama: On UTC the comparison is correct. However the admin that needs to see if it is on the same day on its own time zone.

healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_403)

Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -158,10 +158,10 @@ public int GeUnregisteredInvitedUsersCount()
         }
 
         /// <inheritdoc />
-        public int GetTodayLoggedinUsersCount()
+        public int GetLoggedInUsersCount(DateTime startDate)
         {
             int result = this.dbContext.UserProfile
-                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= DateTime.UtcNow.Date);
+                .Count(u => u.LastLoginDateTime.HasValue && u.LastLoginDateTime.Value >= startDate);
```
FuriousLlama: Not sure this solves the problem, Instead how about converting both dates to a passed timeZone. That way the calculation can still be made and know for sure that the date will match

tiagograf-AOT: Then the database delegate would be handling date logic which i think its wrong. 
I'm pretty sure this should work though, the utc today is converted to vancouver today, and only utc records logged in after that will make to the count. using the scenario we talked about today would be 2020-03-30 08:00:00 utc

FuriousLlama: The issue is that the delegate needs to do the query, which needs to account for the time zone. To make the delegate agnostic we can pass the timezone as a paramater

healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_403)

Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -158,10 +158,12 @@ public int GeUnregisteredInvitedUsersCount()
         }
 
         /// <inheritdoc />
-        public int GetLoggedInUsersCount()
+        public int GetTodayLoggedInUsersCount(TimeZoneInfo timezone)
         {
+            DateTime startDate = TimeZoneInfo.ConvertTimeFromUtc(DateTime.UtcNow.Date, timezone);
```
FuriousLlama: nit: startDate =? today

healthgatewaybc:
Medication Project [SonarQube Report](https://sonarqube-gateway.pathfinder.gov.bc.ca/dashboard?id=Medicationservice_pr_403)

Apps/Database/src/Delegates/DBProfileDelegate.cs
bradhead: nit2: startDate => today => todaysDate

PR #416, opened 2020-04-03
Set up CI with Azure Pipelines

tiagograf-AOT:
Creates a new build for all projects which requires sonarqube analysis. (PR Created from azure, so no template was applied) [AB#8144](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8144)

### DIFF
--- a/Tools/SonarQube/azure/build.yaml
+++ b/Tools/SonarQube/azure/build.yaml
@@ -1,3 +1,10 @@
+schedules:
+- cron: "0 0 * * *"
+  displayName: Daily midnight build
+  branches:
+    include:
+    - dev
+
 pool:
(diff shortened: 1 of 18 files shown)


Tools/SonarQube/azure/build.yaml
```diff
@@ -0,0 +1,203 @@
+pool:
+  name: 'HealthGateway'
+
```
sslaws, line 3: We need to have this scheduled as per https://docs.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml

PR #418, opened 2020-04-08
Adds the email list page to admin module

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8126](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8126)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Adds a new page to admin module that display emails with the ability to resend.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required


Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
```diff
@@ -0,0 +1,62 @@
...
+  private http!: IHttpDelegate;
+
+  public initialize(http: IHttpDelegate): void {
+    this.http = http;
+  }
+
+  public getEmails(): Promise<Email[]> {
+    return new Promise((resolve, reject) => {
+      this.http
+        .get<RequestResult<Email[]>>(`${this.BETA_REQUEST_BASE_URI}`)
```
sslaws, line 20: NIT: This probably shouldn't be BETA_REQUET_BASE_URI

Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+  }
+
+  public resendEmails(emailIds: string[]): Promise<string[]> {
+    return new Promise((resolve, reject) => {
+      let headers: Dictionary<string> = {};
+      headers["Content-Type"] = "application/json; charset=utf-8";
+      this.http
+        .patch<RequestResult<string[]>>(
+          `${this.BETA_REQUEST_BASE_URI}`,
```
sslaws, line 37: Same comment as above

Apps/AdminWebClient/src/Server/Controllers/EmailAdminController.cs
```diff
@@ -79,21 +80,28 @@ public IActionResult GetEmails()
...
-        /// <param name="emailId">The email Id to resend.</param>
-        /// <response code="200">Returns the list of beta requests.</response>
+        /// <param name="emailIds">The email Ids to resend.</param>
+        /// <response code="200">Successfully queued all emails.</response>
         /// <response code="401">the client must authenticate itself to get the requested response.</response>
         /// <response code="403">The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401, the client's identity is known to the server.</response>
-        [HttpPost]
-        [Route("{emailId}")]
-        public IActionResult ResendEmail(Guid emailId)
+        [HttpPatch]
```
sslaws, line 87: I don't think this is appropriate as we are creating a net new resource and not updating something.

Apps/AdminWebClient/src/Server/Controllers/EmailAdminController.cs
```diff
@@ -79,21 +80,28 @@ public IActionResult GetEmails()
...
+        public IActionResult ResendEmail(List<Guid> emailIds)
```
sslaws, line 88: I guess this means that you allow multiple selection on the front end?

Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
```diff
@@ -0,0 +1,62 @@
+import { injectable } from "inversify";
+import { IHttpDelegate, IEmailAdminService } from "@/services/interfaces";
+import { Dictionary } from "vue-router/types/router";
+import { ResultType } from "@/constants/resulttype";
+import RequestResult from "@/models/requestResult";
+import Email from '@/models/email';
```
FuriousLlama, line 6: double quotes instead of single. Running **npm run lintfix** will fix this

Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
```diff
@@ -0,0 +1,62 @@
...
+
+@injectable()
+export class RestEmailAdminService implements IEmailAdminService {
+  private readonly BETA_REQUEST_BASE_URI: string = "v1/api/EmailAdmin";
```
FuriousLlama, line 10: Not sure if this should be EmailAdmin. This looks like the resource is "The email of the Admin". An alternative would be v1/Admin/api/Email. Thoughts?

Apps/AdminWebClient/src/Server/Controllers/EmailAdminController.cs
FuriousLlama: +1

Apps/AdminWebClient/src/Server/Controllers/EmailAdminController.cs
```diff
@@ -79,21 +80,28 @@ public IActionResult GetEmails()
         /// Creates a clone of the email id and requeues it for sending.
```
FuriousLlama, line 80: nit: docs dont match multiple email functionality

Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
tiagograf-AOT: just following on stephen's namings, I thought just Email would make more sense but idk.

Apps/AdminWebClient/src/Server/Controllers/EmailAdminController.cs
```diff
@@ -76,24 +69,31 @@ public IActionResult GetEmails()
...
         /// <response code="401">the client must authenticate itself to get the requested response.</response>
         /// <response code="403">The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401, the client's identity is known to the server.</response>
         [HttpPost]
-        [Route("{emailId}")]
-        public IActionResult ResendEmail(Guid emailId)
+        public IActionResult ResendEmail(List<Guid> emailIds)
         {
+            if (emailIds == null)
+            {
+                throw new ArgumentNullException(nameof(emailIds));
```
sslaws, line 84: Instead of throwing an exception we should probably return a 400 or 418 HTTP Error.

Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
```diff
@@ -30,13 +30,10 @@ export class RestEmailAdminService implements IEmailAdminService {
 
   public resendEmails(emailIds: string[]): Promise<string[]> {
     return new Promise((resolve, reject) => {
-      let headers: Dictionary<string> = {};
-      headers["Content-Type"] = "application/json; charset=utf-8";
```
FuriousLlama: Without this it was having issues reaching the server with the correct format. Did you test that it does not break it?

Apps/AdminWebClient/src/ClientApp/src/views/ResendEmail.vue
```diff
@@ -16,9 +23,10 @@
               :items="emailList"
               :items-per-page="5"
               show-select
+              :search="filterText"
             >
               <template v-slot:item.sentDateTime="{ item }">
-                <span>{{ formatDate(item.sentDateTime) }}</span>
+                  <span>{{ item.sentDateTime ? formatDate(item.sentDateTime) : "" }}</span>
```
FuriousLlama, line 29: Instead of having logic on the ui, lets make the formatDate more robust

Apps/AdminWebClient/src/Server/Services/EmailAdminService.cs
```diff
@@ -69,5 +88,15 @@ public RequestResult<IEnumerable<Email>> GetEmails()
             };
             return result;
         }
+
+        private string GetEmailInviteStatus(EmailInvite ei)
+        {
+            if (ei == null)
```
FuriousLlama, line 94: ei => emailInvite or invite

Apps/AdminWebClient/src/Server/Services/EmailAdminService.cs
```diff
@@ -29,13 +31,17 @@ namespace HealthGateway.Admin.Services
     /// </summary>
     public class EmailAdminService : IEmailAdminService
     {
+        private const string StatusNotInvited = "Not Invited";
```
FuriousLlama, line 34: Looks like these could be enums or put in a different file

Apps/AdminWebClient/src/ClientApp/src/services/restEmailAdminService.ts
```diff
@@ -30,13 +30,16 @@ export class RestEmailAdminService implements IEmailAdminService {
...
+      headers["Content-Type"] = "application/json; charset=utf-8";
       this.http
         .post<RequestResult<string[]>>(
           `${this.BASE_URI}`,
-          JSON.stringify(emailIds)
+          JSON.stringify(emailIds),
+          headers
         )
         .then(requestResult => {
-          this.handleResult(requestResult, resolve, reject);
```
FuriousLlama: looks odd, this function does not return anything?

tiagograf-AOT: It does not, result is based on the request status (200, 400 etc)

PR #424, opened 2020-04-14
Fetch email for associated feedback AB#8139 and update DB for AB#8137

sslaws:
## Status
Ready

## Fixes or Implements [AB#8139](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8139), [AB#8137](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8137)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Fetches the email along with the user feedback for the admin module.  The UserFeedback had to be updated to include the HDID (UserProfileId) and I made an associated change to the WebClient to record the HDID on creation of the feedback.

We should probably be mapping our objects in the DB models as it is needed when querying linked tables and doing it after may have to alter our table structure.

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### Steps to Reproduce:
N/A

### UI Changes
No

### Browsers Tested
- [ ] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

Testing will be done in a stand alone ticket.

## Notes/Additional Comments
Any addtional notes or comments that may be relevant.  Include any specialized deployment steps here.


Apps/Database/src/Models/UserFeedbackAdmin.cs
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+//-------------------------------------------------------------------------
+namespace HealthGateway.Database.Models
+{
+    using System.ComponentModel.DataAnnotations;
+
+    /// <summary>
+    /// An administrative view of UserFeedback with Email populated from UserProfile.
+    /// </summary>
+    public class UserFeedbackAdmin : UserFeedback
```
FuriousLlama, line 23: Will this not create a new table?

Apps/Database/src/Models/UserFeedback.cs
```diff
@@ -32,5 +32,10 @@ public class UserFeedback : AuditableEntity
 
         [MaxLength(500)]
         public string? Comment { get; set; }
+
+        [MaxLength(52)]
+        public string? UserProfileId { get; set; }
```
FuriousLlama, line 37: The isSatisfied field is no longer needed.

tiagograf-AOT: +1

Apps/Database/src/Models/UserFeedbackAdmin.cs
sslaws: No, to create a new table it would need to be added to the DB Context and even then it would create an alter to the UserFeedback table and add a discriminator as its a child object.

Apps/AdminWebClient/src/Server/Models/UserFeedbackView.cs
```diff
@@ -54,6 +54,15 @@ public class UserFeedbackView
         /// </summary>
         public uint Version { get; set; } = 0;
 
+        /// <summary>
```
FuriousLlama, line 57: I donth think the **satisfied** field anymore.

Apps/Database/src/Models/UserFeedbackAdmin.cs
FuriousLlama, line 23: Would it not be better to just modify the UserFeedback model instead?

Apps/Database/src/Models/UserFeedback.cs
sslaws: Correct but it exists in the DB and I think that any cleanup of the DB should be done after the code is delivered.  Otherwise, you can't really rollback.

Apps/Database/src/Models/UserFeedbackAdmin.cs
sslaws: No, that would alter the table definition which we don't want.  We query both UserProfile and UserFeedback and merge them into a new object model.

FuriousLlama: Given that the user feedback is tied to a user profile, we can just have the the UserFeedback model hold that link as well.

PR #435, opened 2020-04-21
Add Comment Service, Controller, Delegate, Model, Test

tiagograf-AOT:
## Status
**In Development**

## Implements [AB#8179](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8179)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Implements Comment backend.
Migration scripts to be created once model is approved by this PR.

## Testing
- [x] Unit Tests Updated
- [ ] Not Required

### UI Changes
NO

### DIFF
--- a/Apps/Database/src/Models/Comment.cs
+++ b/Apps/Database/src/Models/Comment.cs
@@ -38,7 +38,7 @@ namespace HealthGateway.Database.Models
         [Column("UserProfileId")]
         [MaxLength(52)]
         [Required]
-        public string HdId { get; set; } = null!;
+        public string UserProfileId { get; set; } = null!;
 
         /// <summary>
         /// Gets or sets the text of the note.
@@ -60,5 +60,10 @@ namespace HealthGateway.Database.Models
         [MaxLength(32)]
         [Required]
         public string ParentEntryId { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Gets or sets the linked user profile.
+        /// </summary>
+        public virtual UserProfile? UserProfile { get; set; }
     }
 }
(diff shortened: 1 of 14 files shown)


Apps/Database/src/Models/Comment.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </summary>
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        [Column("CommentId")]
+        public Guid Id { get; set; }
+
+        /// <summary>
+        /// Gets or sets the user hdid.
+        /// </summary>
+        [Column("UserProfileId")]
```
sslaws, line 38: I think we should be creating relationships in our models for the DB or at the very least in the DB Context.  Doing this will automatically link the tables but does require that we use the proper column name:

        [MaxLength(52)]
        public string? UserProfileId { get; set; }

        public virtual UserProfile? UserProfile { get; set; }

I think that going forward we shouldn't use HDID in our code as it is an external ID and instead it should be locally referenced as UserProfileID.

Apps/WebClient/src/Server/Controllers/CommentController.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <summary>
+        /// Gets all comments for the authorized user and event id.
+        /// </summary>
+        /// <param name="parentEntryId">The parent entry id.</param>
+        /// <returns>The list of comments wrapped in a request result.</returns>
+        /// <response code="200">Returns the list of comments.</response>
+        /// <response code="401">The client must authenticate itself to get the requested response.</response>
+        /// <response code="403">The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401, the client's identity is known to the server.</response>
+        [HttpGet]
+        [Route("{parentEntryId}")]
```
FuriousLlama, line 169: I am not 100% this is restful `Comment/{parent_id}` since it might look similar to the comment id. How about `Comment/?parentId={parent_id}

PR #436, opened 2020-04-21
Hotfix for bug involving accessing length of undefined

sam-warren:
## Status
**Ready**

## Fixes or Implements AB#8326
- [ ] Enhancement
- [X] Bug
- [ ] Documentation

## Description:
Tried to access index of empty array < 0 without a check. This led to an attempt to access the length of undefined which led to an endless spinning wheel for all users who had no timeline entries. Fixed by implementing a quick check if the user had no timeline entries.

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### Steps to Reproduce:
Run the user timeline and uncheck all filters so you see no entries on the page. This should render an empty page rather than throwing an error.

### UI Changes
NO

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

### DIFF
--- a/Apps/WebClient/src/ClientApp/app/views/timeline.vue
+++ b/Apps/WebClient/src/ClientApp/app/views/timeline.vue
@@ -215,6 +215,7 @@
     <ProtectiveWordComponent
       ref="protectiveWordModal"
       :error="protectiveWordAttempts > 1"
+      :is-loading="isLoading"
       @submit="onProtectiveWordSubmit"
       @cancel="onProtectiveWordCancel"
     />
--- a/Apps/WebClient/src/ClientApp/app/views/timeline.vue
+++ b/Apps/WebClient/src/ClientApp/app/views/timeline.vue
@@ -604,7 +604,11 @@ export default class TimelineComponent extends Vue {
       // Set it to new final page
       this.currentPage = result.length;
     }
-    this.visibleTimelineEntries = result[this.currentPage - 1];
+    if (result.length === 0) {
+      this.visibleTimelineEntries = [];
+    } else {
+      this.visibleTimelineEntries = result[this.currentPage - 1];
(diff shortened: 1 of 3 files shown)


Apps/WebClient/src/ClientApp/app/views/timeline.vue
```diff
@@ -604,7 +604,11 @@ export default class TimelineComponent extends Vue {
       // Set it to new final page
       this.currentPage = result.length;
     }
-    this.visibleTimelineEntries = result[this.currentPage - 1];
+    if (result.length === 0) {
```
FuriousLlama, line 607: I am not 100$ sure on the getPages method. Looks like the filter is being called anytime the current page changes as well. That will cause the logic to re-create the index for the pages without using the other pages in the array (result is an array of arrays).

We should have a separate function that is watching the currentPage and adjust the visible timeline accordingly and use the filter to calculate the max size of the visible page.

Apps/WebClient/src/ClientApp/app/views/timeline.vue
```diff
@@ -399,7 +399,7 @@ export default class TimelineComponent extends Vue {
 
   private get numberOfPages(): number {
     let result = Math.ceil(
-      this.filteredEntriesLength / this.getNumberOfEntriesPerPage
+      this.filteredTimelineEntries.length / this.numberOfEntriesPerPage
     );
     if (result < 1) {
```
sslaws, line 404: I'd rather see:
let result = 1;
if (this.filteredEntriesLength > this.numberOfEntriesPerPage)
{
     result = Math.ceil(this.filteredTimelineEntries.length / this.numberOfEntriesPerPage);
}    

which is more explicit then the if afterwards.

PR #437, opened 2020-04-21
Snowplow security fixes

FuriousLlama:
## Status
**Ready**

## Fixes or Implements [AB#8249](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8249)
- [ ] Enhancement
- [X] Bug
- [ ] Documentation

## Description:
Fixes the issue where snowplow was not loading due to security flags

### DIFF
--- a/Apps/WebClient/src/appsettings.Development.json
+++ b/Apps/WebClient/src/appsettings.Development.json
@@ -67,6 +67,7 @@
   ],
   "ContentSecurityPolicy": {
     "connect-src": "https://spt.apps.gov.bc.ca/com.snowplowanalytics.snowplow/tp2 https://sso-dev.pathfinder.gov.bc.ca/ http://localhost:*",
-    "frame-src": "https://sso-dev.pathfinder.gov.bc.ca/"
+    "frame-src": "https://sso-dev.pathfinder.gov.bc.ca/",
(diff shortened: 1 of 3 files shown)


Apps/WebClient/src/appsettings.Development.json
```diff
@@ -67,6 +67,7 @@
   ],
   "ContentSecurityPolicy": {
     "connect-src": "https://spt.apps.gov.bc.ca/com.snowplowanalytics.snowplow/tp2 https://sso-dev.pathfinder.gov.bc.ca/ http://localhost:*",
-    "frame-src": "https://sso-dev.pathfinder.gov.bc.ca/"
+    "frame-src": "https://sso-dev.pathfinder.gov.bc.ca/",
+    "script-src": "https://sp-js.apps.gov.bc.ca/MDWay3UqFnIiGVLIo7aoMi4xMC4y.js"
```
tiagograf-AOT, line 71: Do you really need the file name in it? i thought just the domain would do it!

FuriousLlama: I dont think is necessary to name the file, you are right the domain would work as well. However I believe having the file in there makes sure that even if the domain gets compromised only that particular file will be allowed from that external source

PR #439, opened 2020-04-21
Pagination section calculation

FuriousLlama:
## Status
**Ready**

## Fixes or Implements [AB#8236](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8236)
- [ ] Enhancement
- [X] Bug
- [ ] Documentation

## Description:
Refactored the calculation of the pagination and sorting on the timeline

### DIFF
--- a/Apps/WebClient/src/ClientApp/app/views/timeline.vue
+++ b/Apps/WebClient/src/ClientApp/app/views/timeline.vue
@@ -276,14 +276,13 @@ export default class TimelineComponent extends Vue {
 
   private filterText: string = "";
   private timelineEntries: TimelineEntry[] = [];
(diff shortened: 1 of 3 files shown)


Apps/WebClient/src/ClientApp/app/views/timeline.vue
```diff
@@ -551,6 +539,7 @@ export default class TimelineComponent extends Vue {
     this.isAddingNote = false;
     if (note) {
       this.timelineEntries.push(new NoteTimelineEntry(note));
+      this.sortEntries();
```
tiagograf-AOT, line 542: shouldn't we move this to the applyTimelineFilter method since we are calling both of them in all those cases?

PR #445, opened 2020-04-23
Update Medication Card

sam-warren:
## Status
**Ready**

## Fixes or Implements [AB#8177](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8177)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
- Get timeline entry info from ODR instead of PharmaNet to use unique non-random IDs for timeline entries
- Moved details toggle and added comments counter that will be used as an expand/collapse button
- Updated comment service and model
- Fixed dependency injection error in startup
- Wire up comments backend to frontend

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### UI Changes
YES

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox


sam-warren:
@sslaws Updated description

Apps/WebClient/src/ClientApp/app/models/medicationTimelineEntry.ts
```diff
@@ -14,7 +14,15 @@ export default class MedicationTimelineEntry extends TimelineEntry {
   public prescriptionIdentifier: string;
 
   public constructor(model: MedicationStatement | MedicationStatementHistory) {
-    super("id-" + Math.random(), EntryType.Medication, model.dispensedDate);
+    
+    if (model.prescriptionIdentifier) {
+      // ODR result
+      super("id-" + model.prescriptionIdentifier, EntryType.Medication, model.dispensedDate);
```
FuriousLlama, line 20: I dont think  'id-' is necessary

Apps/WebClient/src/ClientApp/app/models/timelineEntry.ts
```diff
@@ -10,9 +10,9 @@ export enum EntryType {
 export default abstract class TimelineEntry {
   public readonly id: string;
   public readonly type: EntryType;
-  public readonly date: Date;
+  public readonly date?: Date;
```
FuriousLlama, line 13: I think the date is not optional

Apps/WebClient/src/ClientApp/app/components/timeline/medication.vue
```diff
@@ -159,6 +166,17 @@ $radius: 15px;
             </b-collapse>
           </b-col>
         </b-row>
+        <b-row>
```
FuriousLlama, line 169: Not necessary right now, but. How hard would it be to move all logic and styles to a different component? That way it can be reused in the future for imms or other type of data.

Apps/WebClient/src/ClientApp/app/components/timeline/medication.vue
```diff
@@ -36,6 +36,11 @@ $radius: 15px;
   padding: 0px;
 }
 
+.commentButton {
```
FuriousLlama, line 39: nit instead of this the element could use py-2

Apps/WebClient/src/ClientApp/app/services/restUserCommentService.ts
```diff
@@ -18,24 +17,10 @@ export class RestUserCommentService implements IUserCommentService {
     this.isEnabled = config.webClient.modules["Comment"];
   }
 
-  public getComments(): Promise<RequestResult<UserComment[]>> {
+  public getCommentsForEntry(parentEntryId: string): Promise<RequestResult<UserComment[]>> {
     return new Promise((resolve, reject) => {
-      if (!this.isEnabled) {
```
FuriousLlama, line 23: we will need to have the isEnabled flag in the retrieving methods. That is because the Comment feature depends on the ODR feature so we want to enable/disable based on configuration

Apps/WebClient/src/ClientApp/app/models/timelineEntry.ts
sam-warren: In medicationStatement and medicationStatementHistory models the dispensedDate of the medication is optional... do we want to change this to mandatory to match timelineEntry then?

Apps/WebClient/src/ClientApp/app/components/timeline/medication.vue
sam-warren: Can look into that in the next ticket.

Apps/WebClient/src/ClientApp/app/models/timelineEntry.ts
FuriousLlama: I believe so, the date is a must for all sorting and rendering. It must also be on the backed

PR #446, opened 2020-04-23
Features/8194

FuriousLlama:
## Status
**Ready**

## Fixes or Implements [AB#8194](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8194)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Adds encryption to the note service and comments. On initial load of notes will convert previous un-encrypted and add a user profile key

## Testing
- [x] Unit Tests Updated
- [ ] Not Required

### UI Changes
NO

## Notes/Additional Comments
Profile generation of the key happens on profile creation


Apps/WebClient/src/Server/Models/UserNote.cs
```diff
@@ -0,0 +1,151 @@
+//-------------------------------------------------------------------------
```
sslaws, line 1: Question:  Why repeat the DB model for both Notes and Comments?  Why not add the utility classes into the services?  Not sure I see a lot of value unless the annotations interact with the controller return data?

Apps/WebClient/src/Server/Services/NoteService.cs
```diff
@@ -29,72 +31,145 @@ public class NoteService : INoteService
...
         /// <inheritdoc />
-        public RequestResult<Note> CreateNote(Note note)
+        public RequestResult<UserNote> CreateNote(UserNote userNote)
         {
+            UserProfile profile = this.profileDelegate.GetUserProfile(userNote.HdId).Payload;
+            string? key = profile.EncryptionKey;
+            if (key == null)
+            {
+                this.logger.LogInformation($"User does not have a key: ${userNote.HdId}");
+                throw new ApplicationException("Profile key not set");
```
sslaws, line 60: Should this be an ApplicationException or a failed result which would be processed cleanly by the UI?

Apps/WebClient/src/Server/Services/NoteService.cs
```diff
@@ -29,72 +31,145 @@ public class NoteService : INoteService
...
             return result;
         }
 
         /// <inheritdoc />
-        public RequestResult<Note> UpdateNote(Note note)
+        public RequestResult<UserNote> UpdateNote(UserNote userNote)
         {
+            UserProfile profile = this.profileDelegate.GetUserProfile(userNote.HdId).Payload;
+            string? key = profile.EncryptionKey;
+            if (key == null)
```
sslaws, line 116: Same comment as above, should we be throwing exceptions to the UI.

Apps/WebClient/src/Server/Services/NoteService.cs
```diff
@@ -29,72 +31,145 @@ public class NoteService : INoteService
...
 
         /// <inheritdoc />
-        public RequestResult<Note> DeleteNote(Note note)
+        public RequestResult<UserNote> DeleteNote(UserNote userNote)
         {
+            UserProfile profile = this.profileDelegate.GetUserProfile(userNote.HdId).Payload;
+            string? key = profile.EncryptionKey;
+            if (key == null)
+            {
+                this.logger.LogInformation($"User does not have a key: ${userNote.HdId}");
```
sslaws, line 141: ditto

Apps/Common/src/Delegates/HMACHashDelegate.cs
```diff
@@ -136,7 +136,7 @@ public IHash Hash(string? key)
         }
 
         /// <inheritdoc />
-        public bool Compare(string key, IHash compareHash)
+        public bool Compare(string? key, IHash compareHash)
```
sslaws, line 139: Is this a real change?

Apps/WebClient/src/Server/Models/UserNote.cs
FuriousLlama: The main reason to have them separate now is because they dont represent the same data anymore. Once is encrypted (db model) while the other is a user representation. Having them separate makes a clear distinction between them and gives us the flexibility to further change each of them individually and map them at the object level.

Apps/WebClient/src/Server/Services/NoteService.cs
FuriousLlama: I think in this particular case the error should not be returned to the user since there is not much for the user to do. If there is no key at this point is a 'catastrophic' failure that we will have to look into it. However I am not completely sold on the type of exception. Thoughts?

Apps/Common/src/Delegates/HMACHashDelegate.cs
FuriousLlama: Yeah, It was causing a smell since the implementation was not implementing the interface exactly

PR #465, opened 2020-04-30
Add covid modal component

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8286](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8286)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Add covid modal.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required


Apps/WebClient/src/ClientApp/app/views/timeline.vue
```diff
@@ -297,6 +305,8 @@ export default class TimelineComponent extends Vue {
 
   @Ref("protectiveWordModal")
   readonly protectiveWordModal!: ProtectiveWordComponent;
+  @Ref("covidModal")
+  readonly covidModal!: CovidModalComponent;
```
FuriousLlama, line 309: not sure we need this reference since its only used to set the showing to the component. Instead we can just pass it as a prop (I think is already passing it)

Apps/WebClient/src/ClientApp/app/views/timeline.vue
```diff
@@ -319,6 +329,9 @@ export default class TimelineComponent extends Vue {
     EventBus.$on("idleLogoutWarning", function(isVisible: boolean) {
       self.idleLogoutWarning = isVisible;
     });
+
+    // To be updated with Enabled/Disabled covid message configuration
+    this.covidModal.showModal();
```
FuriousLlama, line 334: Instead of using  the component show modal method we could set the banner property to true show-covid-modal = true

Apps/WebClient/src/ClientApp/app/views/timeline.vue
```diff
@@ -473,6 +486,14 @@ export default class TimelineComponent extends Vue {
       });
   }
 
+  private onCovidSubmit() {
```
FuriousLlama, line 489: What do you envision these will be useful for?

Apps/WebClient/src/ClientApp/app/components/modal/covid.vue
```diff
@@ -0,0 +1,130 @@
...
+        </form>
+      </b-col>
+    </b-row>
+    <template v-slot:modal-footer>
+      <b-row>
+        <b-col>
+          <b-row>
+            <b-col>
+              <b-button size="lg" variant="outline-primary" @click="handleYes($event)">
+                YES
```
FuriousLlama, line 45: nit: YES => Yes | NO => No

Apps/WebClient/src/ClientApp/app/components/modal/covid.vue
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  public hideModal() {
+    this.isShowing = false;
+    this.$bvModal.hide(this.modalId);
+  }
+
+  @Watch("isLoading")
+  private onIsLoading() {
+    if (!this.isLoading && this.isShowing) {
```
FuriousLlama, line 83: I think we want to show the modal right away

Apps/WebClient/src/ClientApp/app/components/modal/covid.vue
```diff
@@ -0,0 +1,130 @@
...
+    // Trigger cancel handler
+    this.cancel();
+
+    // Hide the modal manually
+    this.$nextTick(() => {
+      this.hideModal();
+    });
+  }
+
+  private handleSubmit() {
```
FuriousLlama, line 121: this method can be put on handleYes

Apps/WebClient/src/ClientApp/app/views/timeline.vue
tiagograf-AOT: Opening up the new tab with the results page I believe.

Apps/WebClient/src/ClientApp/app/views/timeline.vue
tiagograf-AOT: Yes, the problem with showing the component is that the close is handled by the actual modal which makes this a two way binding for a "show", it just raises the complexity, + i'm following how the protective word modal works today!

FuriousLlama: fair point

PR #469, opened 2020-05-05
Cleanup project warnings for Medication AB#8301

sslaws:
## Status
Ready

## Fixes or Implements [AB#8301](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8301)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Resolved all solutions warnings for Medication.

## Testing
- [X] Unit Tests Updated
- [ ] Not Required

### Steps to Reproduce:
N/A

### UI Changes
No

### Browsers Tested
- [ ] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

Unit tests and validation of ODR protected word were completed, I also ran Pharmanet but was unable to confirm Protective word.

## Notes/Additional Comments
N/A

### DIFF
--- a/Apps/Common/src/Models/RequestResult.cs
+++ b/Apps/Common/src/Models/RequestResult.cs
@@ -27,7 +27,7 @@ namespace HealthGateway.Common.Models
         /// <summary>
         /// Gets or sets the result payload.
         /// </summary>
-        public T ResourcePayload { get; set; } = null!;
+        public T? ResourcePayload { get; set; } = null;
 
         /// <summary>
         /// Gets or sets the total result count for the request for pagination.
--- a/Apps/Medication/src/Delegates/RestHNClientDelegate.cs
+++ b/Apps/Medication/src/Delegates/RestHNClientDelegate.cs
@@ -106,7 +106,15 @@ namespace HealthGateway.Medication.Delegates
                 if (response.IsSuccessStatusCode)
                 {
                     HNMessage<string> responseMessage = JsonConvert.DeserializeObject<HNMessage<string>>(payload);
-                    retVal = this.medicationParser.ParseResponseMessage(responseMessage.Message!);
+                    if (responseMessage != null && responseMessage.Message != null)
+                    {
+                        retVal = this.medicationParser.ParseResponseMessage(responseMessage.Message);
+                    }
+                    else
+                    {
+                        this.logger.LogError($"Parsed payload is null or not valid: {payload}");
+                        retVal = new HNMessage<List<MedicationStatement>>(Common.Constants.ResultType.Error, $"Unable to parse response");
+                    }
                 }
                 else
                 {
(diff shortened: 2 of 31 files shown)


Apps/Common/src/Models/RequestResult.cs
```diff
@@ -27,7 +27,7 @@ public class RequestResult<T>
         /// <summary>
         /// Gets or sets the result payload.
         /// </summary>
-        public T ResourcePayload { get; set; } = null!;
+        public T? ResourcePayload { get; set; } = null;
```
FuriousLlama, line 30: Can the resource payload type be null?

Apps/Medication/src/Delegates/RestHNClientDelegate.cs
```diff
@@ -151,7 +150,15 @@ public async Task<HNMessage<Pharmacy>> GetPharmacyAsync(string pharmacyId, strin
                 if (response.IsSuccessStatusCode)
                 {
                     HNMessage<string> responseMessage = JsonConvert.DeserializeObject<HNMessage<string>>(payload);
-                    retVal = this.pharmacyParser.ParseResponseMessage(responseMessage.Message);
+                    if (responseMessage != null && responseMessage.Message != null)
```
FuriousLlama, line 153: should we have the same on line 109?

Apps/Medication/src/Models/FederalDrugSource.cs
```diff
@@ -31,6 +31,6 @@ public class FederalDrugSource
         /// <summary>
         /// Gets or sets the drug product.
         /// </summary>
-        public DrugProduct DrugProduct { get; set; }
+        public DrugProduct? DrugProduct { get; set; }
```
FuriousLlama, line 34: I dont think it makes sense to have the DrugProduct as nullable if its parent object exist at all. It would be a date holder without it.

Apps/Medication/src/Models/HNMessage.cs
```diff
@@ -67,7 +67,7 @@ public HNMessage(T response, ResultType result, string resultMessage)
         /// <summary>
         /// Gets or sets the response message.
         /// </summary>
-        public T Message { get; set; }
+        public T? Message { get; set; }
```
FuriousLlama, line 70: if the message is null, would that not mean that something went bad?

Apps/Medication/src/Models/FederalDrugSource.cs
FuriousLlama: btw, this is the only one that is a _show stopper_ for me

sslaws: I can change this, the only reason I set it to nullable was because we have to initialize it to an empty Drug which we throw away.   Its wasted Object allocation but I've added it.

Apps/Medication/src/Models/HNMessage.cs
sslaws: We have two constructors that set the Message to null.

Apps/Medication/src/Delegates/RestHNClientDelegate.cs
sslaws: Yeah... fixed.

PR #502, opened 2020-05-20
update modal component

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements AB#????
- [ ] Enhancement
- [x] Bug
- [ ] Documentation

## Description:
Fixes protective word modal showing up with covid alert modal.

## Testing
- [ ] Unit Tests Updated
- [ ] Not Required

### Steps to Reproduce:
Login with protected user, both alerts should show up "Covid" and "Protective Word" only 1 modal should show at a time. covid and then protective.

### UI Changes
NO


Apps/WebClient/src/ClientApp/app/components/modal/covid.vue
```diff
@@ -7,6 +7,7 @@
 <template>
   <b-modal
     id="covid-modal"
+    v-model="isVisible"
```
FuriousLlama, line 10: So this sets the visibility? I like it! guess that came with the latest package update

PR #507, opened 2020-05-21
Wired up notification settings backend

FuriousLlama:
## Status
**Ready**

## Fixes or Implements [AB#8386](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8386)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Updated tests. 
Added phone service. 
Wired profile update with notification update calls

## Testing
- [X] Unit Tests Updated
- [ ] Not Required

### UI Changes
 NO

### DIFF
--- a/Apps/Common/src/Models/NotificationSettingsRequest.cs
+++ b/Apps/Common/src/Models/NotificationSettingsRequest.cs
@@ -31,7 +31,8 @@ namespace HealthGateway.Common.Models
         }
 
         /// <summary>
-        /// Copy constructor for a instance of the <see cref="NotificationSettingsRequest"/> class.
+        /// Initializes a new instance of the <see cref="NotificationSettingsRequest"/> class.
+        /// Copy constructor that uses a <see cref="NotificationSettingsBase"/> for initialization.
         /// </summary>
         /// <param name="notificationSettings">Initialize values from passed in object.</param>
         public NotificationSettingsRequest(NotificationSettingsBase notificationSettings)
--- a/Apps/Common/src/Models/NotificationSettingsRequest.cs
+++ b/Apps/Common/src/Models/NotificationSettingsRequest.cs
@@ -31,12 +31,26 @@ namespace HealthGateway.Common.Models
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="NotificationSettingsRequest"/> class.
+        /// Copy constructor for a instance of the <see cref="NotificationSettingsRequest"/> class.
         /// </summary>
         /// <param name="notificationSettings">Initialize values from passed in object.</param>
         public NotificationSettingsRequest(NotificationSettingsBase notificationSettings)
             : base(notificationSettings)
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NotificationSettingsRequest"/> class using only the email address and phone number.
+        /// </summary>
+        /// <param name="emailAddress">Email address for the notification settings.</param>
+        /// <param name="smsNumber">SMS Number for the notification settings.</param>
+        public NotificationSettingsRequest(string? emailAddress, string? smsNumber)
+            : base()
+        {
+            this.EmailAddress = emailAddress;
+            this.EmailEnabled = !string.IsNullOrWhiteSpace(emailAddress);
+            this.SMSNumber = smsNumber;
+            this.SMSEnabled = !string.IsNullOrWhiteSpace(smsNumber);
         }
 
         /// <summary>
--- a/Apps/WebClient/src/Server/Controllers/UserProfileController.cs
+++ b/Apps/WebClient/src/Server/Controllers/UserProfileController.cs
@@ -24,6 +24,7 @@ namespace HealthGateway.WebClient.Controllers
     using HealthGateway.Database.Models;
     using HealthGateway.WebClient.Models;
     using HealthGateway.WebClient.Services;
+    using Microsoft.AspNetCore.Authentication;
     using Microsoft.AspNetCore.Authorization;
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Mvc;
@@ -101,7 +102,7 @@ namespace HealthGateway.WebClient.Controllers
                 .Referer
                 .GetLeftPart(UriPartial.Authority);
 
-            string bearerToken = this.httpContextAccessor.HttpContext.Request.Headers["Authorization"][0].Split(" ")[1];
+            string bearerToken = await this.httpContextAccessor.HttpContext.GetTokenAsync("access_token").ConfigureAwait(true);
 
             RequestResult<UserProfileModel> result = this.userProfileService.CreateUserProfile(createUserRequest, new Uri(referer), bearerToken);
             return new JsonResult(result);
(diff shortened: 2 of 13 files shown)


Apps/Common/src/Models/NotificationSettingsRequest.cs
```diff
@@ -31,14 +31,28 @@ public NotificationSettingsRequest()
...
-        /// Initializes a new instance of the <see cref="NotificationSettingsRequest"/> class.
+        /// Copy constructor for a instance of the <see cref="NotificationSettingsRequest"/> class.
         /// </summary>
         /// <param name="notificationSettings">Initialize values from passed in object.</param>
         public NotificationSettingsRequest(NotificationSettingsBase notificationSettings)
             : base(notificationSettings)
         {
         }
 
+        /// <summary>
```
sslaws, line 42: This seems to go against the norm and creates a constructor when we can use an object initializer.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
```diff
@@ -331,7 +339,9 @@ public async Task<IActionResult> UpdateUserEmail(string hdid, [FromBody] string
                 .Referer
                 .GetLeftPart(UriPartial.Authority);
 
-            bool result = this.userEmailService.UpdateUserEmail(hdid, emailAddress, new Uri(referer));
+            string bearerToken = this.httpContextAccessor.HttpContext.Request.Headers["Authorization"][0].Split(" ")[1];
```
sslaws, line 342: This seems more succinct:
` string accessToken = await this.httpContextAccessor.HttpContext.GetTokenAsync("access_token").ConfigureAwait(true);`

Its also consistent with what I did in lab.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
```diff
@@ -373,9 +383,10 @@ public async Task<IActionResult> UpdateUserPhoneNumber(string hdid, [FromBody] s
                 .Referer
                 .GetLeftPart(UriPartial.Authority);
 
-            //bool result = this.userEmailService.UpdateUserEmail(hdid, phoneNumber, new Uri(referer));
-            //return new JsonResult(result);
-            return new JsonResult(null);
+            string bearerToken = this.httpContextAccessor.HttpContext.Request.Headers["Authorization"][0].Split(" ")[1];
```
sslaws, line 386: See comment above.

Apps/WebClient/src/Server/Services/UserEmailService.cs
```diff
@@ -104,5 +120,16 @@ public bool UpdateUserEmail(string hdid, string email, Uri hostUri)
             this.logger.LogDebug($"Finished updating user email");
             return true;
         }
+
+        private async void UpdateNotificationSettings(UserProfile userProfile, string bearerToken)
+        {
+            // Update the notification settings
+            NotificationSettingsRequest request = new NotificationSettingsRequest(userProfile.Email, userProfile.PhoneNumber);
```
sslaws, line 127: I do like this better than object initialization... we should probably have some guidance on when object initialization should be used over constructors.

Apps/Common/src/Models/NotificationSettingsRequest.cs
```diff
@@ -31,14 +31,28 @@ public NotificationSettingsRequest()
         }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="NotificationSettingsRequest"/> class.
+        /// Copy constructor for a instance of the <see cref="NotificationSettingsRequest"/> class.
```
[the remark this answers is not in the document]
sslaws: Approved by mistake … comments included in original.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
FuriousLlama: Yes, but is not testable since GetTokenAsync is an extension method :(

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
FuriousLlama: same as above

Apps/WebClient/src/Server/Services/UserEmailService.cs
FuriousLlama: well in the case of the constructor it carries logic to simplify the setting of the variables.

Apps/WebClient/src/Server/Services/UserPhoneService.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.profileDelegate = profileDelegate;
+            this.notificationSettingsService = notificationSettingsService;
+        }
+
+        /// <inheritdoc />
+        public bool UpdateUserPhone(string hdid, string phone, Uri hostUri, string bearerToken)
+        {
+            this.logger.LogTrace($"Updating user phone numbner...");
+            UserProfile userProfile = this.profileDelegate.GetUserProfile(hdid).Payload;
+            userProfile.PhoneNumber = phone;
```
sslaws, line 55: is phone stripped down to its core digits?  (555) 123-3456 should be stored in DB as 5551233456

FuriousLlama: I would guess that happens on the front end model. Will check with sam since thats only representation

Apps/Common/src/Models/NotificationSettingsRequest.cs
FuriousLlama: We can, but to of the variables can be derived so I believe the constructor is better in this case

Apps/Common/src/Models/NotificationSettingsRequest.cs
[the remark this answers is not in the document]
FuriousLlama: Interesting, I'll update it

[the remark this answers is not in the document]
sslaws: Yes, anything other than the first initializes a new... causes a warning.

Apps/Common/src/Models/NotificationSettingsRequest.cs
sslaws: When I saw the usage, I like it better as well.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
sslaws: Ahh but it is... See Laboratory Unit tests, pretty straight forward.

Apps/WebClient/src/Server/Services/UserPhoneService.cs
sslaws: Sounds good as I created a bug to update the DB model to be a max of 10 chars.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
FuriousLlama: uhm, the amount of stuff mocked for this is quite large...

Apps/Common/src/Models/NotificationSettingsRequest.cs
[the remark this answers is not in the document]
FuriousLlama: got it. Fixed

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
FuriousLlama: fixed

PR #521, opened 2020-05-28
Features/8393

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8393](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8393)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Adds the resend phone verification and verification input to profile page.

## Testing
- [x] Unit Tests Updated
- [ ] Not Required

### UI Changes
YES

### Browsers Tested
- [x] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

Tests to be updated once backend is completed.


Apps/WebClient/src/Server/Controllers/UserProfileController.cs
```diff
@@ -303,6 +331,64 @@ public async Task<IActionResult> GetUserEmailInvite(string hdid)
             }
         }
 
+        /// <summary>
```
sslaws, line 334: I don't understand the purpose of this method.  The actual validation code for SMS should  never be returned to the client and the method above does validation.  Can you elaborate?

tiagograf-AOT: its only used to check if the phone was validated or not, this is the part that I mention that could use some refactoring (together with email validation) the actual code/digit is not sent to the front end.

Apps/Common/src/Services/EmailQueueService.cs
```diff
@@ -192,6 +192,7 @@ private Email ParseTemplate(EmailTemplate emailTemplate, Dictionary<string, stri
             }
 
             Email email = new Email();
+            email.Id = Guid.NewGuid();
```
sslaws, line 195: For a new email, the database will generate the guid.

Apps/Database/src/Models/MessagingVerification.cs
```diff
@@ -47,7 +47,7 @@ public class MessagingVerification : AuditableEntity
         /// Gets or sets the associated email that was sent for this verification.
         /// Required if the VerificationType = MessagingVerificationType.Email.
         /// </summary>
-        public Guid EmailId { get; set; }
+        public Guid? EmailId { get; set; }
```
sslaws, line 50: This seems to have changed the column type... Do we want that?

tiagograf-AOT: Yes, emailId is nullable now.

Apps/Common/src/Services/EmailQueueService.cs
tiagograf-AOT: this was a problem when saving the the message verification, it threw an error saying the emailId cannot be null for Email verification type.

Apps/Common/test/unit/Services/Email_Test.cs
```diff
@@ -56,6 +56,7 @@ public void ShouldProcessTemplate()
                 new Mock<IMessagingVerificationDelegate>().Object,
                 new Mock<IWebHostEnvironment>().Object);
             Email actual = emailService.ProcessTemplate(emailTo, template, d);
+            expected.Id = actual.Id;
```
FuriousLlama, line 59: is this done because we dont know a-priori the id?

tiagograf-AOT: yep!

PR #523, opened 2020-05-29
Refactor PhoneNumber to SMSNumber

sam-warren:
## Status
**Ready**

## Fixes or Implements [AB#8412](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8412)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
- Changed PhoneNumber to SMSNumber for all user profile related applications across front and back end
- Set character limit of SMSNumber to 10
- Add SMSNumber to UserProfileHistory to sync with UserProfile model

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### UI Changes
NO

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox


Apps/Database/src/Migrations/20200528223039_PhoneBugfixes.cs
```diff
@@ -0,0 +1,50 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace HealthGateway.Database.Migrations
```
sslaws, line 3: This migration will result in the loss of information, it should be an alter column instead of drop.  I would prefer if you could remove the migration and reset the model then try making the rename change first and verify if the migration created is an alter.

Apps/Database/src/Migrations/20200528223039_PhoneBugfixes.cs
```diff
@@ -0,0 +1,50 @@
...
+            migrationBuilder.AddColumn<string>(
+                name: "PhoneNumber",
+                schema: "gateway",
+                table: "UserProfile",
+                type: "character varying(20)",
+                maxLength: 20,
+                nullable: true);
+        }
+    }
+}
```
sslaws, line 50: You also have to manually update the trigger see migration 20200422234729_SupportEncryption

Apps/Database/src/Migrations/20200528223039_PhoneBugfixes.cs
```diff
@@ -0,0 +1,50 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
```
FuriousLlama, line 1: also, missing top documentation

Apps/Database/src/Migrations/20200528223039_PhoneBugfixes.Designer.cs
```diff
@@ -0,0 +1,2985 @@
+﻿// <auto-generated />
```
FuriousLlama, line 1: missing top docs

Apps/Database/src/Migrations/20200528223039_PhoneBugfixes.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace HealthGateway.Database.Migrations
+{
+    public partial class PhoneBugfixes : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropColumn(
```
sslaws, line 9: We should be able to avoid dropping the column by generating two migrations.

PR #532, opened 2020-06-04
Updated to use system JWT for Queued Push to PHSA

bradhead:
## Status
**Ready**

## Fixes or Implements AB8450 NotificationSettings queued push to use system JWT
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
NotificationSettings queued push to use system JWT, but only when Queued, not direct.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required

### Steps to Reproduce:
The push notificiation should authenticate with PHSA endpoint, not using user JWT,
which is not available to job scheduler, as either expired or not set. This way the Job Scheduler will obtain JWT at time to delivery to PHSA endpoint, allowing for validation.
the JSON notification structure adds subjectHdid as now required.

### UI Changes
NO

### Browsers Tested
- [ ] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

## Notes/Additional Comments
We will need to update design notes to PHSA.

### DIFF
--- a/Apps/Common/src/Models/NotificationSettingsBase.cs
+++ b/Apps/Common/src/Models/NotificationSettingsBase.cs
@@ -38,6 +38,7 @@ namespace HealthGateway.Common.Models
         /// <param name="notificationSettings">Initialize values from passed in object.</param>
         public NotificationSettingsBase(NotificationSettingsBase notificationSettings)
         {
+            this.SubjectHdid = notificationSettings.SubjectHdid;
             this.SMSEnabled = notificationSettings.SMSEnabled;
             this.SMSNumber = notificationSettings.SMSNumber;
             this.SMSVerified = notificationSettings.SMSVerified;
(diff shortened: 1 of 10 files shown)


Apps/Common/src/Models/NotificationSettingsBase.cs
```diff
@@ -38,6 +38,7 @@ public NotificationSettingsBase()
         /// <param name="notificationSettings">Initialize values from passed in object.</param>
         public NotificationSettingsBase(NotificationSettingsBase notificationSettings)
         {
+            this.SubjectHdid = notificationSettings.SubjectHdid;
```
sslaws, line 41: This needs to be moved to NotificationSettingsRequest as it is specific to the request object.  Unless you convince Tyler to echo it back :).

bradhead: we need hdid in the response to correlate or the response can't really be meaningfully processed in a stateless manner. I sent Tyler the change that we will be putting hdid in request and response payload. If there is a reason that's hardship we can move the hdid to the Request  only.

bradhead:
I didn't touch a line of code in WebClient so why is it failing?

Apps/Common/src/Models/NotificationSettingsBase.cs
bradhead: Can I get consensus on this? Would we ever need to process the response in the context of the hdid outside of the synchronous call? If not, I can put hdid only in request.  I did mention to Tyler we'd put it in both request and response.

PR #542, opened 2020-06-09
Features/8448

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements AB#8448
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Expires email and sms invite in 5 days.
Prevent more than 5 attempts to validate emails and sms.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required


Apps/Common/src/Services/EmailQueueService.cs
```diff
@@ -34,6 +34,7 @@ namespace HealthGateway.Common.Services
     /// </summary>
     public class EmailQueueService : IEmailQueueService
     {
+        private const byte VerificationExpiryDays = 5;
```
FuriousLlama, line 37: integer instead of byte

Apps/WebClient/src/Server/Services/UserSMSService.cs
```diff
@@ -30,6 +30,8 @@ namespace HealthGateway.WebClient.Services
     /// <inheritdoc />
     public class UserSMSService : IUserSMSService
     {
+        private const byte VerificationExpiryDays = 5;
```
FuriousLlama, line 33: why byte instead of integer

Apps/WebClient/src/Server/Services/UserSMSService.cs
```diff
@@ -63,6 +65,8 @@ public bool ValidateSMS(string hdid, string validationCode, string bearerToken)
             if (smsInvite != null &&
                 smsInvite.HdId == hdid &&
                 !smsInvite.Validated &&
+                !smsInvite.Deleted &&
+                smsInvite.VerificationAttempts < MaxVerificationAttempts &&
```
FuriousLlama, line 69: Do the verification attempts ever get reset?

sslaws: Do we need the deleted?

tiagograf-AOT: Verification attempts do not reset, once its past 5 tries the email/sms is gone and the user have to reenter it.

tiagograf-AOT: We need the deleted for when the user removes their email/sms.

PR #565, opened 2020-06-16
Bugs/8522

sam-warren:
## Status
**Ready**

## Fixes or Implements [AB#8522](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8522)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Changed SMS to Phone in text field, fix period spacing

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### UI Changes
YES

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,6 +2,3 @@
 # Set default behavior to automatically normalize line endings.
 ###############################################################################
 * text=auto
-
-*.vue text eol=lf
-*.ts text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,6 @@
 # Set default behavior to automatically normalize line endings.
 ###############################################################################
 * text=auto
+
+*.vue text eol=lf
+*.ts text eol=lf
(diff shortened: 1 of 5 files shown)


.gitattributes
```diff
@@ -2,3 +2,6 @@
 # Set default behavior to automatically normalize line endings.
 ###############################################################################
 * text=auto
+
+*.vue text eol=lf
```
FuriousLlama, line 6: not sure this is what we want since I think all files should be in lf. Thoughts?

sam-warren: want me to change it to `* text eol=lf`?

sslaws: I think we do!

FuriousLlama: Uhm not sure. Do we also want the CS and json files to be lf?

PR #572, opened 2020-06-19
Add update communication delegate

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8542](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8542)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Add update to communication delegate.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required

### UI Changes
NO

DB Delegates do not require unit tests.


Apps/Database/src/Delegates/DBCommunicationDelegate.cs
```diff
@@ -74,6 +74,13 @@ public DBResult<Communication> Add(Communication communication, bool commit = tr
                 Payload = communication,
                 Status = DBStatusCode.Deferred,
             };
+            if (this.CheckIsOverlappingDate(communication))
```
FuriousLlama, line 77: CheckIsOverlappingDate => IsOverlappingDate since its a private method

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
```diff
@@ -93,5 +100,48 @@ public DBResult<Communication> Add(Communication communication, bool commit = tr
...
+            this.logger.LogTrace($"Updating Communication in DB...");
+            DBResult<Communication> result = new DBResult<Communication>()
+            {
+                Payload = communication,
+                Status = DBStatusCode.Deferred,
+            };
+            if (this.CheckIsOverlappingDate(communication))
+            {
+                this.logger.LogDebug($"Communication date range overlap with existing entry");
+                result.Status = DBStatusCode.Error;
```
FuriousLlama, line 116: Lets add why it the error in the message

leobcgov:
can we confirm it should be ok if communication.EffectiveDateTime == c.ExpiryDateTime ||
                c.EffectiveDateTime == communication.ExpiryDateTime ?

tiagograf-AOT:
I don't see a problem with that Leo since one is starting when the other one is ending!

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
```diff
@@ -93,5 +100,48 @@ public DBResult<Communication> Add(Communication communication, bool commit = tr
...
+                return result;
+            }
+
```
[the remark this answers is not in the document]
bradhead: can you block it out in a transaction?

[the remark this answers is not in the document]
tiagograf-AOT: Added the migration that should do the work for us!

Apps/Database/src/Migrations/20200619221413_CommunicationDateConstraint.cs
```diff
@@ -0,0 +1,48 @@
...
+            migrationBuilder.Sql(constraint);
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            string schema = "gateway";
+            string triggerFunction = @$"
+                ALTER TABLE {schema}.""Communication"" DROP CONSTRAINT unique_date_range";
+
+            migrationBuilder.Sql(triggerFunction);
```
sslaws, line 45: Should be called constraint instead of trigger.

Apps/Database/src/Migrations/20200619221413_CommunicationDateConstraint.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace HealthGateway.Database.Migrations
+{
+    public partial class CommunicationDateConstraint : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            string schema = "gateway";
+            // Delete all records to prevent invalid data.
+            string deleteRecords = @$"
```
sslaws, line 28: I don't think we should be doing this - do we have existing data that blocks the migration?

PR #573, opened 2020-06-19
Fix build

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8537](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8537)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Fixes compilation error in communication component.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/communication.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/communication.vue
@@ -26,28 +26,31 @@ import { ICommunicationService } from "@/services/interfaces";
 
 @Component
 export default class CommunicationComponent extends Vue {
-  private communication: Communication | null = null;
+  private isLoaded: boolean = false;
+  private communication!: Communication;
 
   private mounted() {
     this.fetchCommunication();
   }
 
-  private get text(): string {
-    return this.communication?.text ?? "";
+  private get hasCommunication(): boolean {
+    return this.isLoaded && this.communication != null;
   }
 
-  private get hasCommunication(): boolean {
-    return this.communication !== null;
+  private get text(): string {
+    return this.communication ? this.communication.text : "";
   }
 
   private fetchCommunication() {
+    let self = this;
     const communicationService: ICommunicationService = container.get(
       SERVICE_IDENTIFIER.CommunicationService
     );
     communicationService
       .getActive()
       .then((requestResult) => {
-        this.communication = requestResult.resourcePayload;
+        self.isLoaded = true;
+        self.communication = requestResult.resourcePayload;
       })
       .catch((err) => {
         console.log(err);
--- a/Apps/WebClient/src/ClientApp/src/components/communication.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/communication.vue
@@ -26,7 +26,7 @@ import { ICommunicationService } from "@/services/interfaces";
 
 @Component
 export default class CommunicationComponent extends Vue {
-  private communication?: Communication = null;
+  private communication: Communication | null = null;
 
   private mounted() {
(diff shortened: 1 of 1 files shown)


Apps/WebClient/src/ClientApp/src/components/communication.vue
```diff
@@ -26,18 +26,14 @@ import { ICommunicationService } from "@/services/interfaces";
 
 @Component
 export default class CommunicationComponent extends Vue {
-  private communication?: Communication = null;
+  private communicationText: string = "";
```
FuriousLlama, line 29: having the communication object sounds right, it can be undefined at the begining so

```
 private communication?: Communication;
private get text(): string {
    return this.communication?.text || "";
  }
...

Comment (tiagograf-AOT) [reply Apps/WebClient/src/ClientApp/src/components/communication.vue]: the communication needs an initial value of null otherwise the computed properties won't work properly, not sure if this is a bug with vue but i was struggling with it for a while.

**Comment by tiagograf-AOT** [comment]
If the variable is set to undefined vue won't understand the object was declared and all computed properties won't work properly, I had to use null.

File: Apps/WebClient/src/ClientApp/src/components/communication.vue [hunk 2]
```diff
@@ -26,18 +26,18 @@ import { ICommunicationService } from "@/services/interfaces";
 
 @Component
 export default class CommunicationComponent extends Vue {
-  private communication?: Communication = null;
+  private communication: Communication | null = null;
```
FuriousLlama, line 29: why null instead of undefined?

tiagograf-AOT: If the variable is set to undefined vue won't understand the object was declared and all computed properties won't work properly, I had to use null.

FuriousLlama:
```
<style lang="scss" scoped>
@import "@/assets/scss/_variables.scss";

.communication {
  background-color: $bcgold;
  color: black;
}
</style>
<template>
  <b-row v-if="hasCommunication">
    <b-col class="p-0">
      <div class="m-0 py-3 text-center communication">
        <span>{{ text }}</span>
      </div>
    </b-col>
  </b-row>
</template>

<script lang="ts">
import Vue from "vue";
import { Component } from "vue-property-decorator";
import container from "@/plugins/inversify.config";
import { SERVICE_IDENTIFIER } from "@/plugins/inversify";
import Communication from "@/models/communication";
import { ICommunicationService } from "@/services/interfaces";

@Component
export default class CommunicationComponent extends Vue {
  private isLoaded: boolean = false;
  private communication!: Communication;

  private mounted() {
    this.fetchCommunication();
  }

  private get hasCommunication(): boolean {
    return this.isLoaded && this.communication != null;
  }

  private get text(): string {
    return this.communication ? this.communication.text : "";
  }

  private fetchCommunication() {
    let self = this;
    const communicationService: ICommunicationService = container.get(
      SERVICE_IDENTIFIER.CommunicationService
    );
    communicationService
      .getActive()
      .then((requestResult) => {
        self.isLoaded = true;
        self.communication = requestResult.resourcePayload;
      })
      .catch((err) => {
        console.log(err);
      });
  }
}
</script>

```

FuriousLlama: Yeah, there are otherways to set the object reactivenes but they looked quite ugly. What do you think about that instead of null?

PR #584, opened 2020-06-23
Fix long list query bug for AB#8252

sslaws:
## Status
**Ready**

## Fixes or Implements [AB#8252](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8252)
- [ ] Enhancement
- [X] Bug
- [ ] Documentation

## Description:
Queries to fetch drug information with lists were not normalizing the list before running the query.
Updated the queries to use the distinct list.

Updated AppSettings.Development.json to use the healthgateway audience as keycloak is not returning that for the local client.

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### Steps to Reproduce:
Not visible via UI, data being returned to UI should be smaller and may have some performance gain.

### UI Changes
No

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

Manually tested the App which appear to be working normally.

## Notes/Additional Comments
N/A

### DIFF
--- a/Apps/Database/src/Delegates/DBDrugLookupDelegate.cs
+++ b/Apps/Database/src/Delegates/DBDrugLookupDelegate.cs
@@ -50,7 +50,8 @@ namespace HealthGateway.Database.Delegates
         public List<DrugProduct> GetDrugProductsByDIN(List<string> drugIdentifiers)
         {
             this.logger.LogTrace($"Getting list of drug products from DB... {JsonSerializer.Serialize(drugIdentifiers)}");
-            List<DrugProduct> retVal = this.dbContext.DrugProduct.Where(dp => drugIdentifiers.Contains(dp.DrugIdentificationNumber))
+            List<string> uniqueDrugIdentifers = drugIdentifiers.Distinct().ToList();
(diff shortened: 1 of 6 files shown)


Apps/Database/src/Delegates/DBDrugLookupDelegate.cs
```diff
@@ -50,7 +50,8 @@ public class DBDrugLookupDelegate : IDrugLookupDelegate
         public List<DrugProduct> GetDrugProductsByDIN(List<string> drugIdentifiers)
         {
             this.logger.LogTrace($"Getting list of drug products from DB... {JsonSerializer.Serialize(drugIdentifiers)}");
-            List<DrugProduct> retVal = this.dbContext.DrugProduct.Where(dp => drugIdentifiers.Contains(dp.DrugIdentificationNumber))
+            List<string> uniqueDrugIdentifers = drugIdentifiers.Distinct().ToList();
```
leobcgov, line 53: Is it possible to centralize this code for getting distinct list, as it seems that this is called in multiple places?

sslaws: I moved it to the lower layer in the stack so that it is centralized.  We have 3 different methods that take 3 different lists though.

PR #586, opened 2020-06-24
Server side unit tests for Comms

sam-warren:
## Status
**Ready**

## Fixes or Implements [AB#8538](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8538)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
- Added test project to AdminWebClient
- Wrote unit tests for Communication service/controller
- Added SonarQube support in AdminWebClient
- Modified structure of test folder on AdminWebClient

## Testing
- [X] Unit Tests Updated
- [ ] Not Required

### UI Changes
NO

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox


Apps/AdminWebClient/src/ClientApp/src/jest.config.js
```diff
@@ -0,0 +1,33 @@
+module.exports = {
```
FuriousLlama, line 1: nit, this file is actually not used

Apps/AdminWebClient/test/functional/api/coverage.opencover.xml
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
FuriousLlama, line 1: this file should not be in git

sam-warren: removed

Apps/AdminWebClient/src/ClientApp/src/jest.config.js
sam-warren: removed

Apps/AdminWebClient/test/unit/Services.Test/CommunicationService_Test.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            // Set up delegate
+            DBResult<Communication> insertResult = new DBResult<Communication>
+            {
+                Payload = comm,
+                Status = DBStatusCode.Created
+            };
+
+            Mock<ICommunicationDelegate> communicationDelegateMock = new Mock<ICommunicationDelegate>();
+            communicationDelegateMock.Setup(s => s.Add(It.Is<Communication>(x => x.Text == comm.Text), true)).Returns(insertResult);
```
leobcgov, line 56: Few questions, but they should not block the approval:
- are we testing also if other columns/fields will be added correctly too?
- are we testing also if text or other fields are null or empty?
- Are we expected any specific errors for these odd cases above?

sam-warren: If any columns / fields are not added correctly, or are null, the assertion at the bottom of the test would fail since the data retrieved from the database would not be exactly equal to what it was given. Not sure about specific errors, the test would just fail.

PR #596, opened 2020-06-29
Update SonarQube project names

tiagograf-AOT:


Tools/SonarQube/azure/build.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
           dotnet sonarscanner begin \
             /d:sonar.verbose=false \
             /k:"commonlibrary_dev" \
+            /n:"Azure Common Library" \
```
FuriousLlama, line 33: Not sure this is 100% the name seems to be set on the sonarq site

PR #626, opened 2020-07-08
Fixes note being added to the wrong date

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8646](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8646)
- [ ] Enhancement
- [x] Bug
- [ ] Documentation

## Description:
Fixes the note being added to the wrong date.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required

### Steps to Reproduce:
Add a new note
After the note is saved it is added to the timeline with a wrong date.
A refresh of the page fixes the issue.

### UI Changes
NO

### DIFF
--- a/Apps/WebClient/src/Server/Models/UserNote.cs
+++ b/Apps/WebClient/src/Server/Models/UserNote.cs
@@ -96,7 +96,7 @@ namespace HealthGateway.WebClient.Models
             {
                 Id = model.Id,
                 HdId = model.HdId,
-                JournalDateTime = model.JournalDateTime,
+                JournalDateTime = DateTime.SpecifyKind(model.JournalDateTime, DateTimeKind.Unspecified),
                 Version = model.Version,
                 CreatedDateTime = model.CreatedDateTime,
                 CreatedBy = model.CreatedBy,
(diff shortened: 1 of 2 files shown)


Apps/WebClient/src/Server/Models/UserNote.cs
```diff
@@ -96,7 +96,7 @@ public static UserNote CreateFromDbModel(Database.Models.Note model, ICryptoDele
             {
                 Id = model.Id,
                 HdId = model.HdId,
-                JournalDateTime = model.JournalDateTime,
+                JournalDateTime = DateTime.SpecifyKind(model.JournalDateTime, DateTimeKind.Unspecified),
```
FuriousLlama, line 99: Is this effective because after creation it is fetched back to the frontend?

tiagograf-AOT: When it comes from the front end the kind comes as UTC, while when it loads from the database it comes as Unspecified, when its in UTC the frontend converts it to local time and messes up with the date.

PR #628, opened 2020-07-09
Implementation of Vue, Vue Service and Server User Profile/Preference Controller for the Actions menu (Add a Note) Description Popover

leobcgov:
## Status
**Ready**

## Fixes or Implements [[AB#8631](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8631)
](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8631
) and 8628
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
 - Implementation of Vue, Vue Service and Server User Profile/Preference Controller for the Actions menu (Add a Note) except showing vue on the UI.

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### Steps to Reproduce:

### UI Changes
NO

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

## Notes/Additional Comments
- This current implementation doesn't include mobile. Leo will create a task for mobile version.
- This PR includes also some updates on WebClient Server's Delegate to fix the build issue.


Apps/WebClient/src/ClientApp/src/components/navmenu/sidebar.vue
```diff
@@ -416,12 +460,14 @@ import { WebClientConfiguration } from "@/models/configData";
...
 import { library } from "@fortawesome/fontawesome-svg-core";
 import { faStream } from "@fortawesome/free-solid-svg-icons";
+import { BPopover } from "bootstrap-vue";
 library.add(faStream);
 
 const auth: string = "auth";
 const user: string = "user";
 const sidebar: string = "sidebar";
 
+Vue.component("b-popover", BPopover);
```
FuriousLlama, line 470: lets add it to the rest of the bootstrap objects in the same location as the rest. I believe its in main.ts or app.vue

Apps/WebClient/src/ClientApp/src/components/navmenu/sidebar.vue
```diff
@@ -512,19 +564,45 @@ export default class SidebarComponent extends Vue {
...
+        if (this.dismissedMyNotePopover === true) {
+            (this.$refs.popover as Vue).$emit("close");
+        } else {
+            (this.$refs.popover as Vue).$emit("open");
+        }
+    }
+
+    private loadUserPreference(): void {
+        console.log(
+            "Loading user preference (including dismissedMyNotePopover)..."
```
FuriousLlama, line 593: nit "Loading user preference" is sufficient. Or not log it at all since it will go to prod with this

Apps/WebClient/src/ClientApp/src/models/userPreference.ts
```diff
@@ -0,0 +1,8 @@
+export default interface UserPreference {
+    // Gets or sets the user preference id.
+    id?: string;
+    // Gets or sets the users directed identifier.
+    hdId: string;
+    // Gets or sets a value indicating if the user dismissed the note notification.
+    dismissedMyNotePopover: boolean;
```
FuriousLlama, line 7: What do you think of tutorialPopover instead? Othwerwise we will need one for every single popover

Apps/WebClient/src/ClientApp/src/services/interfaces.ts
```diff
@@ -20,6 +20,7 @@ import UserNote from "@/models/userNote";
 import Communication from "@/models/communication";
 import { LaboratoryOrder, LaboratoryReport } from "@/models/laboratory";
 import UserSMSInvite from "@/models/userSMSInvite";
+import UserPreference from "../models/userPreference";
```
FuriousLlama, line 23: nit: replace .. with @

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
```diff
@@ -7,6 +7,7 @@ import { RootState, UserState } from "@/models/storeState";
 import PatientData from "@/models/patientData";
 import UserEmailInvite from "@/models/userEmailInvite";
 import UserSMSInvite from "@/models/userSMSInvite";
+import UserPreference from "../../../models/userPreference";
```
FuriousLlama, line 10: replace with @

Apps/WebClient/src/ClientApp/src/views/profile.vue
```diff
@@ -432,6 +432,7 @@ import {
     faExclamationTriangle,
 } from "@fortawesome/free-solid-svg-icons";
 import moment from "moment";
+import UserPreference from "../models/userPreference";
```
FuriousLlama, line 435: same as above

Apps/WebClient/src/ClientApp/src/views/profile.vue
```diff
@@ -527,6 +539,9 @@ export default class ProfileView extends Vue {
         var oidcUserPromise = authenticationService.getOidcUserProfile();
         var userEmailPromise = this.getUserEmail({ hdid: this.user.hdid });
         var userSMSPromise = this.getUserSMS({ hdid: this.user.hdid });
+        var userPreferencePromise = this.getUserPreference({
+            hdid: this.user.hdid,
+        });
         var userProfilePromise = this.userProfileService.getProfile(
```
FuriousLlama, line 545: since the profile contains the user information that we have, lets move the preferences to the same object. That way a single call will load both

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
```diff
@@ -320,5 +320,48 @@ public async Task<IActionResult> UpdateUserSMSNumber(string hdid, [FromBody] str
...
+        /// </summary>
+        /// <returns>The user preference model wrapped in a request result.</returns>
+        /// <param name="hdid">The user hdid.</param>
+        /// <response code="200">Returns the user preference json.</response>
+        /// <response code="401">the client must authenticate itself to get the requested response.</response>
+        /// <response code="403">The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401, the client's identity is known to the server.</response>
+        [HttpGet]
+        [Route("{hdid}/userpreference")]
+        [Authorize(Policy = UserPolicy.Read)]
+        public IActionResult GetUserPreference(string hdid)
```
FuriousLlama, line 335: Although it could be a stand alone resource, it looks to me that every time we need the user preference we will use the user profile

FuriousLlama:
Reviewing smaller reviews is usually preferred. Thats one of the reasons why they are split in tasks (I am also bad at this :))

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
leobcgov: right, for me, user preference should be a part of user profile. also, I was thinking to have them all in 1 call between client and server.
Are you suggesting we should separate the user preference from user profile now?

leobcgov: nmv, I have just seen your "lets move the preferences to the same object. That way a single call will load both" above then.

Apps/WebClient/src/ClientApp/src/models/userPreference.ts
leobcgov: can you expand a bit further? do you mean using a custom tutorialPopover class instead of dismissedMyNotePopover: boolean?

Apps/WebClient/src/ClientApp/src/components/navmenu/sidebar.vue
```diff
@@ -540,6 +618,22 @@ export default class SidebarComponent extends Vue {
         EventBus.$emit("timelineCreateNote");
     }
 
+    private dismissNoteNotification() {
+        console.log("Dismissing Note Notification...");
+        this.userProfileService
+            .createUserPreference({
```
FuriousLlama, line 624: Looking deeper this looks like it will attempt to insert a new user preferences. Instead I believe we need to update them. Assuming from the front end that the preferences already existed.

Apps/WebClient/src/ClientApp/src/views/profile.vue
leobcgov: I'm refactoring this now.

Apps/WebClient/src/ClientApp/src/components/navmenu/sidebar.vue
leobcgov: for this requirement, once user dismisses this note popover, this notification will never display again. Next time when user comes back, this flag will be true as we grab its value for db right?

Apps/WebClient/src/ClientApp/src/models/userPreference.ts
FuriousLlama: No I think a boolean is fine. The issue is that the name of the variable implies there can be more popover preferences such as dissmissedPrintPopover. Instead make it general for all popovers

Apps/WebClient/src/ClientApp/src/components/navmenu/sidebar.vue
FuriousLlama: Right, but thats leaking implementation details to the front end. The webclient does not need to know how the state of the preference is set since it only needs to read it and update it. Having it being created or updated is a choice that the service or delegate will do.

Apps/Database/src/Delegates/DBUserPreferenceDelegate.cs
```diff
@@ -0,0 +1,76 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text.Json;
+    using HealthGateway.Database.Constants;
+    using HealthGateway.Database.Context;
+    using HealthGateway.Database.Models;
+    using HealthGateway.Database.Wrapper;
+    using Microsoft.EntityFrameworkCore;
+    using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
```
FuriousLlama, line 27: Ran it locally. Looks like there are some using definitions that are not needed.

Apps/WebClient/src/ClientApp/src/components/navmenu/sidebar.vue
leobcgov: correct, normally this should be implemented on the Db Delegate level right? Or should we do the check at the Server Service level as well?

leobcgov: btw, I recall in other stories, we do the check on DBDelegate, if it is true, we can confirm the check as a part of 8626: create user preference delegate ?

sam-warren:
Lots of errors in the failed build - might want to take a look: https://qslvic.visualstudio.com/Health%20Gateway/_build/results?buildId=7065&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=2b20c7d0-7587-5b0f-aadc-b456e6f0b807&l=57

Apps/Database/src/Delegates/DBUserPreferenceDelegate.cs
```diff
@@ -66,11 +61,11 @@ public DBResult<UserPreference> GetUserPreference(string hdid)
                 Payload = new UserPreference()
                 {
                     HdId = hdid,
-                    DismissedMyNotePopover = false,
+                    TutorialPopover = false,
                     Id = Guid.NewGuid(),
```
sslaws, line 65: There is an outstanding PR from Tiago with changes requested and this goes against those changes.

PR #630, opened 2020-07-09
Add UserPreference model, migrations and delegate

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8625](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8625)
- [x] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
Adds user preference db model, migrations and db delegate.

## Testing
- [x] Unit Tests Updated
- [ ] Not Required

### UI Changes
 NO


Apps/WebClient/test/unit/Services.Test/UserProfileService_Test.cs
```diff
@@ -268,16 +270,18 @@ public void ShouldGetPreference()
             UserPreference dbUserPreference = new UserPreference
             {
                 HdId = hdid,
-                DismissedMyNotePopover = true,
+                Key = "DismissedMyNotePopover",
```
FuriousLlama, line 273: Lets rename it to tutorialPopover to make it general

Apps/Database/src/Models/UserPreference.cs
```diff
@@ -38,8 +32,14 @@ public class UserPreference
         public string HdId { get; set; } = null!;
 
         /// <summary>
-        /// Gets or sets a value indicating whether the user dismissed the note popover.
+        /// Gets or sets the preference key.
+        /// </summary>
+        [Key]
+        public string Key { get; set; } = null!;
```
sslaws, line 38: I'm not sure why we defined the key as the key and in the notion of the database I don't like the name of key (Maybe Preference and value).  I think this table should be hdid/preference as a composite key.

Apps/WebClient/src/Server/Models/UserPreferenceModel.cs
```diff
@@ -16,17 +16,14 @@
 namespace HealthGateway.WebClient.Models
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
 
     /// <summary>
     /// Model that provides a user representation of an user preference database model.
     /// </summary>
     public class UserPreferenceModel
```
sslaws, line 25: Why are we creating a front end model?  I'm not seeing a lot of value.

leobcgov: @sslaws do you mean we can just expose the dbmodel which contains the Preference and Value pair in the Service & Controller right?

Apps/Database/src/Models/UserPreference.cs
tiagograf-AOT: Updated

Apps/WebClient/test/unit/Services.Test/UserProfileService_Test.cs
tiagograf-AOT: Updated

PR #633, opened 2020-07-10
Fixes reload of note after save

tiagograf-AOT:
## Status
**Ready**

## Fixes or Implements [AB#8671](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8671)
- [ ] Enhancement
- [x] Bug
- [ ] Documentation

## Description:
Fixes the reload of the note object after a save to the database.

## Testing
- [ ] Unit Tests Updated
- [x] Not Required

### DIFF
--- a/Apps/Database/src/Delegates/DBNoteDelegate.cs
+++ b/Apps/Database/src/Delegates/DBNoteDelegate.cs
@@ -47,13 +47,13 @@ namespace HealthGateway.Database.Delegates
         }
 
         /// <inheritdoc />
-        public DBResult<Note> GetNote(Guid noteId)
+        public DBResult<Note> GetNote(Guid noteId, string hdid)
         {
             DBResult<Note> result = new DBResult<Note>()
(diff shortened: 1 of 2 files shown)


Apps/Database/src/Delegates/DBNoteDelegate.cs
```diff
@@ -47,13 +47,13 @@ public class DBNoteDelegate : INoteDelegate
         }
 
         /// <inheritdoc />
-        public DBResult<Note> GetNote(Guid noteId)
+        public DBResult<Note> GetNote(Guid noteId, string hdid)
```
FuriousLlama, line 50: Is the hdid necessary?

tiagograf-AOT: We made the notes table a composite key with ID + HDID, so we do need it!

PR #638, opened 2020-07-14
Popover wiring and fixes

FuriousLlama:
## Status
**Ready**

## Fixes or Implements [AB#8629](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8629)
- [X] Enhancement
- [X] Bug
- [ ] Documentation

## Description:
Wired front end and back

- end user preferences. 
- Updated front end model Fixed position of the popover.
- Removed backend methods not called

## Testing
- [X] Unit Tests Updated
- [ ] Not Required

### Steps to Reproduce:
Open the timeline
   - Popover should be visible.
Minimize sidebar
   - Popover should be visible and in the correct position.
Open the Timeline in a mobile window.
   - Popover should not be visible
Open the side menu
   - Popover should be visible and in the correct position and size

### UI Changes
YES

### Browsers Tested
- [x] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox


Apps/WebClient/src/ClientApp/src/services/interfaces.ts
```diff
@@ -91,10 +90,11 @@ export interface IUserProfileService {
...
     updateEmail(hdid: string, email: string): Promise<boolean>;
     updateSMSNumber(hdid: string, smsNumber: string): Promise<boolean>;
-    getUserPreference(hdid: string): Promise<UserPreference>;
-    createUserPreference(
-        userPreference: UserPreference
-    ): Promise<UserPreference>;
+    updateUserPreference(
+        hdid: string,
+        preference: string,
+        value: string
```
sslaws, line 96: We don't have to update yet, but we should always be pulling the version number.

Apps/WebClient/src/Server/Services/IUserProfileService.cs
```diff
@@ -72,13 +73,13 @@ public interface IUserProfileService
         /// </summary>
         /// <param name="userPreference">The userPreference to create.</param>
         /// <returns>A userPreference wrapped in a RequestResult.</returns>
-        RequestResult<UserPreferenceModel> CreateUserPreference(UserPreferenceModel userPreference);
+        bool UpdateUserPreference(string hdid, string name, string value);
```
sslaws, line 76: Updates should only be run against objects as the version number is required.

Apps/WebClient/src/ClientApp/src/services/interfaces.ts
FuriousLlama: preferences are not represented as an object in the front end since they are a dictionary. with that in mind This update is also a create if you look at the delegate.

Apps/WebClient/src/Server/Services/IUserProfileService.cs
FuriousLlama: The delegate method does not have a create method either

PR #654, opened 2020-07-19
Implemented Communication job

leobcgov:
## Status
**Ready**

## Fixes or Implements [AB#8638](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8638)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
- Added Communication Job which is scheduled for every 30 minutes to create Emails which need to be sent to the Gateway's active users

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### Steps to Reproduce:

### UI Changes
NO

### Browsers Tested
- [ ] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

## Notes/Additional Comments
- Testing performed locally.
- Migration Script tested:
image


Apps/Database/src/Constants/EmailConsts.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Represents the To of emails.
+    /// </summary>
+    public static class EmailConsts
+    {
+        /// <summary>
+        /// Constant value to represent undisclosed-recipients of the To field of the email.
+        /// </summary>
+        public const string FromEmailAddressHGDonotreply = "HG_Donotreply@gov.bc.ca";
```
sslaws, line 26: This should be configuration vs a constant as it may change in the future.

Apps/Database/src/Models/Communication.cs
```diff
@@ -74,5 +75,18 @@ public class Communication : AuditableEntity
         /// The lower the value the lower the priority.
         /// </summary>
         public int Priority { get; set; } = EmailPriority.Standard;
+
+
+#pragma warning disable CA2227 // Collection properties should be read only
```
sslaws, line 80: We should be fixing these issues instead of using a pragma to disable them

Apps/Database/src/Models/Email.cs
```diff
@@ -99,5 +100,17 @@ public class Email : AuditableEntity
         [Required]
         [MaxLength(10)]
         public string EmailStatusCode { get; set; } = EmailStatus.New;
+
+#pragma warning disable CA2227 // Collection properties should be read only
```
sslaws, line 104: ditto on the pragmas from earlier

Apps/Database/src/Models/UserProfile.cs
```diff
@@ -72,5 +73,17 @@ public class UserProfile : AuditableEntity
         /// </summary>
         [MaxLength(44)]
         public string? EncryptionKey { get; set; }
+
+#pragma warning disable CA2227 // Collection properties should be read only
```
sslaws, line 77: and again

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
```diff
@@ -0,0 +1,116 @@
...
+
+                hdidsWithMostRecentCreatedCommEmails = hdids.ToString();
+            }
+
+            var userProfiles = this.dbContext.UserProfile
+                .Where(profile => !profile.ClosedDateTime.HasValue
+                        && !string.IsNullOrEmpty(profile.Email)
+                        && (!createdOnOrAfter.HasValue || profile.CreatedDateTime >= createdOnOrAfter)
+
+#pragma warning disable CA1307 // Specify StringComparison
```
sslaws, line 105: NIT:  I like using annotations to disable over pragmas as they look cleaner in the code.

leobcgov:
@sslaws my latest check-in contains the fixes re. your comments, can you please review them? thanks!

Apps/Database/src/Models/CommunicationEmail.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        [Column("CommunicationEmailId")]
+        public Guid Id { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Communication Id (foreign key).
+        /// </summary>
+        [ForeignKey("Communication")]
```
sslaws, line 38: Kudo's as I don't think anyone has done this before and it saves custom code in the context.

Apps/Database/src/Models/UserProfile.cs
```diff
@@ -72,5 +73,10 @@ public class UserProfile : AuditableEntity
         /// </summary>
         [MaxLength(44)]
         public string? EncryptionKey { get; set; }
+
+        /// <summary>
+        /// Gets the list containing all the communication emails related to this user profile.
+        /// </summary>
+        public virtual List<CommunicationEmail> CommunicationEmails { get; } = null!;
```
sslaws, line 80: We did this a lot early on but we've been leaning against it lately and instead have been creating empty lists.  We should discuss during grooming.

leobcgov: Oh I didn't know that. One benefit of this is ability to include the list or a filtered list of child records in the LINQ query (i.e. .Include(...)

leobcgov: updated to "public virtual List<CommunicationEmail> CommunicationEmails { get; } = new List<CommunicationEmail>();"

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
```diff
@@ -0,0 +1,112 @@
...
+                StringBuilder hdids = new StringBuilder();
+                foreach (var commEmail in commEmails)
+                {
+                    hdids.Append(commEmail.UserProfileHdId + "|");
+                }
+
+                hdidsWithMostRecentCreatedCommEmails = hdids.ToString();
+            }
+
+            var userProfiles = this.dbContext.UserProfile
```
sslaws, line 100: I think we could likely optimize this query. Can you create a task in azure for the future?

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,147 @@
...
+                        {
+                            if (usersToSendCommEmails != null && usersToSendCommEmails.Count > 0)
+                            {
+                                lastProcessedProfileHdid = usersToSendCommEmails[usersToSendCommEmails.Count - 1].HdId;
+                            }
+
+                            usersToSendCommEmails = this.commEmailDelegate.GetActiveUserProfilesWithoutCommEmailByCommunicationIdAndByCreatedOnOrAfter(comm.Id, createdOnOrAfterFilter, this.retryFetchSize);
+                            foreach (UserProfile profile in usersToSendCommEmails)
+                            {
+                                // Idea for Improvement: we can look into a lib for Bulk Insert / Update using .NET Core for Postgres (if needed).
```
sslaws, line 99: Yes, bulk insert operation would be faster.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,147 @@
...
+                                this.commEmailDelegate.Add(commEmail, false);
+                            }
+
+                            if (usersToSendCommEmails.Count > 0)
+                            {
+                                // Set the createdOnOrAfterFilter for the next query retrieving the next chunk of user profiles.
+                                createdOnOrAfterFilter = usersToSendCommEmails[usersToSendCommEmails.Count - 1].CreatedDateTime;
+                            }
+
+                            this.dbContext.SaveChanges(); // commit after every retryFetchSize (or 250) pairs of Email & CommunicationEmail.
```
sslaws, line 128: Now that I see this here, I see that we’re missing a state... we should flip the state to processing or something similar when we start and the outer query should look for new and processing’s comms.

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
leobcgov: Created 8717 for this improvement.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
leobcgov: I will add a new "Processing" status to the Communication's Communication Status Code and update it when processing a communication in the job scheduler.

leobcgov: Code updated per @sslaws 's great comments above.

Apps/Database/src/Context/GatewayDbContext.cs
```diff
@@ -687,6 +688,15 @@ private void SeedCommunication(ModelBuilder modelBuilder)
                     UpdatedDateTime = this.DefaultSeedDate,
                 },
                 new CommunicationStatusCode
+                {
+                    StatusCode = CommunicationStatus.Processing,
+                    Description = "A Communication is in process by batch",
```
FuriousLlama, line 693: nit Communication is being processed

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
```diff
@@ -132,5 +133,15 @@ public DBResult<Communication> Update(Communication communication, bool commit =
             this.logger.LogDebug($"Finished updating Communication in DB");
             return result;
         }
+
+        /// <inheritdoc />
+        public List<Communication> GetEmailCommunicationsInNewProcessingOrError()
+        {
+            this.logger.LogTrace($"Getting Communications by Type and Status Code from DB...");
+            List<Communication> retVal = this.dbContext.Communication.Where(c => c.CommunicationTypeCode == CommunicationType.Email && (c.CommunicationStatusCode == CommunicationStatus.New || c.CommunicationStatusCode == CommunicationStatus.Processing || c.CommunicationStatusCode == CommunicationStatus.Error))
```
FuriousLlama, line 141: Instead, we can search for not processed.

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
```diff
@@ -132,5 +133,15 @@ public DBResult<Communication> Update(Communication communication, bool commit =
             this.logger.LogDebug($"Finished updating Communication in DB");
             return result;
         }
+
+        /// <inheritdoc />
+        public List<Communication> GetEmailCommunicationsInNewProcessingOrError()
```
FuriousLlama, line 138: should we not get only the communications for today or past its scheduled date?

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,167 @@
...
+        /// <inheritdoc />
+        [DisableConcurrentExecution(ConcurrencyTimeout)]
+        public void CreateCommunicationEmailsForNewCommunications()
+        {
+            this.logger.LogDebug($"Creating emails & communication emails...");
+            List<Communication> communications = this.communicationDelegate.GetEmailCommunicationsInNewProcessingOrError();
+
+            if (communications.Count > 0)
+            {
+                this.logger.LogInformation($"Found {communications.Count} communications which are in New status.");
```
FuriousLlama, line 79: Not necessarily true, the query gets other statuses as well

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,167 @@
...
+                                    UserProfile = profile,
+                                    Communication = comm,
+                                };
+                                this.commEmailDelegate.Add(commEmail, false);
+                            }
+
+                            if (usersToSendCommEmails.Count > 0)
+                            {
+                                // Set the createdOnOrAfterFilter for the next query retrieving the next chunk of user profiles.
+                                createdOnOrAfterFilter = usersToSendCommEmails[usersToSendCommEmails.Count - 1].CreatedDateTime;
```
FuriousLlama, line 124: why is this necessary? Does the order of the profiles matter?

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,167 @@
...
+            Contract.Requires((configuration != null) && (emailQueueService != null));
+            this.configuration = configuration!;
+            this.logger = logger;
+            this.communicationDelegate = communicationDelegate!;
+            this.commEmailDelegate = commEmailDelegate!;
+            this.emailQueueService = emailQueueService!;
+            this.dbContext = dbContext;
+
+            IConfigurationSection emailJobSection = this.configuration.GetSection("EmailJob");
+            this.retryFetchSize = emailJobSection.GetValue<int>("MaxRetryFetchSize", 250);
```
FuriousLlama, line 65: retryFetchSize is confusion. How about "ProcessMaxBatchSize"

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
```diff
@@ -0,0 +1,112 @@
...
+                .Where(profile => !profile.ClosedDateTime.HasValue
+                        && !string.IsNullOrEmpty(profile.Email)
+                        && (!createdOnOrAfter.HasValue || profile.CreatedDateTime >= createdOnOrAfter)
+                        && hdidsWithMostRecentCreatedCommEmails != profile.HdId)
```
FuriousLlama, line 104: Not sure we need the hdidsWithMostRecentCreatedCommEmails parameter. Instead, the query can select for profiles that do not have a communication entry

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
```diff
@@ -0,0 +1,112 @@
...
+            var commEmails = this.dbContext.CommunicationEmail.Where(c => c.CommunicationId == communicationId).OrderByDescending(c => c.CreatedDateTime).Take(maxRows)
+                    .Include(commEmail => commEmail.UserProfile).Where(c => !createdOnOrAfter.HasValue || c.UserProfile.CreatedDateTime >= createdOnOrAfter).ToList();
+            string hdidsWithMostRecentCreatedCommEmails = string.Empty;
+            if (commEmails.Count > 0)
+            {
+                createdOnOrAfter = commEmails[0].UserProfile.CreatedDateTime;
+                StringBuilder hdids = new StringBuilder();
+                foreach (var commEmail in commEmails)
+                {
+                    hdids.Append(commEmail.UserProfileHdId + "|");
```
FuriousLlama, line 94: why is the concatenation necessary?

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
```diff
@@ -0,0 +1,112 @@
...
+                    result.Message = e.Message;
+                }
+            }
+
+            this.logger.LogDebug($"Finished adding Communication in DB");
+            return result;
+        }
+
+        /// <inheritdoc />
+        public List<UserProfile> GetActiveUserProfilesWithoutCommEmailByCommunicationIdAndByCreatedOnOrAfter(Guid communicationId, DateTime? createdOnOrAfter = null, int maxRows = 500)
```
FuriousLlama, line 80: kinda nit: 76 chars = Overly long method name?

Apps/Database/src/Delegates/ICommunicationEmailDelegate.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Get a list of active user profiles which are not linked to any communication email by the specified Communication Id
+        ///     AND were created on of after a specific date and time .
+        /// </summary>
+        /// <param name="communicationId">The communication id for filtering communication emails.</param>
+        /// <param name="createdOnOrAfter">The profiles must have created on or after to this date.</param>
+        /// <param name="maxRows">The maximum amount of user profiles to return.</param>
+        /// <returns>A list of communication emails of the specified Communication, wrapped in a DBResult.</returns>
+        List<UserProfile> GetActiveUserProfilesWithoutCommEmailByCommunicationIdAndByCreatedOnOrAfter(Guid communicationId, DateTime? createdOnOrAfter = null, int maxRows = 500);
```
FuriousLlama, line 44: This looks like an anti-pattern. The method is only being used in one place but has two optional parameters

Apps/Database/src/Models/Communication.cs
```diff
@@ -74,5 +75,10 @@ public class Communication : AuditableEntity
         /// The lower the value the lower the priority.
         /// </summary>
         public int Priority { get; set; } = EmailPriority.Standard;
+
+        /// <summary>
+        /// Gets the list containing all the communication emails related to this communication.
+        /// </summary>
+        public virtual List<CommunicationEmail> CommunicationEmails { get; } = null!;
```
FuriousLlama, line 82: ```= new List<CommunicationEmail>;```

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,167 @@
...
+                                    Body = comm.Text,
+                                    FormatCode = EmailFormat.HTML,
+                                    Priority = comm.Priority,
+                                };
+                                this.emailQueueService.QueueNewEmail(email, false);
+
+                                // Inser a new CommunicationEmail record into
+                                CommunicationEmail commEmail = new CommunicationEmail()
+                                {
+                                    Email = email,
```
FuriousLlama, line 114: This email object does not have an ID. Does that work when adding it to the table on line 118?

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,167 @@
...
+        public void CreateCommunicationEmailsForNewCommunications()
+        {
+            this.logger.LogDebug($"Creating emails & communication emails...");
+            List<Communication> communications = this.communicationDelegate.GetEmailCommunicationsInNewProcessingOrError();
+
+            if (communications.Count > 0)
+            {
+                this.logger.LogInformation($"Found {communications.Count} communications which are in New status.");
+
+                foreach (Communication comm in communications)
```
FuriousLlama, line 81: nit: comm => communication

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
sslaws: It is really just fetch size, we're pulling a block to process and the majority of the time the block should be processed without retrying.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
sslaws: Yes that works, the DB creates the ID and EF gets the ID when it inserts it and then uses it for inserts down the line.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
sslaws: The order is being used for processing blocks as the job will read from the last date read.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,167 @@
...
+                        // Update Communication Status to Processed
+                        comm.CommunicationStatusCode = CommunicationStatus.Processed;
+                        this.communicationDelegate.Update(comm, true);
+                    }
+                    catch (Exception e)
+                    {
+                        // log the exception as a warning but we can continue
+                        this.logger.LogWarning($"Error while creating new emails and communication email records for Email Communication - skipping for now\n{e.ToString()}");
+
+                        if (comm.CommunicationStatusCode != CommunicationStatus.Processed)
```
sslaws, line 147: This if is unlikely to be ever true (although possible) as the only statements likely to cause an exception are SaveChanges and Update(comm, true).  It is more likely that SaveChanges fails and if that occurs then the update to the comm to the error state is also unlikely to occur.  Hence the reason to set the flag to processing at the start of the loop.

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
leobcgov: I think in general explicitly list all the statuses that need to be returned/processed is better as what if we introduce a new status in the future that we don't need to return here? Also what about "Pending" communication even we don't have any plan to use this "Pending" status.

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
leobcgov: Do you see any scenarios this can cause issue if we return only the communications based on today and scheduled date? Also for info, schedule date has not been introduced yet.

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
leobcgov: Per MS guidance, method name can be up to 128 length. Method name should describe what it filters and returns. Btw, I will try to come up with a shorter version.
<https://github.com/ktaranov/naming-convention/blob/master/C%23%20Coding%20Standards%20and%20Naming%20Conventions.md>

leobcgov: Updating it to "GetActiveUserProfilesByCommunicationId"

Apps/Database/src/Delegates/ICommunicationEmailDelegate.cs
leobcgov: Updating it to "public List<UserProfile> GetActiveUserProfilesByCommunicationId(Guid communicationId, DateTime? createdOnOrAfter, int maxRows)"

Apps/Database/src/Models/Communication.cs
leobcgov: this is an old version, not sure why you can still see it there.

Apps/Database/src/Delegates/DBCommunicationDelegate.cs
FuriousLlama: Since this is retrieving communications to send I would just name it that. Something like "GetEmailCommunicationsToSend"

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
leobcgov: Yes, I have tested this as well, It works fine. As @sslaws mentioned, EF core knows to track these objects.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
FuriousLlama: I think the query that is getting the profiles would resolve this successfully without having this additional variable. Unless we want them to be processed by creation date I think we should remove it.

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
FuriousLlama: I like fetch size as a name

FuriousLlama: On that same line, we do need a max retry counter. Otherwise, the loop might not end

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
leobcgov: Will update it, btw personally I prefer using short name for these in the loop, i.e. foreach(Communication c...) or forearch(Blog b ...). This short name helps the code to be easier to read as It is super hard to read "communication" single object vs. "communications" (list)

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,166 @@
...
+                this.logger.LogInformation($"Found {communications.Count} communications which need to be processed / re-processed.");
+
+                foreach (Communication communication in communications)
+                {
+#pragma warning disable CA1031 //We want to catch exception.
+                    try
+                    {
+                        DateTime? createdOnOrAfterFilter = null;
+                        string lastProcessedProfileHdid = string.Empty;
+                        List<UserProfile>? usersToSendCommEmails = null;
```
FuriousLlama, line 87: Instead of making it nullable how about initializing it as empty. That way the if below can be simplified

leobcgov:
Code updated. Thanks a lot @FuriousLlama and @sslaws for your great comments and feedbacks. We may need to review some of them on a call today to make sure that I have not missed anything.

Apps/Database/src/Delegates/DBCommunicationEmailDelegate.cs
leobcgov: Leo will look into the approach of using a single LINQ instead of 3 steps here

Apps/JobScheduler/src/Jobs/CommunicationJob.cs
```diff
@@ -0,0 +1,166 @@
...
+                {
+#pragma warning disable CA1031 //We want to catch exception.
+                    try
+                    {
+                        DateTime? createdOnOrAfterFilter = null;
+                        string lastProcessedProfileHdid = string.Empty;
+                        List<UserProfile> usersToSendCommEmails = new List<UserProfile>();
+                        do
+                        {
+                            if (usersToSendCommEmails != null && usersToSendCommEmails.Count > 0)
```
FuriousLlama, line 90: usersToSendCommEmails cannot be null

leobcgov:
Close this PR to implement a single LINQ query to get the list of User Profiles to create new Communication Email records.

PR #656, opened 2020-07-21
Implemented "Change page on Timeline vue when user switches month using left-right navigation on the calendar vue"

leobcgov:
## Status
**Ready**

## Fixes or Implements [AB#8691](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8691)
- [X] Enhancement
- [ ] Bug
- [ ] Documentation

## Description:
- Switching month by using left and right arrows on the Calendar vue should update the current page of the linear TimeLine List vue.
- This needs to be tested when Month / Year selection controls are implemented, current, they are in progress as a part of 8603. I assume the notification should be fired the same way in both scenarios including user uses left/right arrows or Month Year Navigation.

## Testing
- [ ] Unit Tests Updated
- [X] Not Required

### Steps to Reproduce:

### UI Changes
Yes

### Browsers Tested
- [X] Chrome
- [ ] Safari
- [ ] Edge
- [ ] Firefox

## Notes/Additional Comments
Any addtional notes or comments that may be relevant.  Include any specialized deployment steps here.


Apps/WebClient/src/ClientApp/src/components/calendar/header.vue
```diff
@@ -105,6 +108,10 @@ export default class CalendarComponent extends Vue {
     private dispatchEvent() {
         let startDate = DateUtil.getMonthFirstDate(this.headerDate);
         this.$emit("update:currentDate", startDate);
+        this.eventBus.$emit(
+            EventMessageName.TimelineCurrentDateUpdated,
+            this.currentDate
+        ); // emit the timelineCurrentDateUpdated event to the linearTImeline vue sibling
```
FuriousLlama, line 114: nit: Code is explicit enough, no need for comment

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -103,6 +104,14 @@ export default class LinearTimelineComponent extends Vue {
             self.setPageFromDate(eventDate);
         });
 
+        this.eventBus.$on(
+            EventMessageName.TimelineCurrentDateUpdated,
+            function (currentDate: Date) {
+                self.setPageFromDate(currentDate);
```
FuriousLlama, line 110: this will probably not work since the date is not guaranteed to exist on that month.

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -103,6 +104,14 @@ export default class LinearTimelineComponent extends Vue {
...
+                // self.currentPage = 3; // debug only, to be updated when Leo works on 8693
```
FuriousLlama, line 111: is this the ticket?

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
FuriousLlama: More specifically below will give errors
```
let self = this;
        // Wait for next render cycle until the pages have been calculated and displayed
        this.$nextTick().then(function () {
            const date = new Date(eventDate).setHours(0, 0, 0, 0);
            let container: HTMLElement[] = self.$refs[date] as HTMLElement[];
            container[0].focus();
        });
```

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: 8693 is the next one that I will work on. This PR is pretty much to notify the linearTimeLine from Calendar vue

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: I have removed these temp codes which need to be cleaned up as a part of the 8693

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
FuriousLlama: I think this ticket about implementing the switch between month view to timeline view. 8693 is about timeline view to calendar view.

Apps/WebClient/src/ClientApp/src/components/calendar/header.vue
```diff
@@ -105,6 +108,10 @@ export default class CalendarComponent extends Vue {
     private dispatchEvent() {
         let startDate = DateUtil.getMonthFirstDate(this.headerDate);
         this.$emit("update:currentDate", startDate);
+        this.eventBus.$emit(
```
sam-warren, line 111: Can we use @Emit() for this similar to how we emit data from a child component in other parts of the app? See the communication modals for example.

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: right, I though the 8693 is supposed to to both. I will implement code for this

Apps/WebClient/src/ClientApp/src/components/calendar/header.vue
leobcgov: for info, this.eventBus.$emit needs to be used for sending communication between sibling (i.e. Calendar and Timeline list in this scenario). this.$emit is used for child vue to communicate with its parent.

Apps/WebClient/src/ClientApp/src/components/calendar/header.vue
```diff
@@ -78,26 +80,21 @@ export default class CalendarComponent extends Vue {
...
     private headerDate: Date = new Date();
+    private eventBus = EventBus;
     private leftIcon: string = "chevron-left";
     private rightIcon: string = "chevron-right";
 
     public get title(): string {
         return moment(this.currentDate).format(this.titleFormat);
     }
 
-    @Watch("availableMonths")
```
FuriousLlama, line 88: without this, the navigation might move to months that don't have data. Can you make sure the month toggle only moves to months that have data?

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -120,10 +121,54 @@ export default class LinearTimelineComponent extends Vue {
...
 
+    private goToPageByCurrentDate(currentDate: Date) {
+        const selectedYearMonth = moment(currentDate).format("YYYY-MM");
+        let currentPage = 1;
+        let currentEntryIndex = 1;
+        let i = 0;
+        let foundLastEntryOfSelectedMonth = false;
+        const timelineEntriesLenght = this.timelineEntries.length;
+        // scan from the most recent entries to the older ones
+        while (i < timelineEntriesLenght && !foundLastEntryOfSelectedMonth) {
```
FuriousLlama, line 143: Instead of using a while with an index, a for is preferred

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -120,10 +121,54 @@ export default class LinearTimelineComponent extends Vue {
...
+                    currentEntryIndex = 1;
+                    currentPage++;
+                }
+            }
+            i++;
+        }
+        if (foundLastEntryOfSelectedMonth) {
+            // If found entry is at the bottom of the to-go page, move to the next page
+            if (
+                currentEntryIndex > this.numberOfEntriesPerPage - 3 &&
```
FuriousLlama, line 163: we don't know if there are -3 entries. This could cause an index out of bounds

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -120,10 +121,54 @@ export default class LinearTimelineComponent extends Vue {
...
+        const timelineEntriesLenght = this.timelineEntries.length;
+        // scan from the most recent entries to the older ones
+        while (i < timelineEntriesLenght && !foundLastEntryOfSelectedMonth) {
+            let entry = this.timelineEntries[i];
+            if (
+                entry.date !== undefined &&
+                entry.date.toString().indexOf(selectedYearMonth) == 0
+            ) {
+                console.log("found the last entry of the selected month");
+                foundLastEntryOfSelectedMonth = true;
```
FuriousLlama, line 150: instead of iterating over the entries here, we can select the first entry in the month calendar view. After that, we can use the same logic to find the date on the list and move the page there. The change would be to split that function between the focus and just the page change

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: how to do you break the "for"? I have tried foreach but breaking it is not recommended.

Apps/WebClient/src/ClientApp/src/components/calendar/header.vue
leobcgov: auto-merge issue, I will need to revert the code back

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: do you think the "currentPage < this.numberOfPages" condition can back this up even if there are less than 3 items right?

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: I see what you mean, only thing that I can see is the order of entries i.e. calendar shows most recent entry last, but timeline shows them first. We can discuss this and other comments above on a call tomorrow to make sure.

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
FuriousLlama: I would write it like:
```
for (let i = 0; i< i < timelineEntriesLenght && !foundLastEntryOfSelectedMonth; i++)
```

leobcgov: sounds good, I will make the changes

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
leobcgov: I remove this part completely then

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -254,5 +267,59 @@ export default class LinearTimelineComponent extends Vue {
             container[0].focus();
         });
     }
+
+    private goToPageByCurrentDate(currentDate: Date) {
```
FuriousLlama, line 271: The logic for this already exists on the method ```setPageFromDate```. How about spliting that after the nextTick section and re-using it?

Apps/WebClient/src/ClientApp/src/components/timeline/linearTimeline.vue
```diff
@@ -117,10 +119,14 @@ export default class LinearTimelineComponent extends Vue {
     @Watch("visibleTimelineEntries")
     private onVisibleEntriesUpdate() {
         if (this.visibleTimelineEntries.length > 0) {
-            this.eventBus.$emit(
-                "timelinePageUpdate",
-                new Date(this.visibleTimelineEntries[0].date)
-            );
+            if (this.emitTimelinePageUpdateEnabled) {
```
FuriousLlama, line 122: Instead, we can set it as if is visible.

leobcgov:
updates re. @FuriousLlama 's comments and my call with him today.

Apps/WebClient/src/ClientApp/src/views/timeline.vue
```diff
@@ -209,13 +209,15 @@
...
                     :timeline-entries="timelineEntries"
+                    :is-list-view="isListView"
                     :total-entries="getTotalCount()"
                     :filter-text="filterText"
                     :filter-types="filterTypes"
                 />
                 <CalendarTimeline
                     v-show="!isListView && !isLoading"
                     :timeline-entries="timelineEntries"
+                    :is-list-view="isListView"
```
FuriousLlama, line 220: instead of the property  named is-list-view it can be just is-visible

Apps/WebClient/src/ClientApp/src/components/calendar/body.vue
```diff
@@ -201,17 +204,33 @@ export default class CalendarBodyComponent extends Vue {
...
                     date: new Date(startDate),
                     events: this.slotEvents(startDate),
-                });
+                };
+                week.days.push(dayEvent);
 
                 startDate.setDate(startDate.getDate() + 1);
+
+                if (
+                    this.firstEventDateOfCurrentMonth === null &&
```
FuriousLlama, line 221: I dont think this is necessary. I would use a lambda instead to find the first event day that is on this month, then the first day from that

Apps/WebClient/src/ClientApp/src/views/timeline.vue
leobcgov: thanks @FuriousLlama , please can you review my last checkin?

Apps/WebClient/src/ClientApp/src/components/calendar/body.vue
leobcgov: I'm not sure how it can be better, but I will try. Btw, it is not an easy change as this logic needs to be tested carefully as well

leobcgov: Per our discussion, created 8741 for this refactoring.

PR #676, opened 2020-07-28
Add new recurring jobs for standard, high and urgent priority emails

sam-warren:
# Fixes or Implements [AB#8679](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8679)

* [X] Enhancement
* [ ] Bug
* [ ] Documentation

## Reviewer Guidelines

The reviewer should then:

1. Confirm the merge branch

1. Review the Code Changes

    1. useful tips

## Description

Added new jobs to periodically send standard, high and urgent priority emails.

## Testing

* [ ] Unit Tests Updated
* [X] Not Required

### UI Changes

NO

### Browsers Tested

* [X] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

### DIFF
--- a/Apps/Database/src/Delegates/DBEmailDelegate.cs
+++ b/Apps/Database/src/Delegates/DBEmailDelegate.cs
@@ -78,6 +78,45 @@ namespace HealthGateway.Database.Delegates
         }
 
         /// <inheritdoc />
+        public List<Email> GetStandardPriorityEmail(int maxRows)
+        {
+            this.logger.LogTrace($"Getting list of standard priority emails from DB... {maxRows}");
+            List<Email> retVal = this.dbContext.Email.Where(p => p.EmailStatusCode == EmailStatus.New &&
+                                                   p.Priority >= EmailPriority.Standard && p.Priority < EmailPriority.High)
+                                        .OrderByDescending(s => s.Priority)
+                                        .Take(maxRows)
+                                        .ToList();
+            this.logger.LogDebug($"Finished getting list of standard priority emails from DB. {JsonSerializer.Serialize(retVal)}");
+            return retVal;
+        }
+
+        /// <inheritdoc />
+        public List<Email> GetHighPriorityEmail(int maxRows)
+        {
+            this.logger.LogTrace($"Getting list of high priority emails from DB... {maxRows}");
+            List<Email> retVal = this.dbContext.Email.Where(p => p.EmailStatusCode == EmailStatus.New &&
+                                                   p.Priority >= EmailPriority.High && p.Priority < EmailPriority.Urgent)
+                                        .OrderByDescending(s => s.Priority)
+                                        .Take(maxRows)
+                                        .ToList();
+            this.logger.LogDebug($"Finished getting list of high priority emails from DB. {JsonSerializer.Serialize(retVal)}");
+            return retVal;
+        }
+
+        /// <inheritdoc />
+        public List<Email> GetUrgentPriorityEmail(int maxRows)
+        {
+            this.logger.LogTrace($"Getting list of urgent priority emails from DB... {maxRows}");
+            List<Email> retVal = this.dbContext.Email.Where(p => p.EmailStatusCode == EmailStatus.New &&
+                                                   p.Priority >= EmailPriority.Urgent)
+                                        .OrderByDescending(s => s.Priority)
+                                        .Take(maxRows)
+                                        .ToList();
+            this.logger.LogDebug($"Finished getting list of urgent priority emails from DB. {JsonSerializer.Serialize(retVal)}");
+            return retVal;
+        }
+
+        /// <inheritdoc />
         public Guid InsertEmail(Email email, bool shouldCommit = true)
         {
             this.logger.LogTrace($"Inserting email to DB... {email}");
--- a/Apps/Database/src/Delegates/IEmailDelegate.cs
+++ b/Apps/Database/src/Delegates/IEmailDelegate.cs
@@ -40,7 +40,7 @@ namespace HealthGateway.Database.Delegates
         Email GetNewEmail(Guid emailId);
 
         /// <summary>
-        /// Gets a sorted list of New low priority emails to process.
+        /// Gets a sorted list of new low priority emails to process.
         /// Low priority for this job is anything less than EmailPriority.Standard
         /// This is to ensure that any manual manipulation in the database that
         /// changes the ordinal values above those defined as EmailPriority.Low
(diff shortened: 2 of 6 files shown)


Apps/Database/src/Delegates/DBEmailDelegate.cs
```diff
@@ -77,6 +77,45 @@ public List<Email> GetLowPriorityEmail(int maxRows)
             return retVal;
         }
 
+        /// <inheritdoc />
+        public List<Email> GetStandardPriorityEmail(int maxRows)
+        {
+            this.logger.LogTrace($"Getting list of standard priority emails from DB... {maxRows}");
+            List<Email> retVal = this.dbContext.Email.Where(p => p.EmailStatusCode == EmailStatus.New &&
+                                                   p.Priority >= EmailPriority.Standard && p.Priority < EmailPriority.High)
```
leobcgov, line 85: as we are sending Standard ones why do we need "..p.Priority < EmailPriority.High"?
Should it just be "p.Priority == EmailPriority.Standard" ?

Apps/Database/src/Delegates/DBEmailDelegate.cs
```diff
@@ -77,6 +77,45 @@ public List<Email> GetLowPriorityEmail(int maxRows)
...
+            this.logger.LogDebug($"Finished getting list of standard priority emails from DB. {JsonSerializer.Serialize(retVal)}");
+            return retVal;
+        }
+
+        /// <inheritdoc />
+        public List<Email> GetHighPriorityEmail(int maxRows)
+        {
+            this.logger.LogTrace($"Getting list of high priority emails from DB... {maxRows}");
+            List<Email> retVal = this.dbContext.Email.Where(p => p.EmailStatusCode == EmailStatus.New &&
+                                                   p.Priority >= EmailPriority.High && p.Priority < EmailPriority.Urgent)
```
leobcgov, line 98: for High priority, should we use "p.Priority == EmailPriority.High" ?

Apps/Database/src/Delegates/DBEmailDelegate.cs
```diff
@@ -77,6 +77,45 @@ public List<Email> GetLowPriorityEmail(int maxRows)
...
+            this.logger.LogDebug($"Finished getting list of high priority emails from DB. {JsonSerializer.Serialize(retVal)}");
+            return retVal;
+        }
+
+        /// <inheritdoc />
+        public List<Email> GetUrgentPriorityEmail(int maxRows)
+        {
+            this.logger.LogTrace($"Getting list of urgent priority emails from DB... {maxRows}");
+            List<Email> retVal = this.dbContext.Email.Where(p => p.EmailStatusCode == EmailStatus.New &&
+                                                   p.Priority >= EmailPriority.Urgent)
```
leobcgov, line 111: Should we get only Urgent emails here?

Apps/Database/src/Delegates/IEmailDelegate.cs
```diff
@@ -51,6 +51,42 @@ public interface IEmailDelegate
...
+        /// but below the ordinal value of EmailPriority.Standard will still be processed
+        /// by this job.
+        /// The list will be ordered by the Priority in descending order.</summary>
+        /// <param name="maxRows">The maximum amount of emails to return.</param>
+        /// <returns>The list of emails.</returns>
+        List<Email> GetStandardPriorityEmail(int maxRows);
+
+        /// <summary>
+        /// Gets a sorted list of new high priority emails to process.
+        /// Standard priority for this job is anything less than EmailPriority.Urgent and more than or equal to EmailPriority.High
```
leobcgov, line 68: need to update this comment?

Apps/Database/src/Delegates/IEmailDelegate.cs
```diff
@@ -51,6 +51,42 @@ public interface IEmailDelegate
...
+        /// but below the ordinal value of EmailPriority.High will still be processed
+        /// by this job.
+        /// The list will be ordered by the Priority in descending order.</summary>
+        /// <param name="maxRows">The maximum amount of emails to return.</param>
+        /// <returns>The list of emails.</returns>
+        List<Email> GetHighPriorityEmail(int maxRows);
+
+        /// <summary>
+        /// Gets a sorted list of new urgent priority emails to process.
+        /// Standard priority for this job is anything more than or equal to EmailPriority.Urgent
```
leobcgov, line 80: this comment is not relevant?

Apps/JobScheduler/src/Jobs/EmailJob.cs
```diff
@@ -107,6 +107,90 @@ public void SendLowPriorityEmails()
...
+                    {
+                        this.SendEmail(email);
+                    }
+                    catch (Exception e)
+                    {
+                        // log the exception as a warning but we can continue
+                        this.logger.LogWarning($"Error while sending {email.Id} - skipping for now\n{e.ToString()}");
+                    }
+#pragma warning restore CA1031 // Restore warnings.
+                }
```
leobcgov, line 160: Can this block of code from "if (resendEmails.Count > 0)" to here can be a private method which can be used in 4 different methods for low/standard/high/urgent emails?

Apps/Database/src/Delegates/DBEmailDelegate.cs
tiagograf-AOT: I think the confusion here is that we use priorities in the database from 0 to 10 (if i remember correctly) but we only have 3 values in our EmailPriority Enum (Standard, High, Urgent) so there are numbers in between those three (lets say standard is 4 and high is 7, if i have an email priority that is 5 its considered as standard because its less than the high number. I don't think there's a problem but i'm sure we could have done this in a simpler way

Apps/Database/src/Delegates/IEmailDelegate.cs
tiagograf-AOT: +1

Apps/JobScheduler/src/Jobs/EmailJob.cs
tiagograf-AOT: I agree with this

Apps/Database/src/Delegates/DBEmailDelegate.cs
leobcgov: right admin client can only set 1 value for standard. The purpose of this email job is just to send that Standard email, nothing else right?

Apps/JobScheduler/src/Jobs/EmailJob.cs
sam-warren: Great idea!

Apps/Database/src/Delegates/DBEmailDelegate.cs
sslaws: It should cover all cases but I think that we could probably create a bug to update the DB to handle only the enumeration values.

sam-warren: Hey - we do it this way so that a change in the numbers in the DB (i.e. if someone manually changed the priority to 150 it would still get sent) doesn't interfere with the functionality of the job

Apps/Database/src/Delegates/IEmailDelegate.cs
sam-warren: good catch

Apps/Database/src/Delegates/DBEmailDelegate.cs
leobcgov: It doesn't seem a good way to use enum here though. Enum is to limit the choices. Changing priority to 150 should not be expected unless we introduce a new priority here, i.e. "Super Urgent" = 150.
We can chat further on our call to make sure.

PR #793, opened 2020-09-01
Encounter service and model updates

FuriousLlama:
# Fixes or Implements [AB#8704](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/8704)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Reviewer Guidelines

The reviewer should then:

1. Confirm the merge branch

1. Review the Code Changes

    1. useful tips

## Description

- Added Encounter service
- Refactored medication statements to use the ODR base class
- Updated frontend model. 
- Added clinic to the backend model of encounter

## Testing

* [x] Unit Tests Updated
* [ ] Not Required


Apps/Encounter/src/Models/EncounterHistory.cs
```diff
@@ -0,0 +1,101 @@
...
+namespace HealthGateway.Encounter.Models
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// Represents a patient Encounter.
+    /// </summary>
+    public class EncounterHistory
```
leobcgov, line 25: It seems to me that we have issue with the naming for the models. Personally, I prefer to call this EncounterDTO  
or ClaimDTO instead so we don't have to think or come up with a new suffix for the new domain models.

FuriousLlama: what does DTO stand for?

leobcgov: it stands for Data Transfer Object which is generally used to expose the object to api client / client UI <ref. https://medium.com/@Colin_But/dissecting-the-dto-pattern-ac3e54d0e4c8>

FuriousLlama: Not 100% sure is better, but I rather see something like EncounterData / EncounterModel.

leobcgov: These sounds good to me to. Btw, just trying to prove this DTO suffix naming to see if you are ok.
https://github.com/PacktPublishing/ASP.NET-Core-3-and-Angular-9-Third-Edition/tree/ca0aca9df29be086bfe736c2c83cc15ed7f31c29/Chapter_07/WorldCities/Data is the source code from a C# MVP guys who wrote the good "ASP.NET Core 3 and Angular 9" book.
Check it out and see if we can decide to use ...DTO (...Dto) vs ...Data / ...Model

FuriousLlama: Lets go with Model then. In general, I prefer not to use acronyms on variable or/and class names since it can bloat the meaning really fast

leobcgov: Sounds good, I update my PR as well

PR #835, opened 2020-09-14
Enhanced filters Features/9076

FuriousLlama:
# Fixes or Implements [AB#9076](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9076)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Filters update

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### UI Changes

YES

### Browsers Tested

* [x] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

image
image
image
image


Apps/WebClient/src/ClientApp/src/components/timeline/filters.vue
```diff
@@ -0,0 +1,277 @@
...
+
+    private toggleMobileView() {
+        this.isVisible = !this.isVisible;
+    }
+
+    private clearFilters(): void {
+        this.selectedFilters = [];
+    }
+
+    private onExternalFilterSelection(filterName: string) {
```
leobcgov, line 258: as this method updates the selectedFilters, can we call this updateSelectedFilter?

leobcgov (changes requested):
nice one, just one naming item to review

Apps/WebClient/src/ClientApp/src/components/timeline/filters.vue
FuriousLlama: the catch is that this is only called when from an external call via the bus. IMHO updateSelectedFilter gives the impression that this updates it in a general way

leobcgov: Mbe an item for retro to define our naming convention as for me method should have a meaningful name and its name should tell what it does. Approve this for now then :-)

leobcgov (approved):

tiagograf-AOT (approved):

PR #844, opened 2020-09-16
Implemented unsupported browser

leobcgov:
# Fixes or Implements [AB#9106](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9106) & 9107

* [X] Enhancement
* [ ] Bug
* [ ] Documentation

## Description
- Implemented browser detection and unsupported browser message using modernizr.js
- Displaying the following message on my IE11
image

## Testing

* [ ] Unit Tests Updated
* [X] Not Required

### Steps to Reproduce

### UI Changes

YES

### Browsers Tested

* [X] Chrome
* [X] IE 11
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments

</br>
</br>
</br>
___
  
## Reviewer Guidelines

The reviewer should then:

1. Confirm the merge branch

1. Review the Code Changes

    1. useful tips


sslaws (changes requested):
I think that we should not be altering the Snowplow script as we will have to think about this every time we update and we will be removing any analytics gathered from non-compliant browsers.  We should simply call both functions onload by chaining.

leobcgov:
> I think that we should not be altering the Snowplow script as we will have to think about this every time we update and we will be removing any analytics gathered from non-compliant browsers. We should simply call both functions onload by chaining.
Good point!

Apps/WebClient/src/ClientApp/public/index.html
```diff
@@ -18,49 +18,57 @@
...
-                    p[i] = function () {
-                        (p[i].q = p[i].q || []).push(arguments);
-                    };
-                    p[i].q = p[i].q || [];
-                    n = l.createElement(o);
-                    g = l.getElementsByTagName(o)[0];
-                    n.async = 1;
-                    n.src = w;
-                    g.parentNode.insertBefore(n, g);
+            function detectBrowser() {
```
FuriousLlama, line 25: This conditionally switches the snowplow script, no need for that. Lets move the detectBrowser to its own script.

Apps/WebClient/src/ClientApp/public/index.html
```diff
@@ -18,49 +18,57 @@
...
+                if (!Modernizr.es6number) {
+                    const unsupportedBrowserMsg = "Health Gateway is not compatible with this browser. Please use another browser.";
+                    document.getElementById("app-root").innerHTML = unsupportedBrowserMsg;
```
FuriousLlama, line 28: instead of adding the test to app-root I say lets remove app-root completly so that we are sure the binding to vue does not happen

FuriousLlama (changes requested):

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: aha I like this one!

leobcgov: aiee, need to find a way to remove a div in JS as the remove() is not supported in IE (worst browser ever)

leobcgov: Found that removeChild works for IE yeah!

Apps/WebClient/src/ClientApp/public/index.html
```diff
@@ -59,8 +60,24 @@
...
+        <script type="text/javascript">
+            function detectBrowser() {
+                const body = document.getElementsByTagName("BODY")[0];
+                const unsupportedBrowser = document.getElementById("unsupported-browser");
+                if (Modernizr.es6number) {          
+                    body.removeChild(unsupportedBrowser);
+                }
+                else {          
+                    body.removeChild(document.getElementById("app-root"));
+                    const unsupportedBrowserMsg = "Health Gateway is not compatible with this browser. Please use another browser.";
```
FuriousLlama, line 73: if the contents of the div are going to be removed, why not set them there instead of programmatically changing it?

Apps/WebClient/src/ClientApp/public/index.html
```diff
@@ -18,7 +18,8 @@
             name="google-site-verification"
             content="doAsS2f0kVQcQMjZcJuo2Pdj60GkhXI2iKbwfaUBImk"
         />
-
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js" integrity="sha512-3n19xznO0ubPpSwYCRRBgHh63DrV+bdZfHK52b1esvId4GsfwStQNPJFjeQos2h3JwCmZl0/LgLxSKMAI55hgw==" crossorigin="anonymous">
```
FuriousLlama, line 21: instead of getting the script out of the internet on demand we can get it as a package and reference the node folder

leobcgov: sure, I can try to bring it down to our node folder.

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: sounds like a better approach, I can chance it

leobcgov: doesn't seem to work as VueJS is automatically looking for its app-root div. I leave it as is for now if you don't mind
image

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: @FuriousLlama I have generated a very small modernizr.js by adding just the es6 number feature and placed it under public folder, do you think it is the right place for it to be in?

tiagograf-AOT (approved):

Apps/WebClient/src/ClientApp/public/index.html
FuriousLlama: I think it will give that errors based on ```body.removeChild(document.getElementById("app-root"));```  What I mean is that "Health Gateway is not compatible with this browser. Please use another browser." should be  in the ```  <div id="unsupported-browser"></div>``` already

Apps/WebClient/src/ClientApp/public/index.html
FuriousLlama: Looking bvetter! Did you check it works if it deploys? i.e. npm run build

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: oh I see, good idea. I 'm fixing it now

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: yes, npm run build runs perfectly. It generates the JS and copies it to the dist folder.

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: btw, in IE, I'm seeing 2 div (loading... and unsupported one) at the same time before the loading... is removed. I will leave them as they are then?

Apps/WebClient/src/ClientApp/public/index.html
FuriousLlama: right, and the index file has access to that path?

leobcgov: > right, and the index file has access to that path?
yes, the modernizr-custom.js is downloaded in both chrome and ie per my tests in local. Do you know any ways that we can do to make sure if it is accessible from dist ?

Apps/WebClient/src/ClientApp/public/index.html
FuriousLlama: both should be there, since we dont know when does the vue loading happen vs the script execution. you can also add a display none to both and remove it on the script

FuriousLlama (approved):

Apps/WebClient/src/ClientApp/public/index.js
```diff
@@ -0,0 +1,16 @@
...
+        body.removeChild(unsupportedBrowser);
+    } else {
+        body.removeChild(document.getElementById("app-root"));
+        const unsupportedBrowserMsg =
+            "Health Gateway is not compatible with this browser. Please use another browser.";
+        unsupportedBrowser.innerHTML = unsupportedBrowserMsg;
+    }
+}
+window.onload = function () {
+    detectBrowser();
```
leobcgov, line 15: moving this code from <body..> of the index.html to this js code has helped to prevent the "Refused to execute inline event handler...violates the following Content Security Policy..." yeah!
image

Apps/WebClient/src/ClientApp/public/index.html
```diff
@@ -62,5 +61,8 @@
     </head>
     <body class="d-flex flex-column">
         <div id="app-root">Loading...</div>
+        <div id="unsupported-browser"></div>
+        <script src="/modernizr-custom.js"></script>
+        <script src="/index.js"></script>
```
FuriousLlama, line 66: intead of index.js lets name it detectBrowser or something like that.

sslaws (approved):

Apps/WebClient/src/ClientApp/public/index.html
```diff
@@ -18,49 +18,57 @@
             name="google-site-verification"
             content="doAsS2f0kVQcQMjZcJuo2Pdj60GkhXI2iKbwfaUBImk"
         />
-
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js" integrity="sha512-3n19xznO0ubPpSwYCRRBgHh63DrV+bdZfHK52b1esvId4GsfwStQNPJFjeQos2h3JwCmZl0/LgLxSKMAI55hgw==" crossorigin="anonymous">
+        </script>
         <!-- Snowplow starts plowing - Standalone vE.2.14.0 -->
         <script type="text/javascript">
-            (function (p, l, o, w, i, n, g) {
```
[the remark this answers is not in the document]
leobcgov: good idea, I will do this one then

Apps/WebClient/src/ClientApp/public/index.html
FuriousLlama: also, does the modernizer script need to be in the body?

Apps/WebClient/src/ClientApp/public/index.html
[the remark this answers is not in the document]
leobcgov: @sslaws per your comment on the adding a cdn to our index, I don't have any concerns with it. @FuriousLlama may have some though. Btw, the custom JS that I generated is way smaller than its cdn and we have option to choose only the ES6 Number feature of the modernizr.js and not the whole package.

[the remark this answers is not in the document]
leobcgov: > Since we're still working on this, an optional request would be to take the Snowplow code into its own js file and embed a reference as we're doing with the modernizr

wow, now I can make the ContentSecurityPolicy > script-src is super light, it can contain only "script-src": "https://www2.gov.bc.ca/StaticWebResources/static/sp/sp-2-14-0.js" without complaining violation due to inline execution.

Apps/WebClient/src/ClientApp/public/index.html
leobcgov: I believe the script tags are recommended to be at the end of the html file. Per the following screenshot, VueJS generates their scripts and inserts them into the same location:
image

leobcgov: and per https://www.w3schools.com/js/js_whereto.asp , "Placing scripts at the bottom of the <body> element improves the display speed, because script interpretation slows down the display."

Apps/WebClient/src/ClientApp/public/index.html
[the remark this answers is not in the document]
FuriousLlama: If you move it to its own file make sure that the snowplow script actually adds the scripts it wants in the position it needs

Apps/WebClient/src/ClientApp/public/snowplow.js
```diff
@@ -0,0 +1,39 @@
+(function (p, l, o, w, i, n, g) {
```
FuriousLlama, line 1: I think we still need the comments before and after

Apps/WebClient/src/ClientApp/public/index.html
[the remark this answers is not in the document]
FuriousLlama: also, there might be a difference for snowplow between runing on the header vs running from the body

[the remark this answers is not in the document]
leobcgov: Even it seems to work find to place them in the body, I have moved scripts back to the header to avoid any side effects for snowplow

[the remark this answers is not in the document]
leobcgov: @FuriousLlama a part of logging at the starting and ending of this blog, is there any ways that we can confirm if snowplow is working correctly?

sslaws (approved):

leobcgov:
@FuriousLlama the <script async="" src="https://www2.gov.bc.ca/StaticWebResources/static/sp/sp-2-14-0.js"> is present per my tests in local for this PR. The local header looks good too.

FuriousLlama (approved):

PR #873, opened 2020-09-25
Aded identityProvider to user Store

FuriousLlama:
# Fixes or Implements [AB#9189](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9189)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Added idp section to authentication and user state.

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### UI Changes

NO

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/store/modules/auth/getters.ts
+++ b/Apps/WebClient/src/ClientApp/src/store/modules/auth/getters.ts
@@ -60,4 +60,9 @@ export const getters: GetterTree<AuthState, RootState> = {
     oidcError(state: AuthState): any {
         return state.error;
     },
+    isBcscLogin(state: AuthState): boolean {
+        return state.authentication === undefined
+            ? false
+            : state.authentication.identityProvider === "BCSC";
+    },
 };
--- a/Apps/WebClient/src/ClientApp/src/store/modules/auth/getters.ts
+++ b/Apps/WebClient/src/ClientApp/src/store/modules/auth/getters.ts
@@ -60,9 +60,9 @@ export const getters: GetterTree<AuthState, RootState> = {
     oidcError(state: AuthState): any {
         return state.error;
     },
-    isBcscLogin(state: AuthState): boolean {
+    isValidIdentityProvider(state: AuthState): boolean {
         return state.authentication === undefined
             ? false
-            : state.authentication.identityProvider === "BCSC";
+            : state.authentication.identityProvider !== "IDIR";
     },
 };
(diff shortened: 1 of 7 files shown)


Apps/WebClient/src/ClientApp/src/store/modules/auth/mutations.ts
```diff
@@ -27,6 +28,7 @@ export const mutations: MutationTree<AuthState> = {
         Vue.set(state.authentication, "accessToken", undefined);
         Vue.set(state.authentication, "scopes", undefined);
         Vue.set(state.authentication, "idToken", undefined);
+        Vue.set(state.authentication, "identityProvider", "");
```
leobcgov, line 31: may be use undefined instead of "", if "" is better to use here, we can get a task for refactoring better to be consistent. Also this should be aligned with the check in the isBcscLogin method in the getters.ts above.

leobcgov (changes requested):

Apps/WebClient/src/ClientApp/src/store/modules/auth/mutations.ts
FuriousLlama: the type of the other variables is a string or undefined, idp is a string only.
To your second point, the authentication object could be undefined, but that is not the variable of said object
```
return state.authentication === undefined
```

leobcgov (approved):

Apps/WebClient/src/ClientApp/src/store/modules/auth/mutations.ts
leobcgov: oh I see, my bad

Apps/WebClient/src/ClientApp/src/store/modules/auth/getters.ts
```diff
@@ -60,4 +60,9 @@ export const getters: GetterTree<AuthState, RootState> = {
     oidcError(state: AuthState): any {
         return state.error;
     },
+    isBcscLogin(state: AuthState): boolean {
```
sslaws, line 63: What is the purpose of this check?  I think we want to specifically determine if IDIR not BCSC as we don't want to prevent Keycloak logins and we don't want to prevent future IDPs.

sslaws (changes requested):

Apps/WebClient/src/ClientApp/src/store/modules/auth/getters.ts
FuriousLlama: At this point the only valid IDP is bcsc, so that is what this is checking. If we get another provider we could change it in the future.

FuriousLlama: updated to check for IDIR

leobcgov (approved):

sslaws (approved):

PR #978, opened 2020-10-08
Expand sidebar on login

sam-warren:
# Fixes or Implements [AB#9343](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9343)

* [X] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Sidebar expands on login.

## Testing

* [ ] Unit Tests Updated
* [X] Functional Tests Updated
* [ ] Not Required

### UI Changes

NO

### Browsers Tested

* [X] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Reviewer Guidelines

The reviewer should then:

1. Confirm the merge branch

1. Review the Code Changes

    1. useful tips


Apps/WebClient/src/ClientApp/src/views/loginCallback.vue
```diff
@@ -37,6 +43,12 @@ export default class LoginCallbackView extends Vue {
                     this.checkRegistration({ hdid: this.user.hdid }).then(
                         () => {
                             this.$router.push({ path: redirectPath });
+                            if (
+                                !this.isSidebarOpen &&
```
FuriousLlama, line 47: Instead how about having the sidebar expanded by default?

sam-warren: If that were the case, every time a user refreshed the sidebar would expand. If this is expected behaviour then I can change it but this implementation follows the story's requirements exactly

sam-warren: We would also need to implement logic to only expand on mobile anyways

FuriousLlama: Well it would expand it on refresh too. Making the default extended.

FuriousLlama: I was thinking about it and it would be the same for mobile. Something like, if is mobile set the default state to collapse, otherwise is extended. I don't think having the animation being shown every time would be necessary

tiagograf-AOT (approved):

Apps/WebClient/src/ClientApp/src/views/loginCallback.vue
FuriousLlama: Chatted win nino and we agree that it should be expanded by default

FuriousLlama (changes requested):
Should be expanded by default on desktop

Apps/WebClient/src/ClientApp/src/store/modules/sidebar/actions.ts
```diff
@@ -5,4 +5,8 @@ export const actions: ActionTree<SidebarState, RootState> = {
     toggleSidebar(context) {
         context.commit("toggle");
     },
+
+    setSidebarState(context, isOpen: boolean) {
```
FuriousLlama, line 9: since the parameter is isOpen, how about setIsOpenState?

FuriousLlama (approved):

leobcgov (approved):

PR #1232, opened 2020-11-04
Fixes Remove Dependant

FuriousLlama:
# Fixes or Implements [AB#9494](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9494)

* [ ] Enhancement
* [x] Bug
* [ ] Documentation

## Description

Added missing logic

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### Steps to Reproduce

Open the dependant's page, delete a dependant

### UI Changes

NO


leobcgov (dismissed):
We though it was missing the User Delegate injection

Apps/Database/src/Delegates/DBUserDelegateDelegate.cs
```diff
@@ -91,14 +91,9 @@ public DBResult<IEnumerable<UserDelegate>> Get(string delegateId, int page, int
         public DBResult<UserDelegate> Delete(string ownerId, string delegateId, bool commit)
         {
             this.logger.LogTrace($"Deleting UserDelegate (ownerId: {ownerId}, delegateId: {delegateId}) from DB...");
-            UserDelegate userDelegate = new UserDelegate()
```
sslaws, line 94: We should give this back to whomever did delete.  We should not be doing a query to do a delete, instead we should be passing in the object that was originally fetched or marginally the key identifiers along with the version number.

sslaws (changes requested):

Apps/Database/src/Delegates/DBUserDelegateDelegate.cs
FuriousLlama: at this point the whole fetch does not have a version in it, so....

sslaws: UserDelegate is defined as public class UserDelegate : AuditableEntity anything that is Auditable has a version attached to it.  The object returned by GetUserDelegate and InsertUserDelegate is returning the full object.  We should be using that...

FuriousLlama: I've updated the model to contain a representation of the user delegate to correctly use the concurrency checks

Apps/WebClient/src/Server/Models/DependentModel.cs
```diff
@@ -26,29 +26,51 @@ public class DependentModel
...
-        /// Gets or sets the Masked PHN of the dependent.
+        /// Gets or sets the hdid which has delegated access to the owner Id.
+        public string DelegateId { get; set; } = null!;
+
+        /// <summary>
+        /// Gets or sets the version.
         /// </summary>
-        [JsonPropertyName("maskedPHN")]
-        public string MaskedPHN { get; set; } = string.Empty;
+        public uint Version { get; set; }
```
sslaws, line 43: No documentation on this property, is it not warning you on web client?

sslaws (changes requested):
Please ensure no smells introduced across projects.

Apps/WebClient/src/Server/Models/DependentModel.cs
FuriousLlama: Not sure what you mean. These are the docs I see 
```
        /// <summary>
        /// Gets or sets the version.
        /// </summary>
```

tiagograf-AOT (approved):

sslaws (approved):

PR #1274, opened 2020-11-11
Dependent with Testdate and PHN masking Features/9509 & Features/9523

FuriousLlama:
# Fixes or Implements [AB#9509](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9509)  & [AB#9523](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9523)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Added Test date and DB functionality to store the delegate statement.
PHN is not masked anymore

## Testing

* [x] Unit Tests Updated
* [ ] Not Required


sslaws (changes requested):
I don't get this PR at all we now have a DBUserDelegateDelegate and a DBUserDelegateStatementDelegate and associated models which appear to be the same thing but one has a Test date which doesn't belong with the dataset (delegation).  There is no history table on the new delegate table and authorization is not updated to use it so I'm not sure which one is used for what purpose.

Apps/Database/src/Models/UserDelegateStatement.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Gets or sets the hdid which has delegated access to the owner Id.
+        /// </summary>
+        [MaxLength(52)]
+        public string DelegateId { get; set; } = null!;
+
+        /// <summary>
+        /// Gets or sets the Test date of the user delegate statement.
+        /// </summary>
+        [MaxLength(52)]
```
leobcgov, line 41: no need MaxLength for DateTime right?

FuriousLlama:
> I don't get this PR at all we now have a DBUserDelegateDelegate and a DBUserDelegateStatementDelegate and associated models which appear to be the same thing but one has a Test date which doesn't belong with the dataset (delegation). There is no history table on the new delegate table and authorization is not updated to use it so I'm not sure which one is used for what purpose.

Putting what we chatted on rc here:
----
Yep, but if we are going to associate the test date we have to have the key from the other relationship. And since the key is all the members.... 🙂

laws
5:26 PM
I don't disagree with that... sounds like w should probably add a surrogate key or we should reluctantly add it to the existing table which is probably preferrable over the naming and duplicate objects 🙂 Just too late for a design change like this

manuel.rodriguez
6:42 PM
We chatted about adding it to the existing table and agreed that was not a good idea.
I can change the name of it, but otherwise we will have to stick to the double key like it's parent table. I feel that changing the other table pk would be a larger ordeal, and more for a field we are not sure how is going to used.
My vote is to leave it as a separate (and kinda ugly table) for now until we get a better idea on how that field is going to hook with the test of the system

Apps/Database/src/Context/GatewayDbContext.cs
```diff
@@ -73,8 +73,9 @@ public GatewayDbContext(DbContextOptions<GatewayDbContext> options)
         public DbSet<Communication> Communication { get; set; } = null!;
         public DbSet<CommunicationEmail> CommunicationEmail { get; set; } = null!;
         public DbSet<Rating> Rating { get; set; } = null!;
-        public DbSet<UserDelegate> UserDelegate { get; set; } = null!;
-        public DbSet<UserDelegateHistory> UserDelegateHistory { get; set; } = null!;
+        public DbSet<ResourceDelegate> ResourceDelegate { get; set; } = null!;
+        public DbSet<ResourceDelegateHistory> ResourceDelegateHistory { get; set; } = null!;
+        public DbSet<ResourceDelegateReasonCode> ResourceDelegateReasonCode { get; set; } = null!;
```
sslaws, line 78: NIT:  You don't need to explicitly call out code tables unless you want to be able to perform operations on them.

Apps/Database/src/Migrations/20201118002050_AddResourceDelegate.cs
```diff
@@ -0,0 +1,216 @@
...
+    {
+        private const string Schema = "gateway";
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            // Drop the UserDelegate
+            migrationBuilder.Sql(@$"DROP TABLE IF EXISTS {Schema}.""UserDelegate""");
+
+            // Drop the UserDelegateHistory
+            migrationBuilder.Sql(@$"DROP TABLE IF EXISTS {Schema}.""UserDelegateHistory""");
+
```
sslaws, line 34: What about the trigger and the function?

sslaws (commented):
Question about the old trigger and function, looks good overall.

Apps/Database/src/Context/GatewayDbContext.cs
FuriousLlama: good catch, I've removed them.

Apps/Database/src/Migrations/20201118002050_AddResourceDelegate.cs
FuriousLlama: I am guessing it gets removed alongside the table.

FuriousLlama: Doubled checked and they are left if not taken manually.  Latest commit removes them

sslaws (dismissed):

sam-warren (approved):

leobcgov (approved):

tiagograf-AOT (dismissed):

leobcgov (dismissed):

tiagograf-AOT (approved):

sslaws (approved):

PR #1341, opened 2020-11-24
Fixes new typescript syntax errors AB#9609

tiagograf-AOT:
# Fixes or Implements [AB#9609](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9609)

* [ ] Enhancement
* [x] Bug
* [ ] Documentation

## Description

Fixes webclient build.
Refactor handle request result.

## Testing

* [x] Unit Tests Updated
* [ ] Not Required

### Steps to Reproduce

### UI Changes

NO

### Browsers Tested

* [ ] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments

</br>
</br>
</br>
___
  
## Reviewer Guidelines

The reviewer should then:

1. Confirm the merge branch

1. Review the Code Changes

    1. useful tips

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/services/restDependentService.ts
+++ b/Apps/WebClient/src/ClientApp/src/services/restDependentService.ts
@@ -32,37 +32,41 @@ export class RestDependentService implements IDependentService {
     public addDependent(
         hdid: string,
         dependent: AddDependentRequest
-    ): Promise<AddDependentRequest> {
-        return new Promise<AddDependentRequest>((resolve, reject) => {
-            if (!this.isEnabled) {
-                reject("dependent module is disabled.");
-                return;
+    ): Promise<AddDependentRequest | undefined> {
+        return new Promise<AddDependentRequest | undefined>(
+            (resolve, reject) => {
+                if (!this.isEnabled) {
+                    resolve(undefined);
+                    return;
+                }
+                this.http
+                    .post<RequestResult<AddDependentRequest>>(
+                        `${this.BASE_URI}/UserProfile/${hdid}/Dependent`,
+                        dependent
+                    )
+                    .then((requestResult) => {
+                        this.logger.verbose(
+                            `addDependent result: ${JSON.stringify(
+                                requestResult
+                            )}`
+                        );
+                        return RequestResultUtil.handleResult(
+                            requestResult,
+                            resolve,
+                            reject
+                        );
+                    })
+                    .catch((err) => {
+                        this.logger.error(err);
+                        return reject(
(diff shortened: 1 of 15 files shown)


Apps/WebClient/src/ClientApp/src/services/restDependentService.ts
```diff
@@ -33,21 +33,25 @@ export class RestDependentService implements IDependentService {
         hdid: string,
         dependent: AddDependentRequest
     ): Promise<AddDependentRequest> {
-        return new Promise((resolve, reject) => {
+        return new Promise<AddDependentRequest>((resolve, reject) => {
             if (!this.isEnabled) {
-                resolve();
+                reject("dependent module is disabled.");
```
FuriousLlama, line 38: This will show errors when the module is disabled. Instead, it needs to return an empty object sucessfully. Same on other locations

tiagograf-AOT: I don't agree, if it is trying to create a dependent when its disabled then it should error. if its trying to get the list it should just return empty.

FuriousLlama (changes requested):
Disabled modules don't need to throw errors

Apps/WebClient/src/ClientApp/src/services/restDependentService.ts
FuriousLlama: Well, I think the service layer should be agnostic of the error handling and logic of how it is used. From that point of view, getting or modifying records just means "don't do anything if disabled". Otherwise, we would need to modify the calling of said functions to depend on the availability of the module.

FuriousLlama: In other words, I think for consistency the service layer should behave the same for retrievals and for updates on disabled modules

tiagograf-AOT: So my code will say, hey service please insert this, the service will basically ignore my request and say it created it?

FuriousLlama: I agree, that it doesn't make sense. I think the problem is on the responsibility of the service to handle that logic. If the services do error when a call is made on functionality that is not enabled then all calls should error out. That way the developer does not have to figure out what kind of operation is to know if it will error out or not if the module is disabled.

FuriousLlama (approved):

leobcgov (approved):

PR #1548, opened 2021-01-04
Add the HSTS

leobcgov:
# Fixes or Implements [AB#9616](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9616)

* [ ] Enhancement
* [X] Bug
* [ ] Documentation

## Description
- Enable HSTS with a long 'max-age' ref. https://www.meziantou.net/security-headers-in-asp-net-core.htm

## Testing

* [ ] Unit Tests Updated
* [X] Not Required

### Steps to Reproduce

### UI Changes

NO

### Browsers Tested

* [X] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments


Apps/WebClient/src/Server/Startup.cs
```diff
@@ -118,6 +119,19 @@ public void ConfigureServices(IServiceCollection services)
...
             services.AddControllersWithViews();
 
+            services.AddHsts(options =>
+            {
+                options.IncludeSubDomains = true;
+                options.MaxAge = TimeSpan.FromDays(365);
+                options.Preload = true;
+            });
+            // Configure HTTPS redirection
+            services.AddHttpsRedirection(options =>
```
sslaws, line 129: This may not work in the deployed environment as currently service proxy hits port 80.  I think we may need to add work to this ticket to validate that or a better solution would be to use native TLS on all of our services which would require the cert installed.

sslaws (approved):

Apps/WebClient/src/Server/Startup.cs
```diff
@@ -118,6 +119,19 @@ public void ConfigureServices(IServiceCollection services)
             // Configure SPA
             services.AddControllersWithViews();
 
+            services.AddHsts(options =>
```
FuriousLlama, line 122: Has this worked locally? I am guessing the https protocol will break vue service connection for hot reloading

Apps/WebClient/src/Server/Startup.cs
sslaws: Hey Leo, each OpenShift service (app for us) is actually served on port 8080 (I said 80 above) as per the image build https://github.com/bcgov/healthgateway/blob/dev/Apps/WebClient/Dockerfile 

The OpenShift route does TLS termination and then forwards to the service.  So having redirection is likely going to be problematic.  If this resolves the issue then good but I'm thinking it likely will break something.

leobcgov: @sslaws I see, it is more complicated than. I think I can just remove the redirection as the 9616 doesn't ask for doing redirection right?

leobcgov: @sslaws per my local testing, we need the redirection for this work. Let's discuss the 2 options that you suggested above then (1) add work to this ticket to validate that or (2) a better solution would be to use native TLS on all of our services which would require the cert installed.

leobcgov: Per my discussion with @sslaws today, we will deploy this to DEV and I can verify this there.

Apps/WebClient/src/Server/Startup.cs
FuriousLlama: On the documentation link posted above the UserHsts method is surrounded by a check of the environment
```
if (!env.IsDevelopment())
        {
            // HSTS should only be enabled on production, not on localhost
            app.UseHsts();
        }

FuriousLlama: It does not look like we have that on our code. When running locally, did Hot reloading worked on Vue?

leobcgov: @FuriousLlama yes previously someone has already added it:
image

leobcgov: for "When running locally, did Hot reloading worked on Vue?", yes it does for me.

FuriousLlama: got it, thanks for checking!

FuriousLlama (approved):

tiagograf-AOT (approved):

PR #1592, opened 2021-01-11
Tasks/9772

FuriousLlama:
# Fixes or Implements [AB#9772](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9772)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Updates the backend to provide a list of dates with the corresponding user count for a given offset

## Testing

* [ ] Unit Tests Updated
* [x] Not Required


Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -160,16 +160,20 @@ public int GetRegisteredUsersCount()
...
+            var unionResult = this.dbContext.UserProfile
+                .Select(x => new { HdId = x.HdId, lastLoginDate = x.LastLoginDateTime })
+                .Union(this.dbContext.UserProfileHistory
+                .Select(x => new { HdId = x.HdId, lastLoginDate = x.LastLoginDateTime })).ToList();
+
+            Dictionary<DateTime, int> dateCount = unionResult.Select(x => new
+            {
+                HdId = x.HdId,
+                lastLoginDate = x.lastLoginDate.Add(offset).Date,
+            }).Distinct().GroupBy(x => x.lastLoginDate).OrderBy(x => x.Key).ToDictionary(x => x.Key, x => x.Count());
```
leobcgov, line 174: nice query!

Apps/AdminWebClient/src/Server/Controllers/DashboardController.cs
```diff
@@ -61,16 +61,16 @@ public IActionResult GetRegisteredUserCount()
...
-        /// <param name="offset">The offset from the client browser to UTC.</param>
+        /// <param name="timeOffset">The offset from the client browser to UTC.</param>
         /// <returns>The count of logged in users in the current day.</returns>
         /// <response code="200">Returns the list of user feedbacks.</response>
         /// <response code="401">The client must authenticate itself to get the requested response.</response>
         /// <response code="403">The client does not have access rights to the content; that is, it is unauthorized, so the server is refusing to give the requested resource. Unlike 401, the client's identity is known to the server.</response>
         [HttpGet]
         [Route("LoggedInCount")]
-        public IActionResult GetTodayLoggedinUsersCount(int offset)
+        public IActionResult GetTodayLoggedinUsersCount(int timeOffset)
```
leobcgov, line 71: @FuriousLlama I believe we need to add the date range to this one per req: "update service and delegate to accept date ranges " on 9772 right?

leobcgov (changes requested):

Apps/AdminWebClient/src/Server/Controllers/DashboardController.cs
FuriousLlama: Instead of performing the same query multiple times with different dates, it would be better if the front end pics what date it wants to show

leobcgov: not sure if I understand the solution design for this one, as it doesn't seem to me this fulfills "in admin site, add dashboard stat for daily logins with date range selector (two dates)" requirement on the user story 9766. We may need to chat on our groom to make sure :-)

Apps/Database/src/Delegates/DBProfileDelegate.cs
sslaws: This doesn't look right, this looks like it will query and bring the entire user profile and user profile history tables back into memory and then does c# side query to filter the results.  This will be a no go for the Production DB, can you confirm?

Apps/Database/src/Delegates/IUserProfileDelegate.cs
```diff
@@ -75,18 +75,18 @@ public interface IUserProfileDelegate
...
+        IDictionary<DateTime, int> GetDailyLoggedInUsersCount(TimeSpan offset);
 
         /// <summary>
         /// Returns the list of all UserProfiles sorted by CreatedDateTime in assending order.
         /// </summary>
         /// <param name="page">The page to request.</param>
         /// <param name="pageSize">The amount of records to retrieve in 1 request.</param>
         /// <returns>A list of UserProfiles wrapped in a DBResult.</returns>
-        public DBResult<IEnumerable<UserProfile>> GetAll(int page, int pageSize);
+        DBResult<IEnumerable<UserProfile>> GetAll(int page, int pageSize);
```
sslaws, line 90: Why did you drop the DBResult?

sslaws (changes requested):

sslaws (changes requested):
I added comments in my prior review but it wasn't flagged.

Apps/Database/src/Delegates/IUserProfileDelegate.cs
FuriousLlama: not sure what you mean. Dropped public

Apps/Database/src/Delegates/DBProfileDelegate.cs
FuriousLlama: at this point, only the hdid and lastLoginDate are loaded. Since the offset needs to be executed on the c# side the queries are separated. I tried executing the date offset calculation on the first query, but EF could not translate it

FuriousLlama: Updated query!

sslaws (approved):

leobcgov (approved):

PR #1623, opened 2021-01-14
Fixes admin site not retrieving feedback

FuriousLlama:
# Fixes or Implements [AB#9881](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9881)

* [ ] Enhancement
* [x] Bug
* [ ] Documentation

## Description

The admin site was failing to retrieve feedback data if the user was no longer in the DB and had made a comment

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### Steps to Reproduce

User has a validated email
User makes a feedback comment.
User removes its account
JobScheduler runs the close account job
Open Admin user-feedback section.

### DIFF
--- a/Apps/Database/src/Delegates/DBFeedbackDelegate.cs
+++ b/Apps/Database/src/Delegates/DBFeedbackDelegate.cs
@@ -114,10 +114,13 @@ namespace HealthGateway.Database.Delegates
 
             foreach (UserFeedbackAdmin feedbackAdmin in feedback)
             {
-                string? email;
-                if (profileEmails.TryGetValue(feedbackAdmin.UserProfileId, out email))
+                if (!string.IsNullOrWhiteSpace(feedbackAdmin.UserProfileId))
                 {
-                    feedbackAdmin.Email = email != null ? email : string.Empty;
+                    string? email;
+                    if (profileEmails.TryGetValue(feedbackAdmin.UserProfileId, out email))
+                    {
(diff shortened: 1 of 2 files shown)


Apps/Database/src/Delegates/DBFeedbackDelegate.cs
```diff
@@ -114,10 +114,13 @@ public DBResult<List<UserFeedbackAdmin>> GetAllUserFeedbackEntries()
 
             foreach (UserFeedbackAdmin feedbackAdmin in feedback)
             {
-                string? email;
-                if (profileEmails.TryGetValue(feedbackAdmin.UserProfileId, out email))
+                if (!string.IsNullOrWhiteSpace(feedbackAdmin.UserProfileId))
                 {
-                    feedbackAdmin.Email = email != null ? email : string.Empty;
```
sslaws, line 120: = email ?? String.Empty;

sslaws (approved):
Approved but we should probably use the clearer syntax.

leobcgov (approved):

Apps/Database/src/Delegates/DBFeedbackDelegate.cs
FuriousLlama: agree, there are other improvements on the qery too that can be done. For now, let's keep it simple to only address the exception

PR #1692, opened 2021-01-22
Fixes expired delegate lookup AB#9666

tiagograf-AOT:
# Fixes or Implements [AB#9666](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9666)

* [ ] Enhancement
* [x] Bug
* [ ] Documentation

## Description

Fixes the laboratory lookup when a resource delegate is expired.

## Testing

* [x] Unit Tests Updated
* [x] Functional Tests Updated
* [ ] Not Required

### Steps to Reproduce

### UI Changes

NO

### Browsers Tested

* [ ] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments


Apps/Common/src/AccessManagement/Authorization/Handlers/FhirResourceAuthorizationHandler.cs
```diff
@@ -219,5 +240,27 @@ private bool ValidateObservationDelegate(AuthorizationHandlerContext context, st
...
+        {
+            if (!this.maxDependentAge.HasValue)
+            {
+                this.logger.LogInformation($"Delegate expired check on resource {resourceHDID} skipped as max dependent age is null");
+                return false;
+            }
+
+            if (this.patientService == null)
+            {
+                this.logger.LogError($"Delegate expired check on resource {resourceHDID} failed as patientService is null");
```
FuriousLlama, line 254: I think this should throw since it should never reach this place if going though the correct flow

Apps/Common/src/Models/RequestResult.cs
```diff
@@ -27,6 +27,26 @@ namespace HealthGateway.Common.Models
...
+        public RequestResult()
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RequestResult{T}"/> class.
+        /// </summary>
+        /// <param name="resourcePayload">The resource payload.</param>
+        /// <param name="resultStatus">The result status.</param>
+        public RequestResult(
```
FuriousLlama, line 42: Why is this necessary? I dont see a new RequestResult(params)

Apps/Common/src/AccessManagement/Authorization/Handlers/FhirResourceAuthorizationHandler.cs
sslaws: It shouldn't throw but you shouldn't have to do the check... Why are the delegates nullable?  Probably an easy fix when we first introduced the nullable checks but we should change them to non-nullable now and remove the check.

Apps/Laboratory/src/appsettings.hgpoc.json
```diff
@@ -10,6 +10,15 @@
     "Laboratory": {
         "EndPoint": "https://phsahealthgatewayapi-dev.azurewebsites.net/api/v1/LabOrders"
     },
+    "PatientService": {
```
sslaws, line 13: Nice of you to maintain this file!

sslaws (changes requested):

Apps/Common/src/AccessManagement/Authorization/Handlers/FhirResourceAuthorizationHandler.cs
FuriousLlama: +1

tiagograf-AOT: If this is not nullable, any project that uses the resource authorization handler would have to initialize patient services in their startup which i think its all projects at this point, but only the lab will actually reach this code that checks for delegation. I think theres no harm so far making it not nullable now since the projects won't reach this code

Apps/Common/src/Models/RequestResult.cs
tiagograf-AOT: its not necessary, its a nice to have

Apps/Laboratory/src/appsettings.hgpoc.json
tiagograf-AOT: if its there we should maintain it, otherwise lets delete it

Apps/Common/src/AccessManagement/Authorization/Handlers/FhirResourceAuthorizationHandler.cs
tiagograf-AOT: i just think its better to get this error instead of a nullreference exception

tiagograf-AOT: I just tested and i was wrong, all projects would have to initialize patient services if I make it not nullable. should i got ahead and enabled it everywhere?

tiagograf-AOT: The error:

tiagograf-AOT: Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Microsoft.AspNetCore.Authorization.IAuthorizationService Lifetime: Transient ImplementationType: Microsoft.AspNetCore.Authorization.DefaultAuthorizationService': Unable to resolve service for type 'HealthGateway.Common.Services.IPatientService' while attempting to activate

FuriousLlama: So the handler is used by all projects, but only some actually reach this code? If that is the case I think throwing an exception will be better than just logging an error since that would imply the startup was not properly set.
Is it possible to have a custom handler for the project? it can inherit from the base one

Apps/Common/src/Models/RequestResult.cs
tiagograf-AOT: its being used in the tests!

FuriousLlama: Not sure am following. Is it being used?

Apps/Common/src/AccessManagement/Authorization/Handlers/FhirResourceAuthorizationHandler.cs
tiagograf-AOT: Yes only Observation/Lab will check for delegation so far, I think we plan to do this for other projects in the future, I can look into making a new custom handler

tiagograf-AOT: New handler implemented.

Apps/Common/src/Models/RequestResult.cs
tiagograf-AOT: it just makes initializing a mock RequestResult easier (less lines) not a big deal, but i also don't see why this would be bad

sslaws (approved):

FuriousLlama (approved):

PR #1698, opened 2021-01-22
Fixes Comment entry type being empty - Bugs/9953

FuriousLlama:
# Fixes or Implements [AB#9953](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9953)

* [ ] Enhancement
* [x] Bug
* [ ] Documentation

## Description

Added entry type code table with FK for comment. Updated code to check for it and backfill it if it does not have it.

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### Steps to Reproduce

Load the timeline with comments.
Check the DB, the comments from the previous timeline should have an updated entry type code.
Create a comment,
New comment should have the required entry type code

### UI Changes

YES


leobcgov (approved):

tiagograf-AOT (approved):

Apps/AdminWebClient/src/ClientApp/src/constants/resulttype.ts
```diff
@@ -1,4 +1,4 @@
-export enum ResultType {
+export const enum ResultType {
```
sslaws, line 1: Why are you changing all of the enums to constants?  This applies to multiple types below.  I disagree with the change as it allows for errors in the code.  Please convert back and then add value convertors as required in the DB Context (we have examples of that).

sslaws (changes requested):

Apps/Database/src/Constants/CommentEntryType.cs
```diff
@@ -38,6 +43,6 @@ public static class CommentEntryType
         /// <summary>
         /// The code representing Encounter.
         /// </summary>
-        public const string Encounter = "ENC";
+        public const string Encounter = "Enc";
```
sslaws, line 46: This is existing but it should likely be converted to an enum with value convertors on the dbcontext.  We can raise a bug on this.

sslaws (changes requested):

Apps/Database/src/Constants/CommentEntryType.cs
FuriousLlama: We have a mix of enums and static classes on the DB constants. Not sure which one we should go forth, but I think we should change them all.

FuriousLlama: created 9955 to track it

Apps/AdminWebClient/src/ClientApp/src/constants/resulttype.ts
FuriousLlama: they are still enum's, only const enum's now :)

sslaws (approved):

PR #1736, opened 2021-01-28
Added route checking for access 

FuriousLlama:
# Fixes or Implements [AB#9656](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/9656)

* [ ] Enhancement
* [X] Bug
* [ ] Documentation

## Description

Re-added checking for the user state after each route change. Fixes Profile, and hopefully fix mobile as well

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### Steps to Reproduce

Open a mobile browser, check if the logout is called.

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/router.ts
+++ b/Apps/WebClient/src/ClientApp/src/router.ts
@@ -270,40 +270,44 @@ router.beforeEach(async (to, from, next) => {
         next();
     } else {
         // Make sure that the route accepts the current state
-        const currentUserState = calculateUserState();
-        logger.debug(`current state: ${currentUserState}`);
-        const isValidState = to.meta.validStates.includes(currentUserState);
-        const availableModules = getAvailableModules();
-        const hasRequiredModules =
-            to.meta.requiredModules === undefined
-                ? true
-                : to.meta.requiredModules.every((val: string) =>
-                      availableModules.includes(val)
-                  );
-        if (isValidState && hasRequiredModules) {
-            next();
-        } else {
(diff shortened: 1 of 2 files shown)


leobcgov:
> > I think we will confirm bug anyway before closing it. So I don't think we need to create a task/bug. Btw, I don't think it will fix the issue though.
> 
> Thanks for the observation. Can you elaborate why not?

I may be wrong, but from what I understand, the Iddle Logout Warning displays by idle.vue embeded in the app.vue. I don't see any relations of fixing the "auto timeout does not appear to be occurring" and the code change of this PR.

FuriousLlama:
> > > I think we will confirm bug anyway before closing it. So I don't think we need to create a task/bug. Btw, I don't think it will fix the issue though.
> > 
> > 
> > Thanks for the observation. Can you elaborate why not?
> 
> I may be wrong, but from what I understand, the Iddle Logout Warning displays by idle.vue embeded in the app.vue. I don't see any relations of fixing the "auto timeout does not appear to be occurring" and the code change of this PR.

That is a good point. When I was testing it it looked like the token was not being validated on route changes, and that could be triggered when the mobile browser comes back from suspension. This change attempts to catch that and re-validate the user coming from oidc as valid. 

Regarding the logout warning on mobile; It is showing when the window is active (the browser is on the forefront of the phone). However, when it goes to the background (home button press) the app goes into suspension, where no JS is running. We cannot expect the idle modal to show in that case since no JS is being executed

PR #1827, opened 2021-02-11
Add expiry hours logic

leobcgov:
# Fixes or Implements [AB#10049](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10049)

* [X] Enhancement
* [ ] Bug
* [ ] Documentation

## Description
- Add Expiry Hours to WebConfig.
- Update Invite Email logic.

## Testing

* [ ] Unit Tests Updated
* [X] Not Required

### Steps to Reproduce

### UI Changes

NO

### Browsers Tested

* [ ] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments


Apps/Common/src/Services/EmailQueueService.cs
```diff
@@ -128,19 +128,20 @@ public void CloneAndQueue(Guid emailId, bool shouldCommit = true)
...
             invite.InviteKey = Guid.NewGuid();
             invite.HdId = hdid;
-            invite.ExpireDate = DateTime.UtcNow.AddDays(VerificationExpiryDays);
+            invite.ExpireDate = DateTime.UtcNow.AddHours(verificationExpiryHours);
 
             string hostUrl = activationHost.ToString();
             hostUrl = hostUrl.Remove(hostUrl.Length - 1, 1); // Strips last slash
 
-            keyValues.Add(INVITE_KEY_VARIABLE, invite.InviteKey.ToString());
+            keyValues.Add(INVITE_KEY_VARIABLE, hdid);
```
FuriousLlama, line 142: we dont want to send the hdid as a key to the user. We want the same invite key as the previous email if the email is the same

Apps/Common/src/Services/IEmailQueueService.cs
```diff
@@ -79,7 +79,8 @@ public interface IEmailQueueService
         /// <param name="hdid">The users hdid to associate the invite with.</param>
         /// <param name="toEmail">The To email address.</param>
         /// <param name="activationHost">The host of the email validation endpoint.</param>
-        void QueueNewInviteEmail(string hdid, string toEmail, Uri activationHost);
+        /// <param name="verificationExpiryHours">The expiry hours of email verification.</param>
+        void QueueNewInviteEmail(string hdid, string toEmail, Uri activationHost, int verificationExpiryHours = 5);
```
FuriousLlama, line 83: no need for a default value

Apps/WebClient/src/Server/Services/UserProfileService.cs
```diff
@@ -185,7 +185,7 @@ public async Task<RequestResult<UserProfileModel>> CreateUserProfile(CreateUserR
 
                 if (!string.IsNullOrWhiteSpace(requestedEmail))
                 {
-                    this.emailQueueService.QueueNewInviteEmail(hdid, requestedEmail, hostUri);
+                    this.emailQueueService.QueueNewInviteEmail(hdid, requestedEmail, hostUri, this.configurationService.GetConfiguration().WebClient.EmailVerificationExpiryHours);
```
FuriousLlama, line 188: Lets get the configuration value on the constructor instead of here

Apps/WebClient/src/appsettings.json
```diff
@@ -41,33 +41,34 @@
...
-            "Encounter": false,
-            "Note": true,
-            "Comment": true,
-            "CovidLabResults": true,
-            "Dependent": true
-        },
-        "HoursForDeletion": 720,
-        "MinPatientAge": 12,
-        "MaxDependentAge": 12
+  "WebClient": {
```
FuriousLlama, line 44: bad indent?

Apps/WebClient/src/appsettings.json
```diff
@@ -41,33 +41,34 @@
...
+      "Encounter": false,
+      "Note": true,
+      "Comment": true,
+      "CovidLabResults": true,
+      "Dependent": true
     },
+    "HoursForDeletion": 720,
+    "MinPatientAge": 12,
+    "MaxDependentAge": 12,
+    "EmailVerificationExpiryHours": 5
```
FuriousLlama, line 70: EmailVerificationExpiryHours => MessageVerificationExpire
Also, it should be in seconds

FuriousLlama (changes requested):
The email should not have the hdid
The expire config needs to be in seconds
The Update email flow needs to be updated as well | Check UpdateUserEmail

Apps/Common/src/Services/EmailQueueService.cs
```diff
@@ -33,11 +33,11 @@ namespace HealthGateway.Common.Services
     /// </summary>
     public class EmailQueueService : IEmailQueueService
     {
-        private const int VerificationExpiryDays = 5;
 #pragma warning disable SA1310 // Disable _ in variable name
         private const string INVITE_KEY_VARIABLE = "InviteKey";
         private const string ACTIVATION_HOST_VARIABLE = "ActivationHost";
         private const string ENVIRONMENT_VARIABLE = "Environment";
+        private const string VERIFICATION_EXPIRY_HOURS = "VerificationExpiryHours";
```
tiagograf-AOT, line 40: please update it to "ExpiryHours", which i used for the template.

Apps/Common/src/Services/EmailQueueService.cs
```diff
@@ -128,19 +128,20 @@ public void CloneAndQueue(Guid emailId, bool shouldCommit = true)
...
-            invite.ExpireDate = DateTime.UtcNow.AddDays(VerificationExpiryDays);
+            invite.ExpireDate = DateTime.UtcNow.AddHours(verificationExpiryHours);
 
             string hostUrl = activationHost.ToString();
             hostUrl = hostUrl.Remove(hostUrl.Length - 1, 1); // Strips last slash
 
-            keyValues.Add(INVITE_KEY_VARIABLE, invite.InviteKey.ToString());
+            keyValues.Add(INVITE_KEY_VARIABLE, hdid);
             keyValues.Add(ACTIVATION_HOST_VARIABLE, hostUrl);
+            keyValues.Add(VERIFICATION_EXPIRY_HOURS, $"{verificationExpiryHours}");
```
tiagograf-AOT, line 144: .ToString() instead of $"{}"

Apps/WebClient/src/Server/Services/UserProfileService.cs
tiagograf-AOT: also this is where you can have a default value

Apps/Common/src/Services/EmailQueueService.cs
leobcgov: @tiagograf-AOT can you explain why .ToString() is better than $"{}" here?

Apps/WebClient/src/appsettings.json
leobcgov: I don't agree with the naming here as it is not Message but Email, the name should be meaningful. MessageVerificationExpiry doesn't tell much about if it is in minutes, hours, days, months.
Also, why it should be in seconds? we need to display this as is in the email template, why do we need to convert seconds to hours? Unless the requirement asks to allow email to expire in seconds or minutes which is not the case here right?

Apps/WebClient/src/Server/Services/UserProfileService.cs
leobcgov: > Lets get the configuration value on the constructor instead of here
I'm not sure about this though, as passing in parameter in clear like this is aligned with principles of Functional Programming, passing it via constructor will hide the param inside but I'm ok if the team wants to go with constructor.

Apps/Common/src/Services/EmailQueueService.cs
leobcgov: @FuriousLlama I agree, I have missed that we need email and user as the key here. Do you think if we can use hash function to generate this invite key for this?

FuriousLlama: I think the generated GUID  should be sufficient

Apps/Common/src/Services/EmailQueueService.cs
tiagograf-AOT: readability

Apps/WebClient/src/Server/Services/UserProfileService.cs
FuriousLlama: The variable would be defined as part of the class implementation and set on the constructor. That would not hide it since is not set as global.

Apps/WebClient/src/appsettings.json
FuriousLlama: Seconds is part of the requirements. The name is arguable

Apps/Common/src/Services/EmailQueueService.cs
leobcgov: @FuriousLlama How do we retrieve the existing verification email for a user with a specific email? Do we need to query db by email and content template of this emails? 
Also, generating GUID requires us to query DB again, but using hash doesn't require a query to DB for this scenario.

Apps/WebClient/src/Server/Services/UserProfileService.cs
leobcgov: I'm not sure about the pros of setting it in the constructor here though, passing by param to the function will make it clear that queue new email, but it expires in X hours, developer doesn't have to guess what this method is about and what details it requires processing, it is the principle of Functional Programming.

Apps/WebClient/src/appsettings.json
leobcgov: ah you re right for the seconds.
We need team discussion for the naming :-)

Apps/Common/src/Services/EmailQueueService.cs
leobcgov: ExpiryHours doesn't have any meeaning in the WebConfig, as it can have too many meaning such as Expiry SMS, Email, etc... Email template may have enought context to understand ExpiryHours but not in WebConfig. Ideally, we should use the same name everywhere as it is hard to maintain software with different names for the same parameter.

Apps/Common/src/Services/EmailQueueService.cs
leobcgov: :-) not sure about that though, mbe just our personal preference as I see `keyValues.Add(VERIFICATION_EXPIRY_HOURS, $"{verificationExpiryHours}");` is readable as well

Apps/Common/src/Services/EmailQueueService.cs
FuriousLlama: The EmailQueueService does not need to know the business logic of generating keys or expiring old ones. That is done on the UserProfileService and UserSmsService. Both already have the previous email that was sent. It just needs to check if it is the same as the new one and if it is pass the invite key

Apps/Common/src/Services/EmailQueueService.cs
FuriousLlama: The ${} syntax implies that you are doing some string operations. So when reading it we need to make sure that only the stuff we want is being made a string.
The .ToString() means that the value is being transformed into a string, no need for further checking if the string is being modified.

TlDr: ${} Is good for string composition ToString() is good for value=> string.

Apps/WebClient/src/Server/Services/UserProfileService.cs
FuriousLlama: The difference is that that value is constant at the time the constructor is called and will not change at any time during the existence of the class instance.  On the other hand, getting the configuration every time for this method is calling extra functions without adding anything.
Also, the method call is more readable without the nesting of functions

Apps/WebClient/src/appsettings.json
FuriousLlama: You can propose a different name and we can agree to that one, too

Apps/Common/src/Services/EmailQueueService.cs
leobcgov: per our team discussion, passing the existing inviteKey instead

leobcgov:
PR updated and ready.

Apps/WebClient/src/Server/Services/UserEmailService.cs
```diff
@@ -143,7 +143,7 @@ public bool UpdateUserEmail(string hdid, string email, Uri hostUri, string beare
             if (!string.IsNullOrEmpty(email))
             {
                 this.logger.LogInformation($"Sending new email invite for user ${hdid}");
-                this.emailQueueService.QueueNewInviteEmail(hdid, email, hostUri);
+                this.emailQueueService.QueueNewInviteEmail(hdid, email, hostUri, emailInvite != null ? emailInvite.InviteKey : null);
```
tiagograf-AOT, line 146: As specified in the story it can only reuse the invite key if the new email address is the same as the previous one

tiagograf-AOT (changes requested):
same email address check

Apps/WebClient/src/Server/Services/UserEmailService.cs
leobcgov: @tiagograf-AOT thanks, I though this code was called when user updates to the new email address. Forgot that it got called also for ReSend scenario.

leobcgov: code updated

tiagograf-AOT (approved):

PR #1830, opened 2021-02-11
Update Email Verification Page AB#10047

tiagograf-AOT:
# Fixes or Implements [AB#10047](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10047)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Updates the email verification page.
image
image
image

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### Steps to Reproduce

### UI Changes

YES

### Browsers Tested

* [x] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
+++ b/Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
@@ -5,6 +5,7 @@ import Vue from "vue";
 import { Component, Prop } from "vue-property-decorator";
 import { Action, Getter } from "vuex-class";
 
+import { ResultType } from "@/constants/resulttype";
 import User from "@/models/user";
 import { SERVICE_IDENTIFIER } from "@/plugins/inversify";
 import container from "@/plugins/inversify.config";
@@ -20,20 +21,29 @@ export default class ValidateEmailView extends Vue {
     @Action("checkRegistration", { namespace: "user" })
     checkRegistration!: (params: { hdid: string }) => Promise<boolean>;
 
-    private isLoading = false;
-    private isSuccess: boolean | null = null;
+    private isLoading = true;
+    private resultStatus: ResultType | null = null;
 
+    private get isSuccess() {
+        return this.resultStatus === ResultType.Success;
+    }
+    private get isActionRequired() {
+        return this.resultStatus === ResultType.ActionRequired;
+    }
     private mounted() {
+        this.verifyEmail();
+    }
+
+    private verifyEmail() {
         this.isLoading = true;
         const userProfileService: IUserProfileService = container.get(
             SERVICE_IDENTIFIER.UserProfileService
         );
-
         userProfileService
             .validateEmail(this.user.hdid, this.inviteKey)
-            .then((isValid) => {
-                this.isSuccess = isValid;
-                if (isValid) {
+            .then((result) => {
+                this.resultStatus = result.resultStatus;
+                if (this.resultStatus == ResultType.Success) {
                     this.checkRegistration({ hdid: this.user.hdid });
                     setTimeout(
                         () => this.$router.push({ path: "/timeline" }),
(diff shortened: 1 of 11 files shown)


Apps/WebClient/src/Server/Services/UserEmailService.cs
```diff
@@ -58,38 +58,44 @@ public class UserEmailService : IUserEmailService
...
+                    emailInvite.VerificationAttempts++;
+                    this.messageVerificationDelegate.Update(emailInvite);
+                }
+                retVal.ResultStatus = ResultType.Error;
+            }
+            else if (emailInvite.VerificationAttempts >= MaxVerificationAttempts ||
+                     emailInvite.ExpireDate < DateTime.UtcNow)
+            {
+                // Verification Expired
+                retVal.ResultStatus = ResultType.ActionRequired;
```
FuriousLlama, line 85: lets add some data into the error message for this

Apps/WebClient/src/Server/Services/UserEmailService.cs
```diff
@@ -58,38 +58,44 @@ public class UserEmailService : IUserEmailService
...
+            PrimitiveRequestResult<bool> retVal = new PrimitiveRequestResult<bool>();
             MessagingVerification? emailInvite = this.messageVerificationDelegate.GetByInviteKey(inviteKey);
 
-            if (emailInvite != null &&
-                emailInvite.HdId == hdid &&
-                !emailInvite.Validated &&
-                !emailInvite.Deleted &&
-                emailInvite.VerificationAttempts < MaxVerificationAttempts &&
-                emailInvite.ExpireDate >= DateTime.UtcNow)
+            if (emailInvite?.HdId != hdid ||
```
FuriousLlama, line 67: will this fail if emailInvite is null?

FuriousLlama:
Looks like some of the UI changed, can you add a screen shot of it?

Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
```diff
@@ -51,46 +61,22 @@ export default class ValidateEmailView extends Vue {
...
+            <b-col class="text-center mb-5 title">
+                <h4 v-if="isLoading">We are verifying your email...</h4>
+                <h4 v-else-if="isSuccess" class="text-success">
                     Your email was successfully verified!
                 </h4>
-                <h4
-                    v-if="!isLoading && isSuccess === false"
-                    class="text-danger"
-                >
+                <h4 v-else-if="isActionRequired">
```
leobcgov, line 69: How do we know "isActionRequired == true" will lead to expired email, it may be user has not validated invite link yet but the link is still valid (or non-expired).

Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
```diff
@@ -20,20 +21,29 @@ export default class ValidateEmailView extends Vue {
...
 
-    private isLoading = false;
-    private isSuccess: boolean | null = null;
+    private isLoading = true;
+    private resultStatus: ResultType | null = null;
 
+    private get isSuccess() {
+        return this.resultStatus === ResultType.Success;
+    }
+    private get isActionRequired() {
```
leobcgov, line 30: what does `isActionRequired` mean? (1) is the invite link expired or (2) invite link is still valid but user has not used it yet?

Apps/WebClient/src/Server/Services/UserEmailService.cs
leobcgov, line 85: why don't we name VerificationExpired instead of ActionRequired? as ActionRequired may also mean that user has not click the link but the link is still valid right?

Apps/WebClient/src/Server/Services/UserEmailService.cs
tiagograf-AOT: null != hdid will enter the if which is the same as the null check "if (null == null || null != hdid)"

Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
tiagograf-AOT: The service returns 3 states, success == "validated", error == "something wrong with the invite key", actionrequired == "expired"

Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
tiagograf-AOT: same as above

Apps/WebClient/src/Server/Services/UserEmailService.cs
FuriousLlama: ActionRequired is the type of error. No need to create a new type for every single error

Apps/WebClient/src/ClientApp/src/views/validateEmail.vue
FuriousLlama: Looks like isActionRequired is just a wrapper for the internal checking of the result type for convenience. So the name should match that (as it does)

Apps/WebClient/src/Server/Services/UserEmailService.cs
tiagograf-AOT: resulttypes are generic and used for common error handling and its not specific to the email verification

tiagograf-AOT: updated

Apps/WebClient/src/Server/Services/UserEmailService.cs
FuriousLlama: Yeah I agree, however I would add a emailInvite == null and then the ? is not needed/

FuriousLlama (approved):

leobcgov (approved):

PR #1878, opened 2021-02-20
Refactor timeline for Vuex - Features/10005

FuriousLlama:
# Fixes or Implements [AB#10005](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10005)

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description
- Updated stores
- Updated Timeline to work with vuex 
- Improved timeline fetching code 
- Added error handling on vuex data store side
- Removed most events using the bus from timeline 
- Added idle store and improved state logic 
- Updated Export page and streamlined logic

## Testing

* [ ] Unit Tests Updated
* [x] Not Required

### UI Changes

 NO

### Browsers Tested

* [x] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

### DIFF
--- a/Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
+++ b/Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
@@ -84,7 +84,7 @@ namespace HealthGateway.Medication.Delegates
                     { "servicePort", servicePort! },
                 };
 
-                this.baseURL = new Uri(StringManipulator.Replace(this.odrConfig.BaseEndpoint, replacementData) !);
+                this.baseURL = new Uri(StringManipulator.Replace(this.odrConfig.BaseEndpoint, replacementData)!);
             }
             else
             {
@@ -192,9 +192,12 @@ namespace HealthGateway.Medication.Delegates
         /// <inheritdoc/>
         public bool ValidateProtectiveWord(string phn, string? protectiveWord, string hdid, string ipAddress)
         {
+            // Only for testing
+            //return protectiveWord == "ASDASD";
+
             using (Source.StartActivity("ValidateProtectiveWord"))
             {
-                bool retVal = false;
+                bool isValid = false;
                 try
                 {
                     IHash? cacheHash = null;
@@ -223,12 +226,12 @@ namespace HealthGateway.Medication.Delegates
                             }
                         }
 
-                        retVal = this.hashDelegate.Compare(protectiveWord, hash);
+                        isValid = this.hashDelegate.Compare(protectiveWord, hash);
                     }
                     else
                     {
                         this.logger.LogDebug("Validating Cached Protective Word");
-                        retVal = this.hashDelegate.Compare(protectiveWord, cacheHash);
+                        isValid = this.hashDelegate.Compare(protectiveWord, cacheHash);
                     }
                 }
 #pragma warning disable CA1031 // We want to fail on any exception
@@ -238,7 +241,7 @@ namespace HealthGateway.Medication.Delegates
                     this.logger.LogError($"Error getting protected word {e.ToString()}");
                 }
 
-                return retVal;
+                return isValid;
             }
         }
 
--- a/Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
+++ b/Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
@@ -192,9 +192,6 @@ namespace HealthGateway.Medication.Delegates
         /// <inheritdoc/>
         public bool ValidateProtectiveWord(string phn, string? protectiveWord, string hdid, string ipAddress)
         {
-            // Only for testing
-            return protectiveWord == "ASDASD";
-
             using (Source.StartActivity("ValidateProtectiveWord"))
             {
                 bool isValid = false;
--- a/Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
+++ b/Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
@@ -193,7 +193,7 @@ namespace HealthGateway.Medication.Delegates
         public bool ValidateProtectiveWord(string phn, string? protectiveWord, string hdid, string ipAddress)
         {
             // Only for testing
-            //return protectiveWord == "ASDASD";
+            return protectiveWord == "ASDASD";
 
             using (Source.StartActivity("ValidateProtectiveWord"))
             {
--- a/Apps/WebClient/src/ClientApp/src/components/modal/covid.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/modal/covid.vue
@@ -1,15 +1,13 @@
 ﻿<script lang="ts">
 import Vue from "vue";
-import { Component, Emit, Prop, Watch } from "vue-property-decorator";
+import { Component, Emit, Prop } from "vue-property-decorator";
 import { Action, Getter } from "vuex-class";
 
 import UserPreferenceType from "@/constants/userPreferenceType";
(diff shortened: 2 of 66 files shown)


Apps/WebClient/src/ClientApp/src/models/storeState.ts
```diff
@@ -10,11 +10,17 @@ import MedicationStatementHistory from "@/models/medicationStatementHistory";
 import User from "@/models/user";
 import { UserComment } from "@/models/userComment";
 
-export enum StateType {
+import Encounter from "./encounter";
+import { ResultError } from "./requestResult";
+import UserNote from "./userNote";
+
+export enum LoadStatus {
```
FuriousLlama, line 17: This enum contains the magic to find out what is the state of a data source.

Apps/WebClient/src/ClientApp/src/views/timeline.vue
```diff
@@ -124,102 +124,124 @@ export default class TimelineView extends Vue {
...
-    private immunizationLoadDeferred = false;
-    private immunizationLoadReady = false;
+    @Watch("immunizationIsDeferred")
+    private whenImmunizationIsDeferred(isDeferred: boolean) {
+        if (isDeferred) {
+            this.immunizationNeedsInput = true;
+        }
+    }
+
+    private get timelineEntries(): TimelineEntry[] {
```
FuriousLlama, line 134: The entries are no longer stored in a stateful way inside the timeline; Instead, they are aggregated from the data sources when they change. This simplifies the logic of keeping the timeline updated when changes happen outside of it

Apps/WebClient/src/ClientApp/src/views/timeline.vue
```diff
@@ -124,102 +124,124 @@ export default class TimelineView extends Vue {
...
-            EventMessageName.IdleLogoutWarning,
-            (isVisible: boolean) => {
-                this.idleLogoutWarning = isVisible;
-            }
-        );
+    private get hasNewTermsOfService(): boolean {
+        return this.user.hasTermsOfServiceUpdated;
+    }
 
-        this.eventBus.$on(
```
FuriousLlama, line 183: Bus no longer needed for keeping the timeline entries up to date

Apps/WebClient/src/ClientApp/src/views/timeline.vue
```diff
@@ -346,347 +308,25 @@ export default class TimelineView extends Vue {
...
+        Promise.all([
+            this.retrieveMedications({ hdid: this.user.hdid }),
+            this.retrieveImmunizations({ hdid: this.user.hdid }),
+            this.retrieveLaboratory({ hdid: this.user.hdid }),
+            this.retrieveEncounters({ hdid: this.user.hdid }),
+            this.retrieveNotes({ hdid: this.user.hdid }),
+            this.retrieveComments({ hdid: this.user.hdid }),
+        ]).catch((err) => {
+            this.logger.error(`Error loading timeline data: ${err}`);
+        });
```
FuriousLlama, line 321: All data is retrieved and stored in VUEX if it was not there before. That way there is no need to handle the state of the retrieval in the timeline

Apps/WebClient/src/ClientApp/src/components/modal/covid.vue
```diff
@@ -24,104 +22,102 @@ export default class CovidModalComponent extends Vue {
...
+    getPreference!: (preferenceName: string) => UserPreference | undefined;
 
-    private logger!: ILogger;
+    @Prop({ default: false }) isLoading!: boolean;
 
-    private mounted() {
-        this.logger = container.get<ILogger>(SERVICE_IDENTIFIER.Logger);
+    private isDismissed = false;
+
+    private get isVisible(): boolean {
```
FuriousLlama, line 35: The logic for showing this modal now resides on it instead of the Timeline. That way there is no need to sync the events on the timeline with the state required for showing the covid alert modal

Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
```diff
@@ -194,7 +194,7 @@ public bool ValidateProtectiveWord(string phn, string? protectiveWord, string hd
         {
             using (Source.StartActivity("ValidateProtectiveWord"))
             {
-                bool retVal = false;
+                bool isValid = false;
```
sslaws, line 197: NIT: Why are we changing server side code at all and why change this? I don’t think it has added any clarity since it’s a return value which is documented.

sslaws (approved):

sslaws (changes requested):
Front end compile fails.

Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
FuriousLlama: I was trying to figure out where the protected word magic happens and thought the variable rename would make it much easier to understand this method. Not deadset on it, if retval is more clear that we can keep it :)

sslaws: It's just a NIT, I'm ok with the commit.

sslaws (approved):

tiagograf-AOT (approved):

Apps/WebClient/src/ClientApp/src/components/modal/covid.vue
```diff
@@ -24,104 +22,102 @@ export default class CovidModalComponent extends Vue {
...
-
-    private actionCovidModal() {
-        this.logger.debug("Actioning Covid Modal...");
-        const actionedCovidPreference = UserPreferenceType.ActionedCovidModalAt;
+    private updateCovidPreference() {
+        const preferenceName = UserPreferenceType.ActionedCovidModalAt;
         let isoNow = new DateWrapper().toISO();
-        if (this.user.preferences[actionedCovidPreference] != undefined) {
-            this.user.preferences[actionedCovidPreference].value = isoNow;
+        if (this.user.preferences[preferenceName] != undefined) {
```
leobcgov, line 63: when we read code, I prefer actionedCovidPreference is more descritive and provides more contexts than preferenceName.

Apps/WebClient/src/ClientApp/src/store/modules/comment/actions.ts
```diff
@@ -1,56 +1,44 @@
...
 const commentService: IUserCommentService = container.get<IUserCommentService>(
     SERVICE_IDENTIFIER.UserCommentService
 );
 
 export const actions: ActionTree<CommentState, RootState> = {
-    retrieveProfileComments(
-        context,
-        params: { hdid: string }
-    ): Promise<RequestResult<Dictionary<UserComment[]>>> {
+    retrieve(context, params: { hdid: string }): Promise<void> {
```
leobcgov, line 18: retrieve method always returns void doesn't make sense. Why don't we just return the same object regardless if we commit vuex or not? Or come up with a name that means we commit only to the Vuex store.
Also, in some other retrieve of the other services i.e. lab... we don't return void.

leobcgov (changes requested):

Apps/Medication/src/Delegates/RestMedStatementDelegate.cs
leobcgov: isValid may get lost in a complex function isValid can be just a temporary variable. I'm ok with retVal as in this function it makes us think of the name of function which returns the state of validation of Protective Word.

PR #1899, opened 2021-02-23
Add robots asset AB#10128

tiagograf-AOT:
# Fixes or Implements [AB#10128](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10128)

* [ ] Enhancement
* [x] Bug
* [ ] Documentation

## Description

Adds robot.txt as a asset.

## Testing

* [ ] Unit Tests Updated
* [ ] Not Required

### Steps to Reproduce

### UI Changes

YES | NO

### Browsers Tested

* [ ] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments

### DIFF
--- a/Apps/WebClient/src/Server/Controllers/RobotsController.cs
+++ b/Apps/WebClient/src/Server/Controllers/RobotsController.cs
@@ -25,11 +25,6 @@ namespace HealthGateway.WebClient.Controllers
     /// </summary>
     public class RobotsController : Controller
     {
-        /// <summary>
-        /// The Default Robots.txt content if not overriden using the robots.txt configuration value.
-        /// </summary>
-        public const string DefaultRobotsContent = "# Default robots.txt for Non-Prod\nUser-agent: *\nDisallow: /\n";
-
         private readonly IConfiguration configuration;
         private readonly string robotsContent;
 
@@ -40,7 +35,8 @@ namespace HealthGateway.WebClient.Controllers
         public RobotsController(IConfiguration configuration)
         {
             this.configuration = configuration;
-            this.robotsContent = this.configuration.GetValue("robots.txt", DefaultRobotsContent);
+            string robotsAssetContent = Common.Utils.AssetReader.Read(@"HealthGateway.WebClient.Server.Assets.Robots.txt");
+            this.robotsContent = this.configuration.GetValue("robots.txt", robotsAssetContent);
         }
 
         /// <summary>
--- a/Apps/WebClient/src/Server/Controllers/RobotsController.cs
+++ b/Apps/WebClient/src/Server/Controllers/RobotsController.cs
@@ -15,6 +15,7 @@
 //-------------------------------------------------------------------------
 namespace HealthGateway.WebClient.Controllers
 {
+    using System;
     using System.Net.Mime;
     using Microsoft.AspNetCore.Http;
     using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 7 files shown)


Apps/WebClient/src/Server/Controllers/RobotsController.cs
```diff
@@ -40,7 +35,8 @@ public class RobotsController : Controller
         public RobotsController(IConfiguration configuration)
         {
             this.configuration = configuration;
-            this.robotsContent = this.configuration.GetValue("robots.txt", DefaultRobotsContent);
```
sslaws, line 43: We should have a default robots which has the old contents (no indexing) and the asset should be read based on the environment key so that we can deliver different robots as needed.

Apps/WebClient/src/Server/Controllers/RobotsController.cs
```diff
@@ -40,7 +35,8 @@ public class RobotsController : Controller
...
+            string robotsAssetContent = Common.Utils.AssetReader.Read(@"HealthGateway.WebClient.Server.Assets.Robots.txt");
```
sslaws, line 38: The asset should probably only be read once on controller initialization.

sslaws (changes requested):

Apps/WebClient/src/Server/Controllers/RobotsController.cs
tiagograf-AOT: so the default is now whatever is in the file, it still allows the configuration key to be added with whatever custom we want and it should replace the default.

Apps/WebClient/src/Server/Controllers/RobotsController.cs
tiagograf-AOT: since we create a transient controller its only initialized once per pod.

Apps/WebClient/src/Server/Controllers/RobotsController.cs
sslaws: so we need to have two checked in, the prod setting and the other environments.

Apps/WebClient/src/Server/Controllers/RobotsController.cs
sslaws: Perfect, missed that!

Apps/WebClient/src/Server/Controllers/RobotsController.cs
```diff
@@ -25,22 +26,20 @@ namespace HealthGateway.WebClient.Controllers
...
         /// </summary>
         /// <param name="configuration">The injected configuration.</param>
         public RobotsController(IConfiguration configuration)
         {
+            string environment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") ?? "Development";
             this.configuration = configuration;
-            this.robotsContent = this.configuration.GetValue("robots.txt", DefaultRobotsContent);
+            string? defaultRobotsAssetContent = Common.Utils.AssetReader.Read("HealthGateway.WebClient.Server.Assets.Robots.txt");
+            string? envRobotsAssetContent = Common.Utils.AssetReader.Read($"HealthGateway.WebClient.Server.Assets.Robots.{environment}.txt");
+            this.robotsContent = this.configuration.GetValue("robots.txt", envRobotsAssetContent ?? defaultRobotsAssetContent);
```
sslaws, line 42: I don't think we need the config robots.txt at this point but it doesn't hurt!

sslaws (approved):

FuriousLlama (approved):

PR #1973, opened 2021-03-04
update covid card and vuex

leobcgov:
# Fixes or Implements [AB#10246](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10246)

* [X] Enhancement
* [ ] Bug
* [ ] Documentation

## Description
- Add patient data (phn, name, dob) to vuex store.
- Add DoB to covid card.
- Use user name from patien data for covid card.
- Add wrapping and breaking text logics.
image

## Testing

* [ ] Unit Tests Updated
* [ ] Not Required

### Steps to Reproduce

### UI Changes

YES

### Browsers Tested

* [X] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments


Apps/WebClient/src/ClientApp/src/components/modal/immunizationCard.vue
```diff
@@ -327,8 +356,8 @@ div[class*=" row"] {
     border-radius: 15px 0px 0px 15px;
 
     .left-pane {
-        max-width: 75px;
-        padding: 20px;
+        max-width: 65px;
```
leobcgov, line 359: new yyyy-MMM-dd (i.e. 2021-Jan-06) display format makes the date showing on 2 lines instead of 1 line on small small such as iPhone6. These css changes fix the issue.

Apps/WebClient/src/ClientApp/src/views/reports.vue
```diff
@@ -36,10 +35,15 @@ Vue.component("BFormTag", BFormTag);
     },
 })
 export default class ReportsView extends Vue {
+    @Action("retrieve", { namespace: "patient" })
```
leobcgov, line 38: I have used this opportunity to refactor the reports which is not required for this task/story, but I can roll these changes back if needed.

leobcgov (commented):
Leo added some comments for explaining some changes.

Apps/WebClient/src/ClientApp/src/components/modal/immunizationCard.vue
```diff
@@ -119,6 +127,13 @@ export default class ImmunizationCardComponent extends Vue {
         this.authenticationService = container.get<IAuthenticationService>(
             SERVICE_IDENTIFIER.AuthenticationService
         );
+
+        this.logger = container.get<ILogger>(SERVICE_IDENTIFIER.Logger);
+        this.retrievePatientData({
+            hdid: this.oidcUser?.hdid === undefined ? "" : this.oidcUser?.hdid,
```
tiagograf-AOT, line 133: why is the hdid nullable here?

Apps/WebClient/src/ClientApp/src/components/modal/immunizationCard.vue
```diff
@@ -46,8 +54,11 @@ export default class ImmunizationCardComponent extends Vue {
     private doses: Dose[] = [];
 
     private get userName(): string {
-        return this.oidcUser
-            ? this.oidcUser.given_name + " " + this.oidcUser.family_name
+        return this.patientData
+            ? this.patientData.firstname +
+                  " " +
+                  this.patientData.lastname +
+                  "ThisIsASuperDuperLongLastName"
```
tiagograf-AOT, line 61: remove this!

tiagograf-AOT (changes requested):
remove test code

Apps/WebClient/src/ClientApp/src/components/modal/immunizationCard.vue
leobcgov: oh man, testing code, thanks Tiago!

leobcgov:
updates per @tiagograf-AOT and @FuriousLlama 's comments

Apps/WebClient/src/ClientApp/src/views/reports.vue
FuriousLlama: +1 For refactoring!

Apps/WebClient/src/ClientApp/src/store/modules/patient/patient.ts
```diff
@@ -0,0 +1,25 @@
+import { Module } from "vuex";
```
FuriousLlama, line 1: Instead of a new store, I think is better to add the data to the User one

Apps/WebClient/src/ClientApp/src/models/storeState.ts
```diff
@@ -70,6 +71,13 @@ export interface UserState {
     status: LoadStatus;
 }
 
+export interface PatientState {
+    patientData: PatientData;
```
FuriousLlama, line 75: I think is better to add the patient data to the user store

Apps/WebClient/src/ClientApp/src/components/modal/immunizationCard.vue
```diff
@@ -27,27 +28,34 @@ interface Dose {
...
 
     private eventBus = EventBus;
 
+    private logger!: ILogger;
     private readonly modalId: string = "covid-card-modal";
     private isVisible = false;
 
     private authenticationService!: IAuthenticationService;
-    private oidcUser: OidcUserProfile | null = null;
+    private oidcUser!: OidcUserProfile;
```
FuriousLlama, line 52: I dont think this is needed anymore

FuriousLlama (changes requested):
User Store should have user data, including the one from patient

Apps/WebClient/src/ClientApp/src/components/modal/immunizationCard.vue
```diff
@@ -139,10 +145,19 @@ export default class ImmunizationCardComponent extends Vue {
         this.authenticationService.getOidcUserProfile().then((oidcUser) => {
```
FuriousLlama, line 145: no need for auth here anymore

Apps/WebClient/src/ClientApp/src/components/report/header.vue
```diff
@@ -80,7 +84,9 @@ export default class ReportHeaderComponent extends Vue {
             <b-col>
                 <strong>Date of Birth:</strong>
                 <span class="px-1">{{
-                    patientData !== null ? patientData.birthdate : ""
+                    patientData !== null
```
FuriousLlama, line 87: why would the date be null?

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
```diff
@@ -241,4 +250,27 @@ export const actions: ActionTree<UserState, RootState> = {
                 });
         });
     },
+    getPatientData(context, params: { hdid: string }): Promise<void> {
+        return new Promise((resolve, reject) => {
```
FuriousLlama, line 254: if retrieved, no need to call again. Lets add a check to make sure the store is not loaded before requesting again

Apps/WebClient/src/ClientApp/src/store/modules/user/mutations.ts
```diff
@@ -90,6 +91,12 @@ export const mutations: MutationTree<UserState> = {
         state.statusMessage = "success";
         state.status = LoadStatus.LOADED;
     },
+    setPatientData(state: UserState, patientData: PatientData) {
+        state.patientData = patientData;
+        state.error = false;
+        state.statusMessage = "success";
+        state.status = LoadStatus.LOADED;
```
FuriousLlama, line 98: this is tricky since having the patient data loaded does not mean that the store is loaded (could be missing user data). Lets add a status of Partial_Loaded to make that difference

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
```diff
@@ -241,4 +250,27 @@ export const actions: ActionTree<UserState, RootState> = {
                 });
         });
     },
+    getPatientData(context, params: { hdid: string }): Promise<void> {
```
FuriousLlama, line 253: lets make an action that loads both userProfile and userPatientData. That one can be called everywhere else without having to coordinate what service it uses to load it.

leobcgov: The better design would be checking if object is null in the getters.ts, then retrieve it, if not return it so that we don't have to worry about initializing a call just for getting data. Currently, we have to explicitly call the action at least once to get data.
This is how Singleton design pattern is recommended. But if we vote for the Singleton design pattern, we need to change our code for store.

Apps/WebClient/src/ClientApp/src/store/modules/user/mutations.ts
leobcgov: added PARTIALLY_LOADED status.

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
leobcgov: Added the check to load PatientData only once.

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
leobcgov: Added the check for loading Patient Data only once in case we call it multi time.

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
leobcgov: Also the userProfile call may need to be called many time depending on the user registration status and user preference. Howerver, the userPatientData should be called only once. Getting them into a single action is not ideal.

tiagograf-AOT:
Can we display the label/value in one line only? (not wrapped)

tiagograf-AOT (changes requested):
label is wrapped

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
FuriousLlama: Vuex Getters are not meant to do logic to retrieve. If we want to retrieve on request we can use the Actions, That is part of the reason we check on the retrieve.

The reason for using getters vs actions is that Getters provide a reactive value. So the Vue class will automatically update if the getter value changes

Apps/WebClient/src/ClientApp/src/store/modules/user/user.ts
```diff
@@ -10,6 +11,7 @@ import { mutations } from "./mutations";
 export const state: UserState = {
     statusMessage: "",
     user: new User(),
+    patientData: new PatientData(),
     error: false,
     status: LoadStatus.LOADED,
```
FuriousLlama, line 16: status None since is not initialized

Apps/WebClient/src/ClientApp/src/store/modules/user/mutations.ts
FuriousLlama: Is that change updated? Dont see it here yet

Apps/WebClient/src/ClientApp/src/store/modules/user/user.ts
leobcgov: > Can we display the label/value in one line only? (not wrapped)

@tiagograf-AOT can you elaborate on this? Is this comment for a specific line of code?

leobcgov:
> Can we display the label/value in one line only? (not wrapped)
@tiagograf-AOT can you elaborate on this? Is this comment for a specific line of code?

Apps/WebClient/src/ClientApp/src/store/modules/user/user.ts
leobcgov: weird, how can our current code work even with this flag as it is always there? Do we use this status to check at all?

leobcgov: fixed code, it seems to work well as well with NONE LoadStatus

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
leobcgov: We can have a wrapper/manager for each Vuex (user, medication...) to ease the load, such as getting value from the Vuex Getters, if empty, do the Action to populate data so that we don't have to worry about which area in the application should retrieve data.

Apps/WebClient/src/ClientApp/src/store/modules/user/mutations.ts
leobcgov: my bad, mbe I have pushed it after my comment above

tiagograf-AOT:
image
date of birth is displayed in two lines

Apps/WebClient/src/ClientApp/src/store/modules/user/mutations.ts
```diff
@@ -90,6 +91,12 @@ export const mutations: MutationTree<UserState> = {
         state.statusMessage = "success";
         state.status = LoadStatus.LOADED;
     },
+    setPatientData(state: UserState, patientData: PatientData) {
+        state.patientData = patientData;
+        state.error = false;
+        state.statusMessage = "success";
+        state.status = LoadStatus.PARTIALLY_LOADED;
```
FuriousLlama, line 98: This needs a bit more logic: If user is loaded and patient is loaded then the status is loaded, otherwise partially_loaded. Same for setUserData

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
FuriousLlama: I dont think that would be necessary since the action function will return a quick promise and we can only call the getter to retrieve the data.

FuriousLlama: Btw, is OK if a single action fills both sets of data

leobcgov:
> image
> date of birth is displayed in two lines

Thanks @tiagograf-AOT , I forgot to update the screenshot for this one.

Apps/WebClient/src/ClientApp/src/store/modules/user/mutations.ts
leobcgov: @FuriousLlama I'm not sure why we want to merge these statuses. i.e. Why do we say LOADED when setUserPreference is completed and PARTIALLY_LOADED here? I think we need to discuss for this design as what we load and when and how do we handle different calls for user. Currently, I see all the other calls making LoadStatus to LOADED.
image

FuriousLlama: We dont know which set of data will be loaded first, thats why the each mutator needs to check it before setting the store state. We will need to refactor the other stores as we go. The data stores we have (meds, lab, etc) all use this logic

FuriousLlama (approved):

tiagograf-AOT (approved):

PR #2077, opened 2021-03-22
Updated launch settings to run on the network interface. AB#10331

FuriousLlama:
# Fixes or Implements AB#10331

* [x] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

Allows the locally hosted servers to be reachable on the local network,

### DIFF
--- a/Apps/Medication/test/unit/Properties/launchSettings.json
+++ b/Apps/Medication/test/unit/Properties/launchSettings.json
@@ -10,7 +10,7 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
-      "applicationUrl": "https://localhost:5001;http://localhost:5000"
+      "applicationUrl": "https://0.0.0.0:5001;http://0.0.0.0:5000"
     }
   }
 }
--- a/Apps/WebClient/src/appsettings.Development.json
+++ b/Apps/WebClient/src/appsettings.Development.json
@@ -12,8 +12,8 @@
         "ClientId": "healthgateway-local",
         "RequireHttpsMetadata": "false",
         "Callbacks": {
-            "Logon": "http://localhost:5000/loginCallback",
-            "Logout": "http://localhost:5000/logout"
+            "Logon": "http://0.0.0.0:5000/loginCallback",
+            "Logout": "http://0.0.0.0:5000/logout"
         }
     },
     "ForwardProxies": {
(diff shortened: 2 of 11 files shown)


Apps/Medication/test/unit/Properties/launchSettings.json
```diff
@@ -10,7 +10,7 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
-      "applicationUrl": "https://localhost:5001;http://localhost:5000"
+      "applicationUrl": "https://0.0.0.0:5001;http://0.0.0.0:5000"
```
sslaws, line 13: Since we're fixing this we should remove the 5001 port as we don't support HTTPS locally.

Apps/WebClient/src/appsettings.Development.json
```diff
@@ -12,8 +12,8 @@
         "ClientId": "healthgateway-local",
         "RequireHttpsMetadata": "false",
         "Callbacks": {
-            "Logon": "http://localhost:5000/loginCallback",
-            "Logout": "http://localhost:5000/logout"
+            "Logon": "http://0.0.0.0:5000/loginCallback",
```
sslaws, line 15: Was this tested?  Generally 0.0.0.0 means bind to all network interfaces but it isn't an actual interface that be be reached.

Apps/WebClient/src/appsettings.Development.json
```diff
@@ -45,11 +45,11 @@
         }
     },
     "ServiceEndpoints": {
-        "Immunization": "http://localhost:3001/",
-        "Patient": "http://localhost:3002/",
-        "Medication": "http://localhost:3003/",
-        "Laboratory": "http://localhost:3004/",
-        "Encounter": "http://localhost:3005/"
+        "Immunization": "http://0.0.0.0:3001/",
```
sslaws, line 48: ditto comment above

sslaws (changes requested):

Apps/WebClient/src/appsettings.Development.json
FuriousLlama: yeah, with that address it binded to the ip address of my pc. I dont know if it will work for mac. but I think it will.

Apps/Medication/test/unit/Properties/launchSettings.json
FuriousLlama: agreed, its removed!

Apps/WebClient/src/appsettings.Development.json
sslaws: so the binding works but you can't hit http://0.0.0.0:5000 or any of the services as its not valid.  I just tested by changing Lab launch settings.  You can hit http://localhost:3004/swagger and https://127.0.0.1/swagger and https://local_ip/swagger but not http://0.0.0.0.  So the changes to the config here I don't think can work for the call back or the endpoints and specifying localhost or anything not your external IP will not work for what we want.  I do think that we should change the launch settings though as it makes it more universal but the endpoints here should remain.

sslaws (changes requested):
As per PR thread, don't think this will work.

sslaws (approved):

tiagograf-AOT (approved):

PR #2113, opened 2021-03-27
Migrate Profile Page to Design System AB#10378

MikeLyttle:
# Fixes and Implements [AB#10378](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10378)

* [x] Enhancement
* [x] Bug
* [ ] Documentation

## Description

Creates reusable components for buttons and status labels matching the Design System, updates the profile page to use them, and moves generic styles out of the profile page.  Also fixes the email and SMS verify buttons misaligned with the input by a pixel or two.

## Testing

* [ ] Unit Tests Updated
* [ ] Not Required

### Steps to Reproduce

### UI Changes

YES

### Browsers Tested

* [x] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/views/profile.vue
+++ b/Apps/WebClient/src/ClientApp/src/views/profile.vue
@@ -22,6 +22,7 @@ import LoadingComponent from "@/components/loading.vue";
 import VerifySMSComponent from "@/components/modal/verifySMS.vue";
 import BaseButtonComponent from "@/components/shared/baseButton.vue";
 import PageTitleComponent from "@/components/shared/pageTitle.vue";
+import StatusLabelComponent from "@/components/shared/statusLabel.vue";
 import BannerError from "@/models/bannerError";
 import type { WebClientConfiguration } from "@/models/configData";
 import { DateWrapper } from "@/models/dateWrapper";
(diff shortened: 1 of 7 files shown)


Apps/WebClient/src/ClientApp/src/views/profile.vue
```diff
@@ -543,15 +547,15 @@ export default class ProfileView extends Vue {
...
+                                                <BaseButtonComponent
                                                     v-if="
                                                         !emailVerified &&
                                                         !isEmailEditable &&
                                                         email
                                                     "
                                                     id="resendEmail"
                                                     data-testid="resendEmailBtn"
-                                                    variant="outline-primary"
+                                                    button-type="secondary"
```
tiagograf-AOT, line 558: if those values are going to match bootstrap variants (primary, secondary, etc) maybe we could you the same property name? variant="secondary" instead of button-type="secondary"

tiagograf-AOT (approved):

FuriousLlama (approved):

Apps/WebClient/src/ClientApp/src/views/profile.vue
MikeLyttle: Thanks, I think I will update it to "variant" after thinking about it some more.  We currently only support "primary" and "secondary" in the design system, but it will probably be updated to include at least "warning" and "danger" in the future, so it may be better to stick with something familiar.

PR #2144, opened 2021-03-31
Reformat code with prettier

MikeLyttle:
# Code Cleanup (Reformatting)

* [ ] Enhancement
* [ ] Bug
* [ ] Documentation

## Description

.editorconfig has been moved from the Apps folder to the root directory so that its settings will apply across the repo, including folders like Testing.  Code formatting has been applied to the entire repo using prettier (`npx prettier --write .`).  Prettier has built-in formatting preferences and also respects the end-of-line and indentation settings that are specified in .editorconfig.

## Testing

* [ ] Unit Tests Updated
* [ ] Not Required

### Steps to Reproduce

### UI Changes

NO

### Browsers Tested

* [ ] Chrome
* [ ] Safari
* [ ] Edge
* [ ] Firefox

## Notes/Additional Comments


MikeLyttle:
I noticed that the various .csproj files each include a reference to the shared .editorconfig file.  Since the file has been moved, the references are now broken and will need to be updated.

.github/dependabot.yml
```diff
@@ -5,32 +5,32 @@
...
-      - "leobcgov"
+    - package-ecosystem: "nuget" # See documentation for possible values
+      directory: "/Apps" # Location of package manifests
+      schedule:
+          interval: "daily"
+      reviewers:
+          - "sslaws"
+          - "FuriousLlama"
+          - "tiagograf-AOT"
+          - "leobcgov"
```
FuriousLlama, line 16: should remove leo and and add mike

Apps/AdminWebClient/src/ClientApp/src/components/core/modals/EmailModal.vue
```diff
@@ -17,7 +17,7 @@
                             <v-text-field
                                 v-model="editedItem.subject"
                                 label="Subject"
-                                :rules="[v => !!v || 'Subject is required']"
+                                :rules="[(v) => !!v || 'Subject is required']"
```
FuriousLlama, line 20: shouldn't the local lint have catch this?

FuriousLlama:
I am curious about the linter, is it following any specific eslint rules or just prettier?

MikeLyttle:
> I am curious about the linter, is it following any specific eslint rules or just prettier?

That's a good question.  It's only applying prettier.  When we lint, prettier style rules are included, but other things are as well, so it's possible that using prettier alone is not achieving the same result.

Perhaps a better solution would be to exclude all of the .ts and .vue files from this generic prettier formatting, and instead add linting to the e2e package and any others that do not have it and run that for them.

.github/dependabot.yml
sslaws: I've updated this

Apps/AdminWebClient/src/ClientApp/src/components/core/modals/EmailModal.vue
MikeLyttle: Yes.  I did some digging and there seems to be something wrong with the prettier formatting in the AdminWebClient ClientApp.  It removes braces from inline functions as seen above.  Also, it removes the trailing comma on the last item in imports and array/object declarations and doesn't format those lists properly if you mess with the whitespace/newlines.

I wasn't able to figure out a way to resolve the underlying issue, but if you run prettier from a directory higher than ClientApp, it will apply the appropriate formatting. (e.g. `npx prettier -w . '!**/{bin,obj,dist,sonar_reports}/**'`)

However, if you run the linter in the project or save one of the affected files in VSCode (with formatOnSave enabled), the bad formatting will be reinstated.

MikeLyttle:
I responded to Manuel's comment on EmailModal.vue about an unresolved issue that was discovered involving the AdminWebClient, but I think the current changes can be reviewed and included prior to resolving that.

The latest set of commits starts by reverting the earlier commit to start fresh.  I addressed the issue involving references to the .editorconfig file in the .csproj files and I added linting and formatting to the e2e project.

I split up the remaining prettier formats by project and then file type to make them easier to review.  I also found and fixed a couple small issues that were being reported as errors by the formatter.

I ran variations of the following commands to apply the formatting, depending on whether I wanted to affect certain filetypes or all files:

```
npx prettier -w '**/*.{yaml,yml}' '!**/{bin,obj,dist,sonar_reports}/**'
npx prettier -w . '!**/{bin,obj,dist,sonar_reports}/**'
```

I made some additions to the .code-workspace file which should improve editing in VSCode:

- the search will exclude files in the *bin*, *obj*, *dist*, and *node_modules* folders by default
- the default formatter has been set to Prettier and formatOnSave has been enabled
- the default formatter for C# files has been set to the C# plugin but formatOnSave has been disabled
- I set the default solution for OmniSharp to be HealthGateway.sln just to help it out because we do have multiple solutions

The reason I disabled formatOnSave for C# files is because proper formatting requires analyzing the entire project, which may take some time, so I plan to add it using a pre-commit hook in git.  It will run the [dotnet-format](https://github.com/dotnet/format) tool to clean up files and report style errors when a commit is being made.  It should be pretty simple to set up, but I'll need to add a couple instructions to the wiki for installing dotnet-format and enabling the hook.  If we wanted, we could expand the hook to make it build the solution or npm packages as well.

tiagograf-AOT (approved):

sslaws (approved):

PR #2218, opened 2021-04-21
Fix Code Smells AB#10444

tiagograf-AOT:
# Fixes or Implements [AB#10444](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10444)

-   [ ] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Implements SonarJS and SonarAnalyzer.CSharp.
Fixes code warnings throughout the solution.

## Testing

-   [x] Unit Tests Updated
-   [ ] Not Required

### Steps to Reproduce

### UI Changes

YES | NO

### Browsers Tested

-   [ ] Chrome
-   [ ] Safari
-   [ ] Edge
-   [ ] Firefox

## Notes/Additional Comments


Apps/AdminWebClient/src/ClientApp/.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
         "@vue/typescript/recommended",
         "@vue/prettier",
         "@vue/prettier/@typescript-eslint",
+        "plugin:sonarjs/recommended",
```
FuriousLlama, line 13: not sure this is needed.

TiagoGraf89: it implements the sonarjs rules as the "sonar way" quality profile used by sonarcloud so it shows some code smells as warnings in our code.

FuriousLlama: yeah, but from the looks of it is just adding an additional linter that is competing with the other ones. The rules look similar enough

FuriousLlama: I guess my concern is that it might override other recommendations coming from vue/typescript settings, and I am not sure they would be better

FuriousLlama (approved):

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
```diff
@@ -18,8 +18,8 @@ import { ICommunicationService } from "@/services/interfaces";
 @Component({
     components: {
         BannerModal,
-        EmailModal,
-    },
+        EmailModal
```
MikeLyttle, line 21: I prefer when a trailing comma is included on the final list item. It makes additions and removals to lists less noisy in the diffs.

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
```diff
@@ -276,12 +280,12 @@ export default class CommunicationTable extends Vue {
             .then((banners: Communication[]) => {
                 this.parseComms(banners);
             })
-            .catch((err) => {
+            .catch(err => {
```
MikeLyttle, line 283: I am in favour of this style change, though it doesn't matter much to me either way.

Apps/JobScheduler/src/Jobs/EmailJob.cs
```diff
@@ -198,10 +198,7 @@ private void SendEmail(Email email)
             {
                 email.LastRetryDateTime = DateTime.UtcNow;
                 email.EmailStatusCode = email.Attempts < this.maxRetries ? EmailStatus.Pending : EmailStatus.Error;
-                if (caught is SmtpCommandException)
-                {
-                    email.SmtpStatusCode = (int)((SmtpCommandException)caught).StatusCode;
-                }
+                email.SmtpStatusCode = (int)((SmtpCommandException)caught).StatusCode;
```
MikeLyttle, line 201: Why don't we have to verify this is an SmtpCommandException before casting?

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
MikeLyttle: Also, prettier enforces the extra commas.  You should run `npx prettier --write . '!**/{bin,obj,dist,sonar_reports}/**'` from the root directory to auto-resolve this.

Apps/Medication/src/Properties/launchSettings.json
```diff
@@ -1,23 +1,23 @@
...
+      "launchUrl": "swagger",
+      "applicationUrl": "http://0.0.0.0:3003;",
+      "environmentVariables": {
+        "ODRPROXY_SERVICE_PORT": "443",
+        "ODRPROXY_SERVICE_HOST": "dev-odrproxy.pathfinder.gov.bc.ca",
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      }
+    }
+  }
+}
```
MikeLyttle, line 23: Similar to the commas at the end of lists, prettier enforces our .editorconfig settings which specify an indentation size of 4 for all files except .yml and .yaml.  This can be auto-resolved with prettier.

MikeLyttle (changes requested):
There are several style changes which should be reverted by running prettier as described in a previous comment.  Is there something that's reporting errors in the prettier-formatted code or auto-"correcting" them in this manner?

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
FuriousLlama: Does that mean that running the linter locally vs the root directory have different outcomes?

MikeLyttle: Yes, specifically for the Admin_ClientApp project.  I didn't end up figuring out why it behaved differently.

Apps/WebClient/src/ClientApp/src/components/navmenu/navHeader.vue
```diff
@@ -113,9 +113,7 @@ export default class HeaderComponent extends Vue {
     }
 
     private mounted() {
-        if (this.oidcIsAuthenticated) {
-            this.loadOidcUser();
-        }
+        this.onPropertyChanged();
```
FuriousLlama, line 116: this looks odd. Do you know it does the same?

Apps/WebClient/src/ClientApp/src/router.ts
```diff
@@ -281,47 +281,57 @@ router.beforeEach(async (to, from, next) => {
     );
     if (to.meta.routeIsOidcCallback || to.meta.stateless) {
         next();
-    } else {
-        // Make sure that the route accepts the current state
-        store.dispatch("auth/oidcCheckUser").then((isValid: boolean) => {
-            logger.info("User is valid: " + isValid);
+        return;
```
FuriousLlama, line 284: looks like sonar really likes 'early returns' vs 'elses'

Apps/WebClient/src/ClientApp/src/components/navmenu/navHeader.vue
MikeLyttle: I checked - onPropertyChanged() consists of the same code, so this should be fine.

Apps/WebClient/src/ClientApp/src/utility/fragmentStorage.ts
```diff
@@ -83,11 +82,7 @@ export class FragmentedStorage implements Storage {
     }
 
     private getFragmentSection(fragment: string): string {
-        const fragmentSection = fragment.substr(
-            this.fragmentIndicator.length,
-            fragment?.length
-        );
-        return fragmentSection;
+        return fragment.substr(this.fragmentIndicator.length, fragment?.length);
```
FuriousLlama, line 85: The downside of the lack of assignment is that debugging is a tad trickier

Apps/WebClient/src/ClientApp/src/router.ts
MikeLyttle: Yeah.  Could be related to reducing nesting, which can be nice.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
```diff
@@ -46,7 +46,7 @@ public class UserProfileController : ControllerBase
         private readonly IUserProfileService userProfileService;
         private readonly IHttpContextAccessor httpContextAccessor;
         private readonly IUserEmailService userEmailService;
-        private readonly IUserSMSService userSMSService;
+        private readonly IUserSmsService userSMSService;
```
FuriousLlama, line 49: Not sure forcing the variables names to be camel case for abbreviations is good

FuriousLlama (changes requested):
There are far too many changes here to be reviewed in one go. Considering that some of these look like changes of logic and not only of variable naming or spacing I think we should create one for each project.

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
FuriousLlama: That might be problematic. If anything, the local settings should trump the higher ones

Apps/WebClient/src/ClientApp/src/components/navmenu/navHeader.vue
TiagoGraf89: Yes that was raised as duplicated since both functions did the same thing, i renamed the method so hopefully it makes more sense.

Apps/WebClient/src/ClientApp/src/router.ts
TiagoGraf89: the smell was "code complexity too high" which i was able to mitigate by reducing the nesting! there would be other ways of refactoring it, I personally like early returns than nesting

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
TiagoGraf89: I'm not sure why this happened, looking into it, it seems to that i happened only on AdminWebClient and not in WebClient

TiagoGraf89: It was missing the prettier library in adminwebclient

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
TiagoGraf89: missing the prettier library caused this too

Apps/JobScheduler/src/Jobs/EmailJob.cs
TiagoGraf89: "Replace this type-check-and-cast sequence with an 'as' and a null check.", I missunderstood this fix!

Apps/WebClient/src/ClientApp/src/utility/fragmentStorage.ts
TiagoGraf89: I agree, we can get around the code smell by logging the value if needed for debugging

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
TiagoGraf89: We need to discuss this if its a concern

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
MikeLyttle: I tried that at the time and still had some issues.  I'm not sure if it's specific to my machine or not.  I originally posted about it [here](https://github.com/bcgov/healthgateway/pull/2144#discussion_r606164832):

> Yes. I did some digging and there seems to be something wrong with the prettier formatting in the AdminWebClient ClientApp. It removes braces from inline functions as seen above. Also, it removes the trailing comma on the last item in imports and array/object declarations and doesn't format those lists properly if you mess with the whitespace/newlines.
> 
> I wasn't able to figure out a way to resolve the underlying issue, but if you run prettier from a directory higher than ClientApp, it will apply the appropriate formatting. (e.g. `npx prettier -w . '!**/{bin,obj,dist,sonar_reports}/**'`)
> 
> However, if you run the linter in the project or save one of the affected files in VSCode (with formatOnSave enabled), the bad formatting will be reinstated.

The best lead I had was that when I run the same `npx prettier --write . '!**/{bin,obj,dist,sonar_reports}/**'` via command-line from the AdminWebClient directory, it gives an error, while it works fine in WebClient or in a parent directory.

Apps/Medication/src/Properties/launchSettings.json
TiagoGraf89: not sure why this happened here only, any thoughts? it seems like everywhere else is fine

MikeLyttle (approved):

Apps/AdminWebClient/src/ClientApp/src/components/core/CommunicationTable.vue
MikeLyttle: Per our discussion, the addition of the prettier library seems to have fixed it, so I'll have to investigate separately what's going wrong on my machine.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
MikeLyttle: The result of the discussion was that these naming changes will be rolled back and the rule should be disabled so that we can use abbreviations as we please.

Apps/Encounter/src/Delegates/IMSPVisitDelegate.cs
```diff
@@ -21,7 +21,7 @@ namespace HealthGateway.Encounter.Delegates
     using HealthGateway.Encounter.Models.ODR;
 
     /// <summary>
-    /// Interface to retrieve MSP Visit History.
+    /// Interface to retrieve Msp Visit History.
```
MikeLyttle, line 24: Could stay as MSP.

Apps/WebClient/src/Server/Controllers/UserProfileController.cs
TiagoGraf89: I reverted the Pascal naming changes and disabled the rule

MikeLyttle (dismissed):

MikeLyttle (approved):

FuriousLlama (dismissed):

FuriousLlama (approved):

MikeLyttle (approved):

PR #2223, opened 2021-04-21
Timeline unittests - Tasks/10428

FuriousLlama:
# Fixes or Implements [AB#10428](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10428)

-   [ ] Enhancement
-   [x] Bug
-   [ ] Documentation

## Description

Re-enabled timeline tests. 
Created store stubs
Added vuex types.

## Testing

-   [x] Unit Tests Updated
-   [ ] Not Required


Apps/WebClient/src/ClientApp/tsconfig.json
```diff
@@ -19,6 +20,11 @@
         },
         "lib": ["esnext", "dom", "dom.iterable", "scripthost", "es2018.promise"]
     },
-    "include": ["./src/**/*.ts", "./src/**/*.vue", "./test/**/*.ts"],
+    "include": [
+        "./src/**/*.ts",
+        "./src/**/*.vue",
+        "./test/**/*.ts",
+        "test/stubs/store/config.ts2"
```
TiagoGraf89, line 27: not sure whats this about?

FuriousLlama: not sure either, will remove it

Apps/WebClient/src/ClientApp/src/views/timeline.vue
```diff
@@ -329,7 +329,7 @@ export default class TimelineView extends Vue {
             <b-col id="timeline" class="col-12 col-lg-9 column-wrapper">
                 <div class="px-2">
                     <b-alert
-                        :show="hasNewTermsOfService"
+                        v-show="hasNewTermsOfService"
```
TiagoGraf89, line 332: what this necessary for the unit tests?

Apps/WebClient/src/ClientApp/test/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+    rules: {
+        "no-empty-function": "off",
```
TiagoGraf89, line 3: are we having nested linting configurations?, should we move this up to the root level?

TiagoGraf89 (approved):
I think everything is good other than the few questions, I'm guessing the refactoring was done to make it easier to write the unit tests

Apps/WebClient/src/ClientApp/test/.eslintrc.js
FuriousLlama: Only for the test, we want to be able to have an empty function on stubs and mocks but not on the ClientApp

Apps/WebClient/src/ClientApp/src/views/timeline.vue
FuriousLlama: Yes. Not sure why, but jest was reporting that it was visible if ```:show``` was used

MikeLyttle: This is a difference between using the show property on the BootstrapVue component and Vue's built in way to toggle "display: none" and I don't think they're equivalent.

This is what [the docs](https://bootstrap-vue.org/docs/components/alert#visible-state) say about the component:
> Use the show prop to control the visibility state of the alert. By default alerts are not shown. Set the prop show to explicitly display them.
>
> The show prop accepts boolean true or false to show and hide the alert respectively. It can also be set to a positive integer (representing seconds) to create a self dismissing alert.

It seems like this is the intended way to control visibility of the component and we should stick with that.

MikeLyttle (approved):

PR #2247, opened 2021-04-28
Webclient router unittest - Tasks/10497

FuriousLlama:
# Fixes or Implements [AB#10497](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10497)

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Increased the tests coverage for the router

## Testing

-   [x] Unit Tests Updated
-   [ ] Not Required


Apps/WebClient/src/ClientApp/src/store/modules/auth/actions.ts
```diff
@@ -125,9 +145,15 @@ export const actions: AuthActions = {
             });
     },
     signOutOidc(): void {
+        const authService: IAuthenticationService = container.get<IAuthenticationService>(
```
tiagograf-AOT, line 148: why is it necessary to move it inside the methods? it looks redundant to me!

Apps/WebClient/src/ClientApp/src/store/modules/config/actions.ts
```diff
@@ -1,39 +1,14 @@
...
-                })
-                .catch((error) => {
-                    handleError(context.commit, error);
-                    reject(error);
-                })
-                .finally(() => {
-                    console.log("Finished initialization");
-                });
-        });
+        console.log("Configuration: ", config);
```
tiagograf-AOT, line 9: lets use the logger instead of console

Apps/WebClient/src/ClientApp/src/store/modules/auth/actions.ts
FuriousLlama: That is necessary to avoid the dependency being solved at inclusion time by the import. We could make this a class so that it has members that it can use, but otherwise, the definition needs to per method

Apps/WebClient/src/ClientApp/src/store/modules/config/actions.ts
FuriousLlama: That would be good, but at this point in the code, the logger is not guaranteed to be initialized.

Apps/WebClient/src/ClientApp/test/landing.test.ts
```diff
@@ -26,24 +28,16 @@ describe("Landing view", () => {
         maxDependentAge: 12,
     };
 
-    const configGetters = {
-        webClient: (): WebClientConfiguration => {
-            return a;
-        },
+    const options = new StoreOptionsStub();
+    options.modules.config.getters.webClient = (): WebClientConfiguration => {
+        return a;
```
MikeLyttle, line 33: Can the variable `a` be renamed to be more descriptive?

MikeLyttle (approved):

Apps/WebClient/src/ClientApp/test/landing.test.ts
FuriousLlama: Sure it can! will rename it

MikeLyttle (dismissed):

MikeLyttle (approved):

tiagograf-AOT (approved):

PR #2248, opened 2021-04-29
Cleanup webclient package.json Features/10499

FuriousLlama:
# Fixes or Implements [AB#10499](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10499)

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Removes packages not used or implied by its dependencies


Apps/WebClient/src/ClientApp/package.json
```diff
@@ -35,46 +34,28 @@
         "babel-preset-env": "^1.7.0",
         "bootstrap": "~4.6.0",
         "compression-webpack-plugin": "^6.1.1",
-        "css-loader": "^5.2.4",
-        "eslint": "^7.25.0",
-        "eslint-config-prettier": "^8.3.0",
+        "eslint": "^6.8.0",
         "eslint-plugin-prettier": "^3.4.0",
-        "eslint-plugin-prettier-vue": "^2.1.1",
+        "eslint-plugin-prettier-vue": "^3.0.0",
```
MikeLyttle, line 39: I'm a bit confused about the interaction between `eslint-config-prettier`, `eslint-plugin-prettier`, and `eslint-plugin-prettier-vue`.

eslint-plugin-prettier-vue [says](https://www.npmjs.com/package/eslint-plugin-prettier-vue#eslint-config):

> **DO NOT** use `eslint-plugin-prettier` together. This plugin is based on `eslint-plugin-prettier` so you do not need it.

So, it sounds like we can get rid of `eslint-plugin-prettier`.  Are we not using `eslint-config-prettier`?

Apps/WebClient/src/ClientApp/package.json
```diff
@@ -35,46 +34,28 @@
         "babel-preset-env": "^1.7.0",
         "bootstrap": "~4.6.0",
         "compression-webpack-plugin": "^6.1.1",
-        "css-loader": "^5.2.4",
-        "eslint": "^7.25.0",
-        "eslint-config-prettier": "^8.3.0",
+        "eslint": "^6.8.0",
```
MikeLyttle, line 37: Why the change to an older version of `eslint`?

FuriousLlama: Apparently, that is the version that is compatible with our other dependencies. Not sure why it was working before, but after clearing the lock file npm was complaining it couldn't find the dependency.

Apps/WebClient/src/ClientApp/package.json
FuriousLlama: That's what I thought too, but after removing it npm complained about incompatible versions of it. I will take a second look to see if there is another dependency that is conflicting with the shared dependency between eslint-plugin-prettier and vue

tiagograf-AOT (approved):

Apps/WebClient/src/ClientApp/package.json
FuriousLlama: Spent some time removing the library and it was needed by a client configuration. Removing it didn't cause any different linting rules, so I am guessing it was redundant. 
TLDR; library is gone

MikeLyttle (approved):

tiagograf-AOT (approved):

PR #2254, opened 2021-04-30
Rename stub files AB#10444

tiagograf-AOT:
# Fixes or Implements [AB#10444](https://qslvic.visualstudio.com/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10444)

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Rename stubs.

## Testing

-   [ ] Unit Tests Updated
-   [ ] Not Required

### Steps to Reproduce

### UI Changes

YES | NO

### Browsers Tested

-   [ ] Chrome
-   [ ] Safari
-   [ ] Edge
-   [ ] Firefox

## Notes/Additional Comments


MikeLyttle (approved):

Apps/WebClient/src/ClientApp/test/stubs/store/storeOptionsStub.ts
```diff
@@ -3,22 +3,22 @@ import { injectable } from "inversify";
...
-import errorBannerStub from "./error";
-import idleStub from "./idle";
-import immunizationStub from "./immunization";
-import laboratoryStub from "./laboratory";
-import medicationStub from "./medication";
-import navbarStub from "./navbar";
-import noteStub from "./note";
-import timelineStub from "./timeline";
-import userStub from "./user";
+import authStub from "./authStub";
```
FuriousLlama, line 6: not sure this is necessary since it's already in the stub folder
```@test/stubs/store/auth```
vs
```@test/stubs/store/authstub```

Comment (FuriousLlama) [reply Apps/WebClient/src/ClientApp/test/stubs/store/storeOptionsStub.ts:6]: I think same reason why Stub/store is not stub/storeStub

**Review by FuriousLlama — CHANGES_REQUESTED** [review]
The parent folder already implies the type of object the file is.
```stub/store/auth```
vs
```stub/storeStub/storeAuthStub```

File: Apps/WebClient/src/ClientApp/src/main.ts [hunk 2]
```diff
@@ -168,7 +168,7 @@ configService.getConfiguration().then((config: ExternalConfiguration) => {
 const App = () => import(/* webpackChunkName: "entry" */ "./app.vue");
 
 function initializeVue(store: Store<RootState>) {
-    new Vue({
+    return new Vue({
```
FuriousLlama, line 171: good catch

Apps/WebClient/src/ClientApp/src/views/login.vue
```diff
@@ -82,7 +82,7 @@ export default class LoginView extends Vue {
 </script>
 
 <template>
-    <div class="container my-5" align="center">
```
FuriousLlama, line 85: This looks like is changing things outside the scope of the PR

Apps/WebClient/src/ClientApp/test/app.test.ts
```diff
@@ -1,4 +1,6 @@
 import "@/plugins/inversify.config";
+
+import StoreOptionsStub from "@test/stubs/store/storeOptionsStub";
```
FuriousLlama, line 3: the path looks redundant
```"@test/stubs/store/options``` sounds better to me

Apps/WebClient/src/ClientApp/test/stubs/store/storeOptionsStub.ts
tiagograf-AOT: the same goes for services, controllers and other objects, but we do add the suffix to make the file unique, one example is when you have multiple tabs open in vscode, by having the authStub name you automatically know what that is, otherwise you have to look at the full path to know the difference between auth vs stubs/auth

tiagograf-AOT: store was renamed to storeOptionsStub

Apps/WebClient/src/ClientApp/src/views/login.vue
tiagograf-AOT: is this really a big deal?

Apps/WebClient/src/ClientApp/test/app.test.ts
tiagograf-AOT: I agree but i rather have redundancy than not descriptive enough, in you suggestion the name of the class would be Options

Apps/WebClient/src/ClientApp/test/stubs/store/storeOptionsStub.ts
MikeLyttle: I think @tiagograf-AOT has convinced me.

Apps/WebClient/src/ClientApp/src/views/login.vue
FuriousLlama: it could be, but I am not sure since it could be changing the display of the div changing the UI. Moreover is not aligned with the stated purpose of the PR. Can you explain why you changed it?

Apps/WebClient/src/ClientApp/test/app.test.ts
FuriousLlama: The name of the class does not necessarily need to be aligned to the name of the file. However, we can start to do that if we choose to.  At this point I think renaming it to Options would be good as well

Apps/WebClient/src/ClientApp/test/stubs/store/storeOptionsStub.ts
FuriousLlama: That's where the folder name is acting as a namespace. In c# we have explicit namespaces defined so the confusion and variable collision does not happen. We can also do it here if necessary since Typescript has the concept of namespacing. 

Following the same line, we don't call the classes TestStoreOptionsStub since it is already included in the path for inclussion.

FuriousLlama: > I can revert the changes if @MikeLyttle also thinks this is how we should name our files but in that case, the solution would be to rename the classes and drop the "stub" word

The code styles that we practice are not defined by _majority vote_ but instead _consensus_. There is always room for improvement and we should aim for the highest quality of code. If there is a disagreement about which one is better we need to articulate why and how it can be improved as well as the downsides.

Apps/WebClient/src/ClientApp/src/views/login.vue
tiagograf-AOT: styling attributes like "align, width, height" are obsolete

Apps/WebClient/src/ClientApp/test/app.test.ts
tiagograf-AOT: The name of the class does need to be aligned to the name of the file if it has only one class, we already do that in the C# files

FuriousLlama (approved):

MikeLyttle (approved):

Apps/WebClient/src/ClientApp/test/stubs/store/storeOptionsStub.ts
FuriousLlama: Chatted about this online (zoom) and agreed to leave the changes as is since it aligns with SonarQube's recommendations. We also agreed to work on communicating better the intent and the questions on PRs

PR #2284, opened 2021-05-07
Create AdminTag DB Model AB#10540

tiagograf-AOT:
# Fixes or Implements AB#10540

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Create AdminTag database model and related tables.

## Testing

-   [ ] Unit Tests Updated
-   [ ] Not Required

### Steps to Reproduce

### UI Changes

YES | NO

### Browsers Tested

-   [ ] Chrome
-   [ ] Safari
-   [ ] Edge
-   [ ] Firefox

## Notes/Additional Comments


Apps/Database/src/Migrations/20210507171736_CreateAdminTag.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "AdminTag",
+                schema: "gateway",
+                columns: table => new
+                {
+                    AdminTagId = table.Column<Guid>(type: "uuid", nullable: false),
+                    Name = table.Column<string>(type: "character varying(20)", maxLength: 20, nullable: false),
```
MikeLyttle, line 33: Shouldn't there be a unique constraint on the Name column?

Apps/Database/src/Models/AdminTag.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// </summary>
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        [Required]
+        public Guid AdminTagId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the tag.
+        /// </summary>
+        [MaxLength(20)]
```
sslaws, line 38: Seems reasonable should we check with Nino?

Apps/Database/src/Models/UserFeedbackTag.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// Gets or sets the primary key.
+        /// </summary>
+        [Required]
+        public Guid AdminTagId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the related user feedback id.
+        /// </summary>
+        [Required]
+        public Guid UserFeedbackId { get; set; }
```
sslaws, line 45: You should put Navigation Properties
public virtual UserFeedback {get; set;}

and also for the AdminTag

sslaws (changes requested):

Apps/Database/src/Models/UserFeedback.cs
```diff
@@ -16,24 +16,49 @@
...
+        /// Gets or sets the related user profile id.
+        /// </summary>
         [MaxLength(52)]
         public string? UserProfileId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the related list of tags.
+        /// </summary>
+#pragma warning disable CA2227
+        public ICollection<UserFeedbackTag> Tags { get; set; } = null!;
```
FuriousLlama, line 61: We don't usually set variables to null if the parameter is not nullable. Why not creating a List<> or another collection like other fields?

Apps/Database/src/Migrations/20210507171736_CreateAdminTag.cs
tiagograf-AOT: Good point, I've added the constraint

Apps/Database/src/Models/UserFeedbackTag.cs
tiagograf-AOT: added the virtual to the properties

Apps/Database/src/Models/UserFeedback.cs
tiagograf-AOT: modified to readonly and assigned to a list

Apps/Database/src/Models/UserFeedback.cs
```diff
@@ -16,24 +16,47 @@
...
+        /// <summary>
+        /// Gets or sets the related user profile id.
+        /// </summary>
         [MaxLength(52)]
         public string? UserProfileId { get; set; }
+
+        /// <summary>
+        /// Gets the related list of tags.
+        /// </summary>
+        public ICollection<UserFeedbackTag> Tags { get; } = new List<UserFeedbackTag>();
```
FuriousLlama, line 60: Not sure it will work if it is not modifiable since there is no Ctor to set it initially

tiagograf-AOT: its being initialized inline

FuriousLlama: Sure, but initialized as empty.  If that works with the ef then no prob

FuriousLlama (approved):

Apps/Database/src/Models/UserFeedbackTag.cs
```diff
@@ -0,0 +1,57 @@
...
+        [Required]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        [Key]
+        public Guid UserFeedbackTagId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the primary key.
+        /// </summary>
+        [Required]
+        public Guid AdminTagId { get; set; }
```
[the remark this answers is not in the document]
FuriousLlama: You are right, it is implied. However, I think is better if it's explicit.
https://docs.microsoft.com/en-us/ef/core/modeling/relationships?tabs=fluent-api%2Cfluent-api-simple-key%2Csimple-key#no-foreign-key-property

PR #2299, opened 2021-05-11
Admin tag user interface and FE services - Tasks/10541

FuriousLlama:
# Fixes or Implements [AB#10541](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10541)

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Adds user interface and front end services as well as a draft backend controller

image


Apps/AdminWebClient/src/ClientApp/src/models/userFeedback.ts
```diff
@@ -1,23 +1,32 @@
...
 
-    // Gets or sets the email if known for this feedback.
+    // Email if known for this feedback.
     email: string;
+
+    // The feedback tags
+    tags: AdminTag[];
+}
+
+export interface AdminTag {
```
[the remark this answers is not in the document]
FuriousLlama: We don't have that requirement on the front end. Multiple interfaces and classes exist on the same file

Apps/AdminWebClient/src/Server/Models/UserFeedbackView.cs
```diff
@@ -17,13 +17,53 @@ namespace HealthGateway.Admin.Models
...
+        [JsonConstructor]
+        public UserFeedbackView(
+            Guid id,
+            string? comment,
+            bool isSatisfied,
+            bool isReviewed,
+            DateTime createdDateTime,
+            uint version,
+            string email,
+            IList<AdminTagView> tags)
```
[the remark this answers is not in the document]
FuriousLlama: Sure, whichever works

Apps/AdminWebClient/src/ClientApp/src/views/Feedback.vue
```diff
@@ -168,5 +220,82 @@ export default class FeedbackView extends Vue {
...
+                        console.log(err);
+                    })
+                    .finally(() => {
+                        this.isLoadingTag = false;
+                    });
+            }
+        });
+    }
+
+    private filter(item: AdminTag, queryText: string): boolean {
```
tiagograf-AOT, line 297: Its nice of you to want to implement everything in one shot but we created another user story to handle this, as a matter of fact I don't know if we want to use a partial search (textbox) or would that be a exact match (combobox) when filtering tags, also notice that we do have case sensitive tags but your filter doesn't take that in consideration

Apps/AdminWebClient/src/ClientApp/src/services/restUserFeedbackService.ts
```diff
@@ -58,4 +62,96 @@ export class RestUserFeedbackService implements IUserFeedbackService {
...
     }
+
+    public getAllTags(): Promise<AdminTag[]> {
+        return new Promise((resolve, reject) => {
+            const headers: Dictionary<string> = {};
+            headers[this.contentTypeHeader] = this.contentType;
+            this.http
+                .get<AdminTag[]>(`${this.BASE_URI}/Tag`, headers)
+                .then((requestResult) => {
+                    if (requestResult) {
```
tiagograf-AOT, line 73: most likely this condition should be almost always true, we do have a status of the request that must be checked

Apps/AdminWebClient/src/ClientApp/src/views/Feedback.vue
FuriousLlama: Part of the nice-ness of using v-combobox is that it allows search and filter pretty much out of the box. As it is it pretty much does what we need to do to complete that story.

Regarding using a textbox vs combobox that is an implementation detail that was not discussed during design.

We can check with Nino about the case-sensitive search, if you read the method it accounts for the case.

Apps/AdminWebClient/src/ClientApp/src/services/restUserFeedbackService.ts
FuriousLlama: fair point, I will remove it in the coming review that ties the front end and backend

Apps/AdminWebClient/src/ClientApp/src/models/userFeedback.ts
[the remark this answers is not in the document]
FuriousLlama: Not sure I am following your comment. The user feedback on the admin side is directly related to the admin tag. If in the future the tag is used elsewhere we can mode it.

tiagograf-AOT (approved):

PR #2474, opened 2021-06-21
In app communication UI changes

FuriousLlama:
# Fixes or Implements [AB#10709](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10709)

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Added communication changes to support in-app

image


Apps/WebClient/src/ClientApp/src/models/communication.ts
```diff
@@ -8,9 +8,46 @@ export default interface Communication {
...
+
+    // Draft communication.
+    Draft,
+}
+
+export enum CommunicationType {
+    // Banner communication.
+    Banner,
+
+    // Email communication.
```
sslaws, line 48: The front-end at least currently should never see an email communication type.  The code will throw an exception if this occurs on the server so I'm not sure if we should include the Email type.  I was thinking server side that a front end enum could be created but we didn't have a constants package and it didn't seem worth the hassle.

sslaws (approved):

PR #2544, opened 2021-07-06
Format dates for current locale on admin support page AB#10758

MikeLyttle:
# Implements [AB#10758](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/10758)

-   [x] Enhancement
-   [ ] Bug
-   [ ] Documentation

## Description

Changes date display to local time instead of UTC on the Support page in the Admin application.

## Testing

-   [ ] Unit Tests Updated
-   [x] Not Required

### Steps to Reproduce

### UI Changes

NO

### Browsers Tested

-   [x] Chrome
-   [ ] Safari
-   [ ] Edge
-   [ ] Firefox

## Notes/Additional Comments

### DIFF
--- a/Apps/AdminWebClient/src/ClientApp/src/models/messageVerification.ts
+++ b/Apps/AdminWebClient/src/ClientApp/src/models/messageVerification.ts
@@ -13,10 +13,10 @@ export default interface MessageVerification {
     verificationType: VerificationType;
     smsNumber: string | null;
     smsValidationCode: string;
-    expireDate: string;
+    expireDate: Date;
     verificationAttempts: number;
     deleted: boolean;
-    updatedDateTime: string;
+    updatedDateTime: Date;
 }
 
 export interface Email {
@@ -25,7 +25,7 @@ export interface Email {
     to: string;
     subject: string;
     body: string;
-    sentDateTime: string;
+    sentDateTime: Date;
     lastRetryDateTime?: string;
     attempts: number;
     smtpStatusCode: number;
--- a/Apps/AdminWebClient/src/ClientApp/src/views/Support.vue
+++ b/Apps/AdminWebClient/src/ClientApp/src/views/Support.vue
@@ -1,5 +1,4 @@
 <script lang="ts">
-import moment from "moment";
 import { Component, Vue } from "vue-property-decorator";
 
 import BannerFeedbackComponent from "@/components/core/BannerFeedback.vue";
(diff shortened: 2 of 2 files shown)


tiagograf-AOT (approved):

Apps/AdminWebClient/src/ClientApp/src/models/messageVerification.ts
```diff
@@ -13,10 +13,10 @@ export default interface MessageVerification {
     verificationType: VerificationType;
     smsNumber: string | null;
     smsValidationCode: string;
-    expireDate: string;
+    expireDate: Date;
```
FuriousLlama, line 16: This is not a good idea since it presumes the return model is date where instead is an ISOstring that represents a date

Apps/AdminWebClient/src/ClientApp/src/views/Support.vue
```diff
@@ -121,8 +120,11 @@ export default class SupportView extends Vue {
         this.supportService = container.get(SERVICE_IDENTIFIER.SupportService);
     }
 
-    private formatDate(date: string): string {
-        return moment(date).format("l LT");
+    private formatDate(date: Date): string {
```
FuriousLlama, line 123: We need to move away from using Date since its capabilities are limited when compared to moment or luxon.

FuriousLlama (changes requested):
Instead of using Date as a plain data type, we should use a more feature-rich library. Since we abstract that on WebClient already we should do that here too. IMO Copyng the DateWrapper class here would suffice and would allow us to re-use that code

MikeLyttle:
> Instead of using Date as a plain data type, we should use a more feature-rich library. Since we abstract that on WebClient already we should do that here too. IMO Copyng the DateWrapper class here would suffice and would allow us to re-use that code

I agree with this and your other comments, but shouldn't we merge this & create a separate story that addresses all of the dates in AdminWebClient?  This PR is just following the established pattern.

FuriousLlama:
> > Instead of using Date as a plain data type, we should use a more feature-rich library. Since we abstract that on WebClient already we should do that here too. IMO Copyng the DateWrapper class here would suffice and would allow us to re-use that code
> 
> I agree with this and your other comments, but shouldn't we merge this & create a separate story that addresses all of the dates in AdminWebClient? This PR is just following the established pattern.

Other sections of the admin client use moment to do date calculations and formating (although not fully). We already have a story about replacing moment with luxon.  I think it would be better to bring the date utility from webclient and use it in this case since that will both address only this problem and pave the way for tackling the other story that addresses all other dates.

PR #3725, opened 2022-05-27
Add confirm terms of service page to login flow for AB#13101.

BrianMaki:
# Fixes or Implements [AB#13101](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13101)

## Description

Add confirm terms of service page to login flow.

- Updated router.ts so that terms of service page is shown when user is required to accept
- Updated app.vue so that side menu is not shown when user has not accepted updated terms of service
- Updated server side UserProfileModel.cs so that 'AcceptedTermsOfService' is calculated when UI model is being created so that both UpdateAcceptedTerms and GetUserProfile from UserProfileService calls return the calculated value.

Note:

- While logged in and registered, clicking on FAQ, Release Notes, Terms of Service and Contact Us in footer with require acknowledgment of terms of service if terms of service has been updated.  If not logged in, no acknowledgement of terms of service is required.
- If account is pending deletion, terms of service confirmation is not required.

Screen Shot 2022-05-27 at 9 53 48 AM

## Testing

-   [ ] Unit Tests Updated
-   [x] Not Required

### UI Changes

YES

### Browsers Tested

-   [x] Chrome
-   [ ] Safari
-   [ ] Edge
-   [ ] Firefox

## Notes/Additional Comments

## Items to Review:

-   Fulfills Work Item Requirements

    -   The changes meet/implement story's acceptance criteria. Fixes identified problems if bug.
    -   Changes not directly related to the task requirements need to be documented on the PR.

-   Compilation / Tests

    -   The changes do not break compilation and/or unit or functional tests; unless other item addresses them specifically.

-   Logic Problems / Functional Behaviour

    -   The changes work as intended and do not introduce problems.

-   Performance

    -   The changes do not introduce obvious performance issues.

-   Documented Standards

    -   The changes adhere to the team's documented [Coding Standards](https://github.com/bcgov/healthgateway/wiki/standards).

-   Readability / Maintainability
    -   The changes can be easily understood/read and allow for future enhancements without major refactoring. Readability is preferred over "clever code".
    -   Reasoning for changes needs to be clearly articulated. Disagreements should be arbitrated by a third developer.

### DIFF
--- a/Apps/GatewayApi/src/Models/UserProfileModel.cs
+++ b/Apps/GatewayApi/src/Models/UserProfileModel.cs
@@ -123,6 +123,7 @@ namespace HealthGateway.GatewayApi.Models
                 IsSMSNumberVerified = !string.IsNullOrEmpty(model.SMSNumber),
                 LastLoginDateTime = model.LastLoginDateTime,
                 ClosedDateTime = model.ClosedDateTime,
+                AcceptedTermsOfService = model.TermsOfServiceId != Guid.Empty,
             };
         }
     }
--- a/Apps/GatewayApi/src/Services/UserProfileService.cs
+++ b/Apps/GatewayApi/src/Services/UserProfileService.cs
@@ -145,7 +145,6 @@ namespace HealthGateway.GatewayApi.Services
             }
 
             userProfile.HasTermsOfServiceUpdated = retVal.Payload?.TermsOfServiceId != termsOfServiceResult.ResourcePayload?.Id;
-            userProfile.AcceptedTermsOfService = retVal.Payload?.TermsOfServiceId != Guid.Empty;
 
             if (!userProfile.IsEmailVerified)
             {
--- a/Apps/WebClient/src/ClientApp/src/router.ts
+++ b/Apps/WebClient/src/ClientApp/src/router.ts
@@ -73,9 +73,9 @@ const HealthInsightsView = () =>
     import(
         /* webpackChunkName: "healthInsights" */ "@/views/HealthInsightsView.vue"
     );
-const ConfirmTermsOfServiceView = () =>
+const AcceptTermsOfServiceView = () =>
     import(
-        /* webpackChunkName: "confirmTermsOfService" */ "@/views/ConfirmTermsOfService.vue"
+        /* webpackChunkName: "acceptTermsOfService" */ "@/views/AcceptTermsOfService.vue"
     );
 const ReportsView = () =>
     import(/* webpackChunkName: "reports" */ "@/views/ReportsView.vue");
@@ -99,7 +99,7 @@ export enum UserState {
     pendingDeletion = "pendingDeletion",
     invalidLogin = "invalidLogin",
     offline = "offline",
-    confirmTermsOfService = "confirmTermsOfService",
+    acceptTermsOfService = "acceptTermsOfService",
 }
 
 function calculateUserState() {
@@ -126,7 +126,7 @@ function calculateUserState() {
     } else if (!userIsActive) {
         return UserState.pendingDeletion;
     } else if (hasTermsOfServiceUpdated) {
-        return UserState.confirmTermsOfService;
+        return UserState.acceptTermsOfService;
     } else {
         return UserState.registered;
     }
@@ -158,7 +158,7 @@ const REGISTRATION_INFO_PATH = "/registrationInfo";
 const ROOT_PATH = "/";
 const TIMELINE_PATH = "/timeline";
 const UNAUTHORIZED_PATH = "/unauthorized";
-const CONFIRM_TERMS_OF_SERVICE_PATH = "/confirmTermsOfService";
+const ACCEPT_TERMS_OF_SERVICE_PATH = "/acceptTermsOfService";
 
 const routes = [
     {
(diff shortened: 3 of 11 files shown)


Apps/GatewayApi/src/Models/UserProfileModel.cs
```diff
@@ -123,6 +123,7 @@ public static UserProfileModel CreateFromDbModel(UserProfile model)
                 IsSMSNumberVerified = !string.IsNullOrEmpty(model.SMSNumber),
                 LastLoginDateTime = model.LastLoginDateTime,
                 ClosedDateTime = model.ClosedDateTime,
+                AcceptedTermsOfService = model.TermsOfServiceId != Guid.Empty,
```
sslaws, line 126: This initialization doesn't seem write to me as having a TermsOfServiceId doesn't imply they have accepted the terms of service.  I think any logic like this should be in service code and simple a default boolean initialization of false should be sufficient.  Do we need this for some reason?

MikeLyttle, line 126: This value needs to reflect whether the model.TermsOfServiceId matches the most recent Terms of Service.

MikeLyttle (changes requested):

Apps/GatewayApi/src/Services/UserProfileService.cs
```diff
@@ -145,7 +145,6 @@ public RequestResult<UserProfileModel> GetUserProfile(string hdid, DateTime jwtA
             }
 
             userProfile.HasTermsOfServiceUpdated = retVal.Payload?.TermsOfServiceId != termsOfServiceResult.ResourcePayload?.Id;
-            userProfile.AcceptedTermsOfService = retVal.Payload?.TermsOfServiceId != Guid.Empty;
```
sslaws, line 148: This goes along with the code above but this is a special scenario where if the Guid is empty it has to be not accepted and otherwise the code flow would indicate a true value.

sslaws (changes requested):

Apps/GatewayApi/src/Models/UserProfileModel.cs
MikeLyttle: Scratch that - this is captured in the HasTermsOfServiceUpdated property.

Apps/WebClient/src/ClientApp/src/router.ts
```diff
@@ -180,7 +190,19 @@ const routes = [
...
+    {
+        path: CONFIRM_TERMS_OF_SERVICE_PATH,
+        component: ConfirmTermsOfServiceView,
+        meta: {
+            validStates: [
+                UserState.confirmTermsOfService,
+                UserState.registered,
+            ],
+        },
     },
```
MikeLyttle, line 206: Shouldn't the only valid state be UserState.confirmTermsOfService?

mk08sh (approved):

Apps/WebClient/src/ClientApp/src/views/ConfirmTermsOfService.vue
```diff
@@ -0,0 +1,218 @@
...
+        event.preventDefault();
+    }
+
+    private redirect(): void {
+        this.logger.debug("Redirecting to /profile");
+
+        this.checkRegistration().then(() => {
+            this.$router.push({ path: "/profile" });
+        });
+    }
```
MikeLyttle, line 150: Nit: using `checkRegistration` is performing an extra network call to retrieve the user profile when we already have the updated profile from calling `updateAcceptedTerms`.

However, the actions behave differently because the code that sets default values for the user's tutorial preferences is set only in the `checkRegistration` action.

The usage of `checkRegistration` could be removed from here if the preference-setting code from `checkRegistration` was moved into the `setProfileUserData` mutation.

Apps/WebClient/src/ClientApp/src/router.ts
BrianMaki: Agreed.

Apps/GatewayApi/src/Models/UserProfileModel.cs
MikeLyttle: Having a non-empty `TermsOfServiceId` does imply they have accepted the terms of service.  That is how we are using that database field.

Setting the value here instead of in the service ensures that the `AcceptedTermsOfService` value is set properly for all actions that return a `UserProfileModel` instead of just the `GetUserProfile` action.

Apps/WebClient/src/ClientApp/src/views/ConfirmTermsOfService.vue
BrianMaki: Agreed. Not necessary anymore because of backend server side change.

Apps/GatewayApi/src/Models/UserProfileModel.cs
BrianMaki: This code was moved over to the UserProfileModel because both 'UpdateAcceptedTerms' and 'GetUserProfile' needed it.  

Scenario:

1. User logs in and is required to accept terms of service
2. User accepts term of service via 'UpdateAcceptedTerms' - but UpdateAcceptedTerms call does not update boolean AcceptedTermsOfService.  It is defaulted to false.
3. At this point user cannot navigate to the home page because the route requires user to be registered i.e., accepted terms of service.

Question:

If a new Terms of Service is added to the Legal Agreement table, can we not null out the TermsOfServiceId column in User Profile.  Also right now, it looks UserProfile.TermsOfServiceId is a NOT NULL column.

Apps/GatewayApi/src/Services/UserProfileService.cs
BrianMaki: See my reply to line 126 change.

MikeLyttle (approved):

sslaws (approved):

PR #3798, opened 2022-06-15
Implement add vaccine card quick link button for AB#13359

rayking29:
# Implements AB#13359

## Description

BC Vaccine Card button can be added back in "add quick link" selector type "BC Vaccine Card" 

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

Screen Shot 2022-06-15 at 9 32 01 AM
Screen Shot 2022-06-15 at 9 33 14 AM
Screen Shot 2022-06-15 at 9 33 39 AM

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
@@ -4,9 +4,11 @@ import { Component } from "vue-property-decorator";
 import { Action, Getter } from "vuex-class";
 
 import { EntryType, entryTypeMap } from "@/constants/entryType";
+import UserPreferenceType from "@/constants/userPreferenceType";
 import type { WebClientConfiguration } from "@/models/configData";
 import { QuickLink } from "@/models/quickLink";
 import User from "@/models/user";
+import { UserPreference } from "@/models/userPreference";
 import container from "@/plugins/container";
 import { SERVICE_IDENTIFIER } from "@/plugins/inversify";
 import { ILogger } from "@/services/interfaces";
@@ -29,6 +31,11 @@ export default class AddQuickLinkComponent extends Vue {
     updateQuickLinks!: (params: {
         hdid: string;
         quickLinks: QuickLink[];
+    }) => Promise<void>;
+
+    @Action("updateUserPreference", { namespace: "user" })
+    updateUserPreference!: (params: {
+        userPreference: UserPreference;
     }) => Promise<void>;
 
     @Getter("quickLinks", { namespace: "user" }) quickLinks!:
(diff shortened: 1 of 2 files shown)


sslaws (approved):

mk08sh (commented):
Looks good!

Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
```diff
@@ -165,8 +190,8 @@ export default class AddQuickLinkComponent extends Vue {
     >
         <form data-testid="quick-link-modal-text">
             <b-row
-                v-for="(quickLink, index) in enabledQuickLinkFilter"
-                :key="index"
```
BrianMaki, line 169: Why was the :key value changed?

BrianMaki (changes requested):
I have a question regarding line 169.

Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
rayking29: When Mike and I were going over the strategy for my changes he noticed that and suggested the change to use quicklink.module instead so that it is a unique value to that item only. You may want to check it with him.

Apps/WebClient/src/ClientApp/src/views/HomeView.vue
```diff
@@ -209,7 +209,7 @@ export default class HomeView extends Vue {
                             existingLink.filter.modules.length === 1 &&
                             existingLink.filter.modules[0] === details.type
                     ) === undefined
-            ).length === 0
+            ).length === 0 && this.showVaccineCardButton
```
MikeLyttle, line 212: This condition won't work correctly if the "VaccinationStatus" module is disabled.

I'd suggest creating getters for the separate conditions represented by `showVaccineCardButton`: `vaccinationStatusModuleEnabled` and `preferenceVaccineCardHidden`.

Then this condition could be adjusted to `noQuickLinksToAdd && (!vaccinationStatusModuleEnabled || preferenceVaccineCardHidden)`.

rayking29: This actually was working for the "VaccinationStatus" module being disabled because it is handled in the showVaccineCardButton method. The other confusion is that this method isAddQuickLinkButtonDisabled is about disabling the button so the suggested logic above is actually backward.
However, I realize this wasn't very clear, so I changed it as you suggested in the new task (13245) to make it better creating the getters for the conditions and using them in both showVaccineStatusButton and isAddQuickLinkButttonDisabled methods.

PR #3801, opened 2022-06-16
Fix behaviour when "VaccinationStatus" module is disabled for AB#13382

rayking29:
# Implements [AB#13382](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13382)

## Description

There is some tricky logic where the "VaccinationStatus" module state was involved for whether the add button should be enabled or not. Clarified with getters and fixed the logic.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
@@ -67,29 +67,15 @@ export default class AddQuickLinkComponent extends Vue {
     }
 
     private get isAddButtonEnabled(): boolean {
-        return (
-            (this.enabledQuickLinkFilter.length > 0 &&
-                this.selectedQuickLinks.length > 0) ||
-            (this.selectedQuickLinks.length > 0 &&
-                (!this.vaccinationStatusModuleEnabled ||
-                    this.preferenceVaccineCardHidden))
-        );
-    }
-
-    private get vaccinationStatusModuleEnabled(): boolean {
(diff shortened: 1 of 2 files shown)


Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
```diff
@@ -70,16 +70,29 @@ export default class AddQuickLinkComponent extends Vue {
         return (
             (this.enabledQuickLinkFilter.length > 0 &&
                 this.selectedQuickLinks.length > 0) ||
-            (this.showVaccineCard && this.selectedQuickLinks.length > 0)
+            (this.selectedQuickLinks.length > 0 &&
+                (!this.vaccinationStatusModuleEnabled ||
+                    this.preferenceVaccineCardHidden))
         );
     }
```
MikeLyttle, line 77: I think this getter can be simplified to just `this.selectedQuickLinks.length > 0`.

MikeLyttle (changes requested):
Base branch should be swapped to `features/13245`.

Apps/WebClient/src/ClientApp/src/components/modal/AddQuickLinkComponent.vue
rayking29: Good call, I thought that method was way more complicated than it needed to be.  I have simplified it as suggested and also simplified the other code since I don't need the getters any more as they are only used in one place.

BrianMaki (approved):

PR #3802, opened 2022-06-16
Fixed recurring user count query for AB#13371.

BrianMaki:
# Fixes [AB#13371](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13371)

## Description

Recurring user count was returning back the correct count.  Date time in where clause was not being truncated. As a result, the group by clause was not accurate and was returning back an invalid count.

Previously 100 unique days would return back an invalid count. That has been addressed.  100 unique days should return 0 as there is no user that has logged in 100 recurring days in the database. Note: date range is one month.

Screen Shot 2022-06-15 at 8 28 16 PM

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:

### DIFF
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -244,7 +244,7 @@ namespace HealthGateway.Database.Delegates
                 .Select(x => new { x.HdId, x.LastLoginDateTime })
                 .Concat(
                     this.dbContext.UserProfileHistory.Select(x => new { x.HdId, x.LastLoginDateTime }))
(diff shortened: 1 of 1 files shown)


sslaws (approved):

BrianMaki:
> Shouldn't this call also being using the same offset that the other calls are using to adjust to the date range we expect? Or is that accounted for elsewhere?

This fixes truncating of the date on the database, which is in UTC.  This is not affecting the date that is being passed in from the UI.  Those values have been addressed in the service.

Note: this is addressing the Posgtrres/EF Core bug where the  postgres truncate functionality is not working with the current Postgres DB Version and EF Core.

PR #3835, opened 2022-06-23
Research AutoMapper and implement on GatewayApi for AB#13374

sslaws:
# Fixes or Implements [AB#13374](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13374)

## Description
Implements AutoMapper on the GatewayApi project to determine if we want to adopt across Health Gateway.
AutoMapper was added to Common so all server side projects can make use of it.

Key Findings:

- As our model objects used to contain To, From, and CreateList, we still need wrappers for those as the Services and Tests often use that code.

- Mapping is straight forward although Notes and Comments had a few additional hurdles.  In these types of cases the mappers are more complex than the straight copy we had.

- Models are cleaner but our newer approach resulted in the same.

## Testing

- [X] Unit Tests Updated
- [ ] Not Required

## Items to Review:


Apps/GatewayApi/src/MapUtils/CommentMapUtils.cs
```diff
@@ -0,0 +1,87 @@
...
+                    opts.AfterMap((src, dest) =>
+                    {
+                        foreach ((Comment comment, UserComment userComment) in src.Zip(dest))
+                        {
+                            userComment.Text = !string.IsNullOrEmpty(comment.Text) ? cryptoDelegate.Decrypt(key, comment.Text) : string.Empty;
+                        }
+                    }));
+
+            return userComments;
+        }
```
MikeLyttle, line 85: I think we should do away with methods that convert enumerables of one type to enumerables of another type.

The same functionality can be achieved using a Select() query in LINQ, e.g.

```
dbComments.Payload.Select(c => CommentMapUtils.CreateFromDbModel(c, this.cryptoDelegate, key, this.autoMapper))
```

sslaws: That is Cleaner and Clearer :)

Apps/GatewayApi/src/MapProfiles/DependentInformationProfile.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public class DependentInformationProfile : Profile
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DependentInformationProfile"/> class.
+        /// </summary>
+        public DependentInformationProfile()
+        {
+            this.CreateMap<PatientModel, DependentInformation>()
+                .ForMember(dest => dest.DateOfBirth, opt => opt.MapFrom(src => src.Birthdate))
```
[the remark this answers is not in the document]
sslaws: The old mapping code had a similar logic so I carried it forward.  In general, I agree that we should align the objects as much as possible but I didn't feel a refactor was in order and if it was, I would have likely carried it forward to the UI as well.

Apps/GatewayApi/src/Models/UserNote.cs
```diff
@@ -77,70 +77,5 @@ public class UserNote
...
-                Id = model.Id,
-                HdId = model.HdId,
-                JournalDate = model.JournalDate,
-                Version = model.Version,
-                CreatedDateTime = model.CreatedDateTime,
-                CreatedBy = model.CreatedBy,
-                UpdatedDateTime = model.UpdatedDateTime,
-                UpdatedBy = model.UpdatedBy,
-                Title = model.Title != null ? cryptoDelegate.Decrypt(key, model.Title) : string.Empty,
-                Text = model.Text != null ? cryptoDelegate.Decrypt(key, model.Text) : string.Empty,
```
MikeLyttle, line 101: It looks like NoteMapUtils only handles encrypting and decrypting the Text property and not the Title property.

Apps/GatewayApi/src/Models/UserProfileModel.cs
```diff
@@ -105,28 +105,5 @@ public UserProfileModel(IDictionary<string, UserPreferenceModel> preferences, IL
...
-                TermsOfServiceId = model.TermsOfServiceId,
-                Email = model.Email,
-                IsEmailVerified = !string.IsNullOrEmpty(model.Email),
-                SMSNumber = model.SMSNumber,
-                IsSMSNumberVerified = !string.IsNullOrEmpty(model.SMSNumber),
-                LastLoginDateTime = model.LastLoginDateTime,
-                ClosedDateTime = model.ClosedDateTime,
-                AcceptedTermsOfService = model.TermsOfServiceId != Guid.Empty,
-                HasTermsOfServiceUpdated = model.TermsOfServiceId != latestTermsOfServiceId,
-            };
```
MikeLyttle, line 129: It looks like UserProfileMapUtils only handles the calculation of the HasTermsOfServiceUpdated property and not AcceptedTermsOfService, IsEmailVerified, or IsSMSNumberVerified.

MikeLyttle (changes requested):

Apps/GatewayApi/src/Models/UserProfileModel.cs
sslaws: The other three are handled directly by the UserProfileProfile.

MikeLyttle (approved):

BrianMaki (approved):

PR #3866, opened 2022-06-30
Admin Dashboard Loading AB#13364

rayking29:
# Implements [AB#13364](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13364)

## Description

add loading indicator to 
- refresh button - loading
- tiles - skeleton loader
- rating summary - skeleton loader
- registration/login table - loading

don't show 0's until queries are complete 

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

Screen Shot 2022-06-29 at 8 59 06 AM

## Items to Review:


ninosamson:
nice skeletons.

mk08sh (commented):
👍

Apps/Admin/Client/Pages/DashboardPage.razor.cs
```diff
@@ -170,7 +178,7 @@ private int UniqueDays
 
             List<DailyDataRow> results = new();
 
-            if (this.RegisteredUsersResult?.Result != null)
+            if (this.RegisteredUsersResult?.Result != null && this.DashboardState.Value.RegisteredUsers.Loaded)
```
MikeLyttle, line 181: Nit: these two conditions are checking the same thing; the `Loaded` getter just returns whether or not `Result` is null.

Apps/Admin/Tests/Functional/cypress/integration/ui/dashboard.cy.js
```diff
@@ -103,6 +105,32 @@ describe("Dashboard", () => {
                 cy.get("[data-testid=daily-data-dependents]").contains("3");
             });
 
+        cy.logout();
+
         cy.log("Dashboard test finished.");
     });
+
+    it("Verify dashboard skeletons.", () => {
```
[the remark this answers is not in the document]
rayking29: I moved the skeleton test into the existing test as suggested.

Apps/Admin/Client/Pages/DashboardPage.razor
```diff
@@ -16,58 +16,100 @@
...
-        </MudPaper>
-    </MudItem>
-    <MudItem xs="6" lg="6">
-        <MudPaper Class="my-2 pa-4" Outlined="true">
-            <MudText Align="Align.Center" Typo="Typo.subtitle1"><strong>Dependents</strong></MudText>
-            <MudText Align="Align.Center" Typo="Typo.subtitle1" data-testid="total-dependents"><strong>@TotalDependents</strong></MudText>
-        </MudPaper>
-    </MudItem>
-</MudGrid>
+@if (RegisteredUsersLoading || LoggedInUsersLoading || DependentsLoading || RecurringUsersLoading)
```
MikeLyttle, line 24: Skeletons should only be displayed for calls that are outstanding. As each call completes, its skeleton should be replaced with the populated component/data.

MikeLyttle (changes requested):

Apps/Admin/Client/Pages/DashboardPage.razor.cs
rayking29: Okay I didn't realize that. I thought it didn't work the original way (this.DependentsResult?.Result != null) but it was actually a bug in the reducer I had to fix. I got rid of the duplication.

Apps/Admin/Client/Pages/DashboardPage.razor
rayking29: Is that how they actually want it to work?  I thought it might be a little busy if there are sections changing all over the place rather than all at once.

MikeLyttle: I thought that was how the old admin site handled it, but it looks like it does it for recurring users and ratings but not registered users and dependents.  I would ask @ninosamson if the current implementation is suitable.

ninosamson: Are we able to show them as they complete @rayking29 ? Logged in call just takes so long.

rayking29: Yes I can break it down for those 4 different loads.  The one I'm not sure about is the Rating Summary?  It is a separate component so I wasn't sure how to handle that.

rayking29: Also the date range selector and unique days parts, when do they show?

MikeLyttle: The inputs appear right away on the old admin site.

rayking29: How about if the skeletons appearance doesn't include the logged in users, since that is only shown in the Daily Data table?  So the table and the refresh button will show as Loading until all the loads are done but the other tiles will show up without waiting for the logged in users.

MikeLyttle: I'm still not sure that the inputs should become skeletons while data is loading, but otherwise the changes look good.

MikeLyttle (approved):

PR #3898, opened 2022-07-08
Convert date to local time for display for bug AB#13440

rayking29:
# Fixes AB#13440

## Description

Added conversion to local date time for display of date in PST/PDT rather than UTC time.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

Screen Shot 2022-07-08 at 9 41 42 AM

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Pages/Support.razor.cs
+++ b/Apps/Admin/Client/Pages/Support.razor.cs
@@ -126,7 +126,7 @@ namespace HealthGateway.Admin.Client.Pages
 
                 };
                 this.Verified = model.Validated ? "true" : "false";
-                this.VerificationDate = DateFormatter.ToShortDateAndTime(model.UpdatedDateTime);
+                this.VerificationDate = DateFormatter.ToShortDateAndTime(model.UpdatedDateTime.ToLocalTime());
                 this.VerificationCode = model.VerificationType switch
                 {
                     MessagingVerificationType.Email => "-",


BrianMaki (changes requested):
Can you provide a screenshot of the database table as well in addition to the UI screenshot you have already attached to confirm the change?

rayking29:
Updated to include the database table screen shot as per review suggestion.

BrianMaki (dismissed):

Apps/Admin/Client/Pages/Support.razor.cs
```diff
@@ -126,7 +126,7 @@ public MessagingVerificationRow(MessagingVerificationModel model)
 
                 };
                 this.Verified = model.Validated ? "true" : "false";
-                this.VerificationDate = DateFormatter.ToShortDateAndTime(model.UpdatedDateTime);
+                this.VerificationDate = DateFormatter.ToShortDateAndTime(model.UpdatedDateTime.ToLocalTime());
```
sslaws, line 129: We can't use ToLocalTime as the servers are on UTC.  You have to have specific conversion logic or have the front-end to the conversion (recommended).

sslaws (changes requested):

Apps/Admin/Client/Pages/Support.razor.cs
BrianMaki: Yes.

BrianMaki (changes requested):

Apps/Admin/Client/Pages/Support.razor.cs
rayking29: I moved the toLocalTime and date formatter code to the front end Support.razor code instead.

BrianMaki: Did you try doing a custom deploy to the mock environment for confirmation?

rayking29: No I haven't.
I have a question though.  Is it true that the .razor.cs code runs on the server machine and the .razor code runs on the client?  If that is true, as Stephen is suggesting, then the DashboardPage.razor.cs code: "private int TimeOffset { get; set; } = (int)TimeZoneInfo.Local.GetUtcOffset(DateTime.Now).TotalMinutes;" would not give back the correct offset if the server is on UTC time.

sslaws: As discussed, this is client code so your original code was fine.  Sorry

sslaws (approved):

BrianMaki (approved):

PR #3972, opened 2022-08-05
Features/13448

rayking29:
## Description

- update login page 
- login button takes user to sso common login page with PHSA/IDIR selection
- adding query parameter of idir or phsa to URL will direct user to appropriate IDP 
- copy: 
 "This application is only for authorized Health Gateway support team members. If you do not have access after login, 
 please contact your administrator." 

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

Screen Shot 2022-08-05 at 1 43 03 PM

## Items to Review:


mk08sh:
Looks good 👍

Apps/Admin/Client/Program.cs
```diff
@@ -63,10 +66,23 @@ public static async Task Main(string[] args)
             // Enable Mud Blazor component services
             builder.Services.AddMudServices(config => config.SnackbarConfiguration.PositionClass = Defaults.Classes.Position.BottomRight);
 
+            builder.Services.AddScoped(_ => new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
```
sslaws, line 69: Do we need this?  It doesn't look used and we use Refit for the APIs

Apps/Admin/Client/Program.cs
```diff
@@ -82,7 +98,8 @@ public static async Task Main(string[] args)
 
             builder.Services.AddBlazoredLocalStorage();
 
-            await builder.Build().RunAsync().ConfigureAwait(true);
```
sslaws, line 85: What's the point of this change?

sslaws (changes requested):

Apps/Admin/Client/Program.cs
rayking29: So that we can get the query string parameter and set the kc_idp_hint and go to the desired login type (IDIR or PHSA).

sslaws (dismissed):

Apps/Admin/Client/Program.cs
```diff
@@ -63,10 +66,23 @@ public static async Task Main(string[] args)
             // Enable Mud Blazor component services
             builder.Services.AddMudServices(config => config.SnackbarConfiguration.PositionClass = Defaults.Classes.Position.BottomRight);
 
+            builder.Services.AddScoped(_ => new HttpClient { BaseAddress = new Uri(builder.HostEnvironment.BaseAddress) });
+
+            WebAssemblyHost[] app = new WebAssemblyHost[1];
```
MikeLyttle, line 71: Using an array here to avoid warnings/errors about variable initialization seems a little hacky, but I'm not sure how best to address that.

MikeLyttle (approved):

MikeLyttle (approved):

sslaws (approved):

PR #3993, opened 2022-08-16
Fix broken Admin functional tests AB#13635

MikeLyttle:
# Fixes [AB#13635](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13635)

## Description

- fixes opening dropdowns in support page tests
    - `{ force: true }` is required because MudSelect puts the data-testid attribute on an element that's hidden
- changes dashboard UI tests to use an unchanging date to avoid time zone issues regardless of where the tests are run
    - `cy.clock()` enables precise control over the time the browser thinks it is
    - moves usage of `cy.clock()` after login
- changes an intercept in the analytics tests to be more resilient to time zone issues regardless of where the tests are run
- pre-emptively increases the timeout values for downloading analytics files
- removes additional login commands in support tests
- removes unused parts of login command

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## Items to Review:

### DIFF
--- a/Apps/Admin/Tests/Functional/cypress/integration/e2e/analytics.cy.js
+++ b/Apps/Admin/Tests/Functional/cypress/integration/e2e/analytics.cy.js
@@ -1,3 +1,5 @@
+const timeout = 45000;
+
 function getFileName(name) {
     // yyyy-mm-dd
     const today = new Date().toLocaleDateString("en-CA");
@@ -18,10 +20,9 @@ describe("System Analytics", () => {
 
         cy.intercept("GET", "**/CsvExport/GetRatings").as("getRatings");
 
-        cy.intercept(
-            "GET",
-            "**/CsvExport/GetInactiveUsers?timeOffset=420&inactiveDays=100"
-        ).as("getInactiveUsers");
+        cy.intercept("GET", "**/CsvExport/GetInactiveUsers*").as(
+            "getInactiveUsers"
+        );
 
         cy.login(
             Cypress.env("keycloak_username"),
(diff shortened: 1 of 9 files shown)


Apps/Admin/Tests/Functional/cypress/integration/e2e/analytics.cy.js
```diff
@@ -18,10 +20,9 @@ describe("System Analytics", () => {
 
         cy.intercept("GET", "**/CsvExport/GetRatings").as("getRatings");
 
-        cy.intercept(
-            "GET",
-            "**/CsvExport/GetInactiveUsers?timeOffset=420&inactiveDays=100"
```
BrianMaki, line 23: Reason why offset and inactiveDays were specified was to ensure that those values were being read in the intercept vs just *.

Apps/Admin/Tests/Functional/cypress/integration/e2e/analytics.cy.js
```diff
@@ -18,10 +20,9 @@ describe("System Analytics", () => {
...
-        ).as("getInactiveUsers");
+        cy.intercept("GET", "**/CsvExport/GetInactiveUsers*").as(
```
BrianMaki, line 23: Reason why offset and inactiveDays were specified was to ensure that those values were being read in the intercept vs just using *.

BrianMaki (changes requested):
See my comment.

Apps/Admin/Tests/Functional/cypress/integration/e2e/analytics.cy.js
MikeLyttle: It seems like it should work with the parameters specified, and it does locally, but it fails when running in CI with this error:

> Timed out retrying after 5000ms: cy.wait() timed out waiting 5000ms for the 1st request to the route: getInactiveUsers. No request ever occurred.

I thought perhaps the parameters were different in dev than locally (due to time zones), which is why I removed them from the intercept.  After checking dev, this doesn't seem to be true, but I'm not sure why else it would be failing.

rayking29 (approved):

Apps/Admin/Tests/Functional/cypress/integration/e2e/analytics.cy.js
BrianMaki: I think you're right.  Hard-coding 420 is not good as that can change with Daylight Savings Time and Timezone of where this is being run.

MikeLyttle: Okay, I'll leave it as it is in the PR for now and we can see if that fixes the CI runs or if something else is at play.

MikeLyttle:
> But I would like to request you run this locally using using cypress run and provide a screenshot.

image

BrianMaki (approved):

rayking29 (approved):

PR #4082, opened 2022-09-01
Update UI and Store to support multiple Hdid vaccinations for AB#13822.

BrianMaki:
# Implements [AB#13822](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13822)

## Description

Update store for VaccinationStatus - authenticatedVaccineRecord to handle multiple Hdids.

- Vaccination download records will be stored in a Map keyed by Hdid.
- DependentCardComponent updated
- HomeView updated
- Covid19View updated
- Spinner when downloading updated
- Status message in spinner when downloading updated
- No Data Found modal updated except for Covid19View, which does not have that implemented
- Some refactoring of function and variable names

Recorded Test:

https://user-images.githubusercontent.com/58790456/187856284-9000154e-5df2-4a5e-a02f-b30b8bb2c6d3.mp4

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

Front end visual no but wiring involving store and components were updated.

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/store/modules/vaccinationStatus/types.ts
+++ b/Apps/WebClient/src/ClientApp/src/store/modules/vaccinationStatus/types.ts
@@ -11,6 +11,7 @@ import CovidVaccineRecord from "@/models/covidVaccineRecord";
 import { StringISODate } from "@/models/dateWrapper";
 import { CustomBannerError, ResultError } from "@/models/errors";
 import { LoadStatus } from "@/models/storeOperations";
+import VaccinationRecord from "@/models/vaccinationRecord";
 import VaccinationStatus from "@/models/vaccinationStatus";
 import { RootState } from "@/store/types";
 
@@ -34,11 +35,9 @@ export interface VaccinationStatusState {
         statusMessage: string;
     };
     authenticatedVaccineRecord: {
-        vaccinationRecord?: CovidVaccineRecord;
-        error?: ResultError;
-        status: LoadStatus;
-        statusMessage: string;
-        resultMessage: string;
+        activeHdid: string;
+        statusChanges: number;
+        vaccinationRecords: Map<string, VaccinationRecord>;
     };
 }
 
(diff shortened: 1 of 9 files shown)


Apps/WebClient/src/ClientApp/src/store/modules/vaccinationStatus/types.ts
```diff
@@ -34,11 +35,9 @@ export interface VaccinationStatusState {
         statusMessage: string;
     };
     authenticatedVaccineRecord: {
-        vaccinationRecord?: CovidVaccineRecord;
-        error?: ResultError;
-        status: LoadStatus;
-        statusMessage: string;
-        resultMessage: string;
+        activeHdid: string;
+        statusChanges: number;
```
MikeLyttle, line 39: I think having these 2 properties could be code smells, but if it's working, I'm okay to merge now and address them later.

Relying on `activeHdid` is a little worrisome because the store is async.

`statusChanges` seems to be there just to have a reliable property that can be watched to see if `vaccinationRecords` has changed.  It should be possible to use `Vue.set` on `vaccinationRecords` in the mutations to make this redundant.

MikeLyttle (approved):

Apps/WebClient/src/ClientApp/src/store/modules/vaccinationStatus/types.ts
BrianMaki: Agreed.  I will re-examine when doing functional tests for AB#13782.

rayking29 (approved):

Apps/WebClient/src/ClientApp/src/store/modules/vaccinationStatus/types.ts
BrianMaki: I have done some more investigation. Unfortunately Vue does not support javascript Map. The use of Vue.set will not work. Looks like only javascript Object and Array have solutions for [reactivity.](https://v2.vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats), which is probably why we see the solution for preferences in user module.

PR #4123, opened 2022-09-09
Update AccessTokenService to use sub as identifier for AB#13884

BrianMaki:
# Implements [AB#13884](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13884)

## Description

1. Updated AccessTokenService to use sub identifier instead of hdid when using cache for access token.
2. Also centralized 'cache key' creation to one spot instead of having the key created in multiple locations, which could result in an error if not all keys were updated if a change was required.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:

### DIFF
--- a/Apps/Common/src/Services/AccessTokenService.cs
+++ b/Apps/Common/src/Services/AccessTokenService.cs
@@ -72,9 +72,9 @@ namespace HealthGateway.Common.Services
         public async Task<RequestResult<TokenSwapResponse>> GetPhsaAccessToken()
         {
             using Activity? activity = Source.StartActivity();
-            string? identifier = this.authenticationDelegate.FetchAuthenticatedUserIdentifier();
+            string? userId = this.authenticationDelegate.FetchAuthenticatedUserIdentifier();
             RequestResult<TokenSwapResponse> requestResult = new();
-            string cacheKey = $"{TokenSwapCacheDomain}:Identifier:{identifier}";
+            string cacheKey = $"{TokenSwapCacheDomain}:{userId}";
             TokenSwapResponse? cachedAccessToken = this.GetFromCache(cacheKey);
 
             if (cachedAccessToken == null)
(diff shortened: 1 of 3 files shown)


rayking29 (dismissed):

Apps/Common/src/Services/AccessTokenService.cs
```diff
@@ -72,25 +72,26 @@ public class AccessTokenService : IAccessTokenService
         public async Task<RequestResult<TokenSwapResponse>> GetPhsaAccessToken()
         {
             using Activity? activity = Source.StartActivity();
-            string? hdid = this.authenticationDelegate.FetchAuthenticatedUserHdid();
+            string? identifier = this.authenticationDelegate.FetchAuthenticatedUserIdentifier();
```
sslaws, line 75: nit: userid is probably better than identifier

Apps/Common/src/Services/AccessTokenService.cs
```diff
@@ -72,25 +72,26 @@ public class AccessTokenService : IAccessTokenService
...
             RequestResult<TokenSwapResponse> requestResult = new();
-            TokenSwapResponse? cachedAccessToken = this.GetFromCache(hdid);
+            string cacheKey = $"{TokenSwapCacheDomain}:Identifier:{identifier}";
```
sslaws, line 77: I don't think we need to use Identifier in the cacheKey, hdid actually meant something but sub is less useful.   \${TokenSwapCacheDomain}:{userId} is probably better.

sslaws (changes requested):

Apps/Common/src/Services/AccessTokenService.cs
BrianMaki: Changed.

Apps/Common/src/Services/AccessTokenService.cs
BrianMaki: Changed.

rayking29 (approved):

sslaws (approved):

PR #4247, opened 2022-10-12
Write db delegate queries for year of birth counts

rayking29:
# Fixes or Implements [AB#14066](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14066)

## Description

- added method for year of birth to profile delegate
- updated csv export service with new method to get year of birth counts
- changed GetStream method to implement using statements so no need for #pragma warning disable CA2000

* note this logic still needs to be tested with the upcoming Blazor changes, so pushing it to the feature branch

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -19,7 +19,6 @@ namespace HealthGateway.Database.Delegates
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
-    using System.Text.Json;
     using HealthGateway.Database.Constants;
     using HealthGateway.Database.Context;
     using HealthGateway.Database.Models;
@@ -31,8 +30,8 @@ namespace HealthGateway.Database.Delegates
     [ExcludeFromCodeCoverage]
     public class DBProfileDelegate : IUserProfileDelegate
     {
-        private readonly ILogger logger;
         private readonly GatewayDbContext dbContext;
+        private readonly ILogger logger;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DBProfileDelegate"/> class.
@@ -77,7 +76,7 @@ namespace HealthGateway.Database.Delegates
             if (result.Status == DBStatusCode.Read)
             {
                 // Copy certain attributes into the fetched User Profile
-                result.Payload!.Email = profile.Email;
+                result.Payload.Email = profile.Email;
                 result.Payload.TermsOfServiceId = profile.TermsOfServiceId;
                 result.Payload.UpdatedBy = profile.UpdatedBy;
                 result.Payload.Version = profile.Version;
@@ -278,6 +277,24 @@ namespace HealthGateway.Database.Delegates
                 .Take(limit);
             result.Status = DBStatusCode.Read;
             return result;
+        }
+
+        /// <inheritdoc/>
+        public IDictionary<string, int> GetLoggedInUserYearOfBirthCounts(DateTime startDate, DateTime endDate)
+        {
+            Dictionary<string, int> yobCount = this.dbContext.UserProfile
+                .Select(x => new { x.HdId, x.LastLoginDateTime, x.YearOfBirth })
+                .Concat(
+                    this.dbContext.UserProfileHistory.Select(x => new { x.HdId, x.LastLoginDateTime, x.YearOfBirth }))
+                .Where(x => x.YearOfBirth != null && x.LastLoginDateTime >= startDate && x.LastLoginDateTime <= endDate)
+                .Select(x => new { x.HdId, lastLoginDate = GatewayDbContext.DateTrunc("days", x.LastLoginDateTime), x.YearOfBirth })
+                .Distinct()
+                .GroupBy(x => x.YearOfBirth)
+                .Select(x => new { yearOfBirth = x.Key, count = x.Count() })
+                .OrderBy(x => x.yearOfBirth)
+                .ToDictionary(x => x.yearOfBirth!, x => x.count);
+
+            return yobCount;
         }
     }
 }
--- a/Apps/Database/src/Delegates/DBProfileDelegate.cs
+++ b/Apps/Database/src/Delegates/DBProfileDelegate.cs
@@ -287,14 +287,13 @@ namespace HealthGateway.Database.Delegates
                 .Concat(
                     this.dbContext.UserProfileHistory.Select(x => new { x.HdId, x.LastLoginDateTime, x.YearOfBirth }))
                 .Where(x => x.YearOfBirth != null && x.LastLoginDateTime >= startDate && x.LastLoginDateTime <= endDate)
-                .Select(x => new { x.HdId, lastLoginDate = GatewayDbContext.DateTrunc("days", x.LastLoginDateTime), x.YearOfBirth })
+                .Select(x => new { x.HdId, x.YearOfBirth })
                 .Distinct()
                 .GroupBy(x => x.YearOfBirth)
                 .Select(x => new { yearOfBirth = x.Key, count = x.Count() })
-                .OrderBy(x => x.yearOfBirth)
                 .ToDictionary(x => x.yearOfBirth!, x => x.count);
 
-            return yobCount;
+            return new SortedDictionary<string, int>(yobCount);
         }
     }
 }
--- a/Apps/Admin/Server/Services/CsvExportService.cs
+++ b/Apps/Admin/Server/Services/CsvExportService.cs
@@ -35,12 +35,12 @@ namespace HealthGateway.Admin.Server.Services
     {
         private const int PageSize = 100000;
         private const int Page = 0;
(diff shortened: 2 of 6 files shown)


Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -279,5 +278,23 @@ public DBResult<IEnumerable<UserProfileHistory>> GetUserProfileHistories(string
...
         }
+
+        /// <inheritdoc/>
+        public IDictionary<string, int> GetLoggedInUserYearOfBirthCounts(DateTime startDate, DateTime endDate)
+        {
+            Dictionary<string, int> yobCount = this.dbContext.UserProfile
+                .Select(x => new { x.HdId, x.LastLoginDateTime, x.YearOfBirth })
+                .Concat(
+                    this.dbContext.UserProfileHistory.Select(x => new { x.HdId, x.LastLoginDateTime, x.YearOfBirth }))
+                .Where(x => x.YearOfBirth != null && x.LastLoginDateTime >= startDate && x.LastLoginDateTime <= endDate)
```
BrianMaki, line 289: What is the generated SQL for this call?  Date is stored as UTC in the database but this is most likely being called from PST.  You all need to pass in an offset so the right time period is being queried.  See GetDailyLoggedInUsersCount but you will need to confirm when running the SQL itself.  To test this, put GetLoggedINUserYearOfBirthCounts call into DashboardService.GetDailyLoggedInUsersCount() and then put the following into your appsettings local json file to see the generated SQL in the console.     "Logging": {
        "LogLevel": {
            "Default": "Debug"
        },
        "SensitiveDataLogging": {
            "Enabled": true
        }
    },

BrianMaki (changes requested):
Please see comments.  Please run generated SQL in something like PGAdmin to see explain plan.

Apps/Database/src/Delegates/DBProfileDelegate.cs
```diff
@@ -279,5 +278,23 @@ public DBResult<IEnumerable<UserProfileHistory>> GetUserProfileHistories(string
...
+                .Select(x => new { x.HdId, lastLoginDate = GatewayDbContext.DateTrunc("days", x.LastLoginDateTime), x.YearOfBirth })
+                .Distinct()
+                .GroupBy(x => x.YearOfBirth)
+                .Select(x => new { yearOfBirth = x.Key, count = x.Count() })
+                .OrderBy(x => x.yearOfBirth)
+                .ToDictionary(x => x.yearOfBirth!, x => x.count);
```
MikeLyttle, line 295: The Select() and OrderBy() are not needed.

rayking29: I'm pretty sure that's how I get the key and value for the dictionary.  The orderby is to return the list in order of year of birth for a preliminary sort.

Apps/Database/src/Delegates/DBProfileDelegate.cs
rayking29: Added offset for UTC time.

Apps/Database/src/Delegates/DBProfileDelegate.cs
MikeLyttle: ToDictionary() should be able to perform the same transformation that the Select() is doing, i.e.

```
ToDictionary(x => x.Key!, x => x.Count())
```

Using OrderBy() may alter the results, but it is not reliable since Dictionaries have no guaranteed order.  If we wanted the results to be sorted, we could return a SortedDictionary, passing in the unsorted Dictionary to its constructor.

Apps/Database/src/Delegates/DBProfileDelegate.cs
BrianMaki: What is the generated SQL?

Right now both start date and end date param have been changed?  What about the UserProfile.LastLoginDateTime value that both start date and end date are comparing against? By adding an offset to both start date and end date, those parameters are now in the timezone you want them in but what about the UserProfile.LastLoginDateTime and UserProfileHistory.LastLoginDateTime?  Shouldn't last login date time also be adjusted to the time offset?  Example: UserProfile.LastLoginDateTime is October 10, 2022 06:59:00 UTC, which is 23:59:00 PST previous day (October 9, 2022).  If user enters October 5 and 10 at 5:00 am PST in the morning for start date and end date respectively, does this code change make start date October 4 and end date October 9 with the timespan offset change? If end date is changed to October 9. then it will  not pick up the record that had UserProfile.LastLoginDateTime of October 10, 2022 06:59:00 UTC.

rayking29: But last login datetime comes from the database so is already in UTC. So if I use the offset on the start date and end date that will come from the front end in pst, then the comparison is all in UTC.
Either that or apply the offset to the last login datetime only to make it pst like the start and end dates.  Should be the same result.

BrianMaki: Have a look at GetDailyLoggedInUsersCount()? That does an UTC adjustment.  GatewayDbContext.DateTrunc("days", x.LastLoginDateTime.AddMinutes(offset.TotalMinutes))

BrianMaki: I think you need both input parameters and what is being compared to in the database in the same timezone.

rayking29: The database datetimes are in UTC already are they not? Therefore LastLoginDateTime is in UTC when we select it from the database in the Select statements.
Look at DashboardService.GetRatingSummary, it is doing the same thing I am, it converts the startdate and enddate to UTC and then compares to the database datetime, which is already in UTC.
The example of GetDailyLoggedInUsersCount does not do this compare to start and end dates, it is returning the result in pst time for the front end to display because the LastLoginDateTime is in UTC in the database.

BrianMaki (approved):

Apps/Database/src/Delegates/DBProfileDelegate.cs
BrianMaki: What is the SQL saying?

BrianMaki: - Where clause is only being applied to UserProfileHistory.  It will need to be applied to **UserProfile** as well.
- For performance (analyze explain plan in PG Admin) you may need to modify the LINQ query to move the WHERE clause to associate with the SELECT DISTINCT 

- t."LastLoginDateTime" will be in UTC time.  Example: I just logged into Healthgateway at Wed Oct 12 11:03 PM.  In the database, the time is recorded as Thu Oct 13 06:03.  So if your start date is Monday, Oct 10 and end date is Wednesday, Oct 12, the above query should pick up a login that occurs on Wednesday, Oct 12 at 11:03 PM.

Screen Shot 2022-10-12 at 11 06 35 PM

Apps/Database/src/Delegates/DBProfileDelegate.cs
rayking29: I removed the OrderBy.
I tried doing the ToDictionary() as you suggested but I get an error that it can't interpret x.YearOfBirth.  I think it needs the select as part of the query to get the result then it can be put in the dictionary.
I have asked Nino what he wants to do for sorting. If it is needed I will use a SortedDictionary.

Apps/Database/src/Delegates/DBProfileDelegate.cs
rayking29: I removed the OrderBy as per Mike's suggestion. This is the resulting code: 

Screen Shot 2022-10-13 at 1 55 17 PM

Here is the SQL that is created from it along with the Explain Analyze from PgAdmin:

Screen Shot 2022-10-13 at 1 56 38 PM

1. I have tested it with local data and it is giving the correct results as far as I am expecting.
2. The Where clause is applied to both the UserProfile and UserProfileHistory tables.
3. The explain plan looks good from what I can see, no extra loops or scans. I could try to change it to do the Where clause on the Select Distinct if you think that is necessary.

MikeLyttle: I think we can remove `lastLoginDate = GatewayDbContext.DateTrunc("days", x.LastLoginDateTime)` from the query.  We've already identified the users we are interested in by filtering on `LastLoginDateTime`.

rayking29: That is there because we don't want all the logins we want it to only count one per day if there are multiple.

Apps/Database/src/Delegates/DBProfileDelegate.cs
MikeLyttle: It looks like having the Select() statement is preventing a LINQ to SQL issue so keeping it seems fine.

MikeLyttle (approved):

BrianMaki (approved):

PR #4378, opened 2022-11-11
Update UI to use v2 patient endpoint

BrianMaki:
# Implements [AB#14271](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14271)

## Description

- Refactored name and location of ApiPatientException to ApiException
- Refactored location of ApiProblemDetails
- Updated services, delegates and units based on refactoring
- Updated UI to use v2 patient endpoint
- JSON result will only return resource payload and warning if they are not null.  If waning object is null, then it will not be included in the JSON.  Same goes for resource payload.
- Rider format changes applied on some files that had not been previously affected

**200:**

Screenshot 2022-11-10 at 3 56 56 PM

**404:**

Screenshot 2022-11-10 at 2 39 42 PM

**500:**

Screenshot 2022-11-10 at 3 40 55 PM

**Warning:**

Screenshot 2022-11-10 at 3 44 40 PM

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

No but action, service and model were updated.

## Items to Review:

### DIFF
--- a/Apps/CommonData/src/ErrorHandling/ApiException.cs
+++ b/Apps/CommonData/src/ErrorHandling/ApiException.cs
@@ -13,69 +13,70 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 // -------------------------------------------------------------------------
-namespace HealthGateway.Common.Exceptions
+namespace HealthGateway.Common.Data.ErrorHandling
 {
     using System;
     using System.Net;
+    using System.Runtime.Serialization;
     using HealthGateway.Common.Data.Models.ErrorHandling;
 
     /// <summary>
     /// Represents a custom api patient exception.
     /// </summary>
     [Serializable]
-    public class ApiPatientException : Exception
+    public class ApiException : Exception
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="ApiPatientException"/> class.
+        /// Initializes a new instance of the <see cref="ApiException"/> class.
         /// </summary>
         /// <param name="detail">The detail associated with the exception.</param>
         /// <param name="instance">The instance associated with the exception.</param>
         /// <param name="statusCode">The http status code associated with the exception.</param>
-        public ApiPatientException(string detail, string instance, HttpStatusCode statusCode)
+        public ApiException(string detail, string instance, HttpStatusCode statusCode)
         {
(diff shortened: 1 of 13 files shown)


Apps/CommonData/src/ErrorHandling/ApiException.cs
```diff
@@ -13,69 +13,70 @@
...
         /// <param name="instance">The instance associated with the exception.</param>
         /// <param name="statusCode">The http status code associated with the exception.</param>
-        public ApiPatientException(string detail, string instance, HttpStatusCode statusCode)
+        public ApiException(string detail, string instance, HttpStatusCode statusCode)
         {
-            this.ProblemType = "api-patient-exception";
+            this.ProblemType = "api-exception";
             this.Detail = detail;
-            this.Title = "Api Patient Service Exception";
+            this.Title = "Custom Exception Handling";
```
sslaws, line 39: I think the title should be something like Health Gateway Exception

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
```diff
@@ -306,19 +305,9 @@ export const actions: UserActions = {
                                 resolve();
                             });
                     } else {
-                        if (result.resultError?.statusCode === 429) {
```
sslaws, line 309: This code doesn't look right but don't we need 429 logic still?

sslaws (changes requested):

MikeLyttle (approved):

Apps/CommonData/src/ErrorHandling/ApiException.cs
BrianMaki: Changed.

Apps/WebClient/src/ClientApp/src/store/modules/user/actions.ts
BrianMaki: @sslaws  Spoke with @MikeLyttle  about this.  That line of code was erroneous.  The 429 is handled in the catch block on line 314.

sslaws (approved):

PR #4380, opened 2022-11-11
Reduce Rider Warnings 4

rayking29:
# Implements [AB#14299](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14299)

## Description

Reduce Rider Warnings - Rider 4

- Fixes to unit tests
- Fix expression is always true according to nullable reference types' annotations
- Fix naming, nullable, using, and namespaces

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## Items to Review:

### DIFF
--- a/Apps/Common/test/unit/Delegates/AesCryptoDelegateTests.cs
+++ b/Apps/Common/test/unit/Delegates/AesCryptoDelegateTests.cs
@@ -14,7 +14,6 @@
 //  limitations under the License.
 // -------------------------------------------------------------------------
 
-// ReSharper disable CollectionNeverUpdated.Local
 namespace HealthGateway.CommonTests.Delegates
 {
     using System;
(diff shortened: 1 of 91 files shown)


Apps/Common/test/unit/Delegates/AesCryptoDelegateTests.cs
```diff
@@ -13,6 +13,8 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 // -------------------------------------------------------------------------
+
+// ReSharper disable CollectionNeverUpdated.Local
```
MikeLyttle, line 17: We don't need to suppress this; we can resolve it by refactoring.

rayking29: This was in several places and it is specifically testing an empty collection. So it is legitimate and would not ever be updated. Not sure if that can be refactored.

rayking29: Fixed by inlining the collection.

sslaws (approved):

MikeLyttle (approved):

PR #4404, opened 2022-11-17
Oust HttpClientService #4 for AB#14314

sslaws:
# Fixes or Implements [AB#14314](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14314)

## Description
Moves the FileDownload Interface and service into DBMaintainer and removes usage of HttpClientService
Ii removed the unit tests for FileDownload as they were trivial and didn't have a new home.
Refactors SalesforceDelegate to use Refit instead of HttpClientService

If someone can figure out how to use use awaiting I'll update to remove the SuppressMessage.

## Testing

- [X] Unit Tests Updated
- [ ] Not Required

## UI Changes
None

## Notes
Unit Tests are not pretty but refactoring is beyond the scope of this ticket.

## Items to Review:


Apps/Medication/src/Delegates/SalesforceDelegate.cs
```diff
@@ -98,71 +98,20 @@ public async Task<RequestResult<IList<MedicationRequest>>> GetMedicationRequests
...
-                        default:
-                            retVal.ResultError = new RequestResultError
-                            {
-                                ResultMessage = $"Unable to connect to Medication Requests endpoint, HTTP Error {response.StatusCode}",
-                                ErrorCode = ErrorTranslator.ServiceError(ErrorType.CommunicationExternal, ServiceType.Sf),
-                            };
-                            this.logger.LogError("Unable to connect to endpoint {Endpoint}, HTTP Error {StatusCode}\n{Payload}", endpoint, response.StatusCode, payload);
-                            break;
-                    }
+                    ResponseWrapper? replyWrapper = await this.specialAuthorityApi.GetSpecialAuthorityRequests(phn, accessToken).ConfigureAwait(true);
```
BrianMaki, line 103: Line 136 check for NoContent is removed.  Are you sure NoContent will not result in an ApiException.  Have you tried confirming this in a unit test? Have you tried mocking the response to return No Content?

var apiMockedResponse = 
        setupApiRequest.ReturnsAsync(new HttpResponseMessage() 
        { 
            StatusCode = HttpStatusCode.NoContent, 
            Content = null, 
        });

Apps/Medication/test/unit/Delegates/SalesforceDelegateTests.cs
```diff
@@ -290,16 +283,16 @@ public void ShouldEmptyIfNoContent()
...
-            using HttpResponseMessage httpResponseMessage = new()
-            {
-                StatusCode = HttpStatusCode.NoContent,
-            };
-            Mock<IHttpClientService> mockHttpClient = CreateHttpClient(httpResponseMessage, phn, authorizationJwt.AccessToken);
+            ResponseWrapper? result = null;
+            Mock<ISpecialAuthorityApi> mockSpecialAuthorityApi = new();
+            mockSpecialAuthorityApi
+                .Setup(s => s.GetSpecialAuthorityRequests(It.IsAny<string>(), It.IsAny<string>()))
+                .ReturnsAsync(result);
```
BrianMaki, line 290: This does not test HttpStatusCode.NoContent.  This assumes when there is a NoContent response, null is returned and there is not an exception occurring i.e., 200 and 204 are treated the same when deserializing.

BrianMaki (changes requested):

Apps/Medication/src/Delegates/SalesforceDelegate.cs
sslaws: I tested the code against a user without any Special Authority requests and the wrapper object was still returned.  I think that No Content has spread in our code base when in most places it is not used.

BrianMaki (dismissed):

Apps/Medication/src/Delegates/SalesforceDelegate.cs
```diff
@@ -18,33 +18,33 @@ namespace HealthGateway.Medication.Delegates
...
     /// <summary>
     /// Salesforce Implementation that retrieves Medication Requests.
     /// </summary>
     public class SalesforceDelegate : IMedicationRequestDelegate
     {
-        private const string SalesforceConfigSectionKey = "Salesforce";
+        /// <summary>
+        /// The key used to lookup Salesforce configuration.
+        /// </summary>
+        public const string SalesforceConfigSectionKey = "Salesforce";
```
MikeLyttle, line 44: Since these config section constants no longer need to be accessed inside the delegates, I think we should move them into the appropriate Config model, i.e. HealthGateway.Medication.Models.Salesforce.Config.

Apps/Medication/src/Delegates/SalesforceDelegate.cs
```diff
@@ -98,71 +98,20 @@ public async Task<RequestResult<IList<MedicationRequest>>> GetMedicationRequests
...
+                    retVal.ResourcePayload = replyWrapper != null ? this.autoMapper.Map<IEnumerable<SpecialAuthorityRequest>, IList<MedicationRequest>>(replyWrapper.Items) : new List<MedicationRequest>();
+                    retVal.TotalResultCount = retVal.ResourcePayload?.Count;
+                    retVal.PageSize = retVal.ResourcePayload?.Count;
+                    retVal.PageIndex = 0;
+                    retVal.ResultStatus = ResultType.Success;
                 }
-#pragma warning disable CA1031 // Do not catch general exception types
-                catch (Exception e)
-#pragma warning restore CA1031 // Do not catch general exception types
+                catch (Exception e) when (e is ApiException or HttpRequestException)
```
MikeLyttle, line 110: This `when` clause is super handy for catching Exceptions!

Apps/Medication/test/unit/Delegates/SalesforceDelegateTests.cs
sslaws: I've removed this test as it was obsolete.

MikeLyttle (approved):

BrianMaki (approved):

PR #4417, opened 2022-11-21
Fix draft communications shown for AB#14244

rayking29:
# Fixes [AB#14244](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14244)

## Description

Draft Communications - Communications that are set as "Draft" are still being displayed on landing/in-app banners. 
- added check that communication status code is 'new' before adding to cache

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/Common/src/Services/CommunicationService.cs
+++ b/Apps/Common/src/Services/CommunicationService.cs
@@ -142,7 +142,8 @@ namespace HealthGateway.Common.Services
                     {
                         this.logger.LogInformation("{Action} ChangeEvent for Communication {Id} found in Cache", changeEvent.Action, communication.Id);
                         this.RemoveCommunicationFromCache(communication.CommunicationTypeCode);
-                        if (changeEvent.Action is Insert or Update)
+                        if (changeEvent.Action is Insert or Update &&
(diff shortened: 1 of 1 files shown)


Apps/Common/src/Services/CommunicationService.cs
```diff
@@ -142,7 +142,8 @@ public void ProcessChange(BannerChangeEvent changeEvent)
                     {
                         this.logger.LogInformation("{Action} ChangeEvent for Communication {Id} found in Cache", changeEvent.Action, communication.Id);
                         this.RemoveCommunicationFromCache(communication.CommunicationTypeCode);
-                        if (changeEvent.Action is Insert or Update)
```
sslaws, line 145: I think this may cause a problem with hitting the DB when no result is cached... we probably need to create an empty cache item

rayking29: The result is the same as when doing a Delete. Not sure what DB problem this would cause.

BrianMaki (approved):

sslaws (approved):

PR #4423, opened 2022-11-22
ISupportUserApi - refactor away from IApiResponse<T> for AB#14330.

BrianMaki:
# Fixes [AB#14330](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14330)

## Description

- Refactor away from IApiResponse<T> and instead return the T for ISupportApi
- Updated StoreUtility to be more robust when handling parameters
- Updated StoreUtility to handle Exception to get message versus ApiException as HttpRequestException can also. be passed in.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Utils/StoreUtility.cs
+++ b/Apps/Admin/Client/Utils/StoreUtility.cs
@@ -30,7 +30,7 @@ namespace HealthGateway.Admin.Client.Utils
         /// </summary>
         /// <param name="resultError">An error returned from the server.</param>
         /// <returns>The generated <see cref="RequestError"/>.</returns>
-        public static RequestError FormatRequestError(RequestResultError resultError)
+        public static RequestError FormatRequestError(RequestResultError? resultError)
         {
             return FormatRequestError(null, resultError);
         }
--- a/Apps/Admin/Client/Utils/StoreUtility.cs
+++ b/Apps/Admin/Client/Utils/StoreUtility.cs
@@ -16,9 +16,9 @@
 
 namespace HealthGateway.Admin.Client.Utils
 {
+    using System;
     using HealthGateway.Admin.Client.Store;
     using HealthGateway.Common.Data.ViewModels;
-    using Refit;
 
     /// <summary>
     /// Utilities for interacting with the application store.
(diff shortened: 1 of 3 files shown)


Apps/Admin/Client/Utils/StoreUtility.cs
```diff
@@ -28,10 +28,20 @@ public static class StoreUtility
...
+            return FormatRequestError(null, resultError);
+        }
+
+    /// <summary>
+        /// Formats errors returned from external requests.
+        /// </summary>
+        /// <param name="exception">An exception returned by Refit.</param>
+        /// <param name="resultError">An error returned from the server.</param>
+        /// <returns>The generated <see cref="RequestError"/>.</returns>
+        public static RequestError FormatRequestError(Exception? exception, RequestResultError? resultError = null)
```
sslaws, line 44: Why not default exception to null as well?

sslaws (changes requested):

Apps/Admin/Client/Utils/StoreUtility.cs
BrianMaki: That would not work as I had to overload with public static RequestError FormatRequestError(RequestResultError? resultError).  Having exception default to null and result error default to null will not allow StoreUtility.FormatRequestError(response?.ResultError) to compile.  Even though exception defaults to null, it is still expecting an exception to be populated.

Here is what the code supports:

- StoreUtility.FormatRequestError(e) - this is with an exception
- StoreUtility.FormatRequestError(response?.ResultError) - this is with ResultError
- StoreUtility.FormatRequestError(response.Error, response.Content?.ResultError) - this continues to support existing implementations.

sslaws (approved):

Apps/Admin/Client/Utils/StoreUtility.cs
MikeLyttle: What do you think about exposing 3 public methods that have non-null parameters (one for Exception, one for RequestResultError, one for both), and have all 3 call the private method that accepts both as nullable?

MikeLyttle: Oh, maybe we need to support the nullable on RequestResultError to make it easier to pass in ResultError.

MikeLyttle (approved):

PR #4427, opened 2022-11-22
IPersonalAccountsApi - Refactor away from IApiResponse<T>

rayking29:
# Fixes [AB#14336](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14336)

## Description

IPersonalAccountsApi - Refactor away from IApiResponse<T>
- removed IApiResponse and refactored code
- updated and removed unit tests
- updated method names for async naming standard

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

Screenshot 2022-11-22 at 1 31 27 PM

## Items to Review:

### DIFF
--- a/Apps/Common/test/unit/Services/PersonalAccountServiceTests.cs
+++ b/Apps/Common/test/unit/Services/PersonalAccountServiceTests.cs
@@ -18,9 +18,7 @@ namespace HealthGateway.CommonTests.Services
     using System;
     using System.Collections.Generic;
     using System.Linq;
-    using System.Net;
     using System.Net.Http;
-    using System.Threading.Tasks;
     using HealthGateway.Common.Api;
     using HealthGateway.Common.CacheProviders;
     using HealthGateway.Common.Data.Constants;
@@ -30,7 +28,6 @@ namespace HealthGateway.CommonTests.Services
     using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.Logging;
     using Moq;
-    using Refit;
     using Xunit;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


sslaws (approved):

Apps/Common/test/unit/Services/PersonalAccountServiceTests.cs
```diff
@@ -82,31 +79,6 @@ public void ShouldGetPatientAccountUseCache()
             Assert.Equal(id, actualResult.ResourcePayload?.Id);
         }
 
-        /// <summary>
```
BrianMaki, line 85: Where is the unit test to handle ApiException and HttpRequestException.  Won't his have on code coverage?

BrianMaki (changes requested):

Apps/Common/test/unit/Services/PersonalAccountServiceTests.cs
```diff
@@ -82,31 +79,6 @@ public void ShouldGetPatientAccountUseCache()
...
-        /// Get Personal Account returns ApiException.
-        /// </summary>
-        [Fact]
-        public void ShouldGetPatientAccountReturnsApiException()
```
BrianMaki, line 89: Where is the unit test to handle ApiException?. Won't this have an effect on code coverage?  Looks like only HttpRequestException is being handled.

rayking29:
I can't find this comment on my PR.

Either type of exception will go to the same exception code block so it won't have any effect on code coverage.
________________________________

Apps/Common/test/unit/Services/PersonalAccountServiceTests.cs
rayking29: See my previous comment. There shouldn't be any effect on code coverage. This is the same as the test changes that Stephen checked in earlier, one test for an exception.

BrianMaki (approved):

PR #4465, opened 2022-12-01
Fix logout issue by using cypress session to store login

BrianMaki:
# Fixes [AB#14397](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14397)

## Description

Cypress session documentation: [https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/](url)

There have been issues with keycloak logout in tests whether it be in beforeEach or having multiple login/lgout attempts in a single test.

To solve this issue and also improve performance, the use of  cypress session to store logins (username/password) has been implemented for this fix.

- In login call use cypress session to store login by username and auth method as key
- Enable experimentalSessionAndOrigin in cypress.json
- Reverted previous changes to dependent functional test i.e., uncommented out code that was failing.
- For some functional tests the session had to be cleared. This is due to some tests logging out of Healthgateway and this would cause issue with the next test expecting user to be logged in.  Also, some tests are submitting requests between pages and these values stay in the session.

Screenshot 2022-12-01 at 9 43 38 AM

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:

### DIFF
--- a/Testing/functional/tests/cypress/integration/e2e/covid19/pcrTestKit.js
+++ b/Testing/functional/tests/cypress/integration/e2e/covid19/pcrTestKit.js
@@ -31,6 +31,7 @@ const processedBanner = "[data-testid=alreadyProcessedBanner]";
 
 describe("Authenticated PCR Test Kit Registration", () => {
     beforeEach(() => {
+        Cypress.session.clearAllSavedSessions();
         cy.enableModules("PcrTest");
         cy.login(
             Cypress.env("keycloak.username"),
--- a/Testing/functional/tests/cypress/integration/e2e/covid19/pcrTestKit.js
+++ b/Testing/functional/tests/cypress/integration/e2e/covid19/pcrTestKit.js
@@ -31,7 +31,6 @@ const processedBanner = "[data-testid=alreadyProcessedBanner]";
 
 describe("Authenticated PCR Test Kit Registration", () => {
     beforeEach(() => {
-        Cypress.session.clearAllSavedSessions();
         cy.enableModules("PcrTest");
         cy.login(
             Cypress.env("keycloak.username"),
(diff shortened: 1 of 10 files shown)


Testing/functional/tests/cypress/integration/e2e/covid19/pcrTestKit.js
```diff
@@ -31,6 +31,7 @@ const processedBanner = "[data-testid=alreadyProcessedBanner]";
 
 describe("Authenticated PCR Test Kit Registration", () => {
     beforeEach(() => {
+        Cypress.session.clearAllSavedSessions();
```
MikeLyttle, line 34: I suggest we clear the sessions right after actions in the tests that invalidate them (e.g. logouts), instead of at the start of tests (where they're not always needed).

BrianMaki: Taking the clear session out of the beforeEach and putting it in at the end of a test causes some tests to fails. Putting in the beforeEach ensures there is a new login for those tests.  As per description, there are some tests that are being disrupted by tests that have logouts and by tests that maintain data in the session.

MikeLyttle (dismissed):

Testing/functional/tests/cypress/integration/e2e/covid19/pcrTestKit.js
BrianMaki: Moved Cypress.session.clearAllSavedSessions() to an afterEach block.  Bug AB#14444 has been created to find a more elegant solution.

MikeLyttle (approved):

rayking29 (approved):

PR #4466, opened 2022-12-01
Prevent awkward word wrapping on hyphenated text

rayking29:
# Implements [AB#14375](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14375)

## Description

Prevent awkward word wrapping on hyphenated text
- added 'text-nowrap' class to dates in tables and timeline cards
- changed hyphen to non-breaking hyphen in text for things like 'Covid-19', phone numbers, etc.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/DependentCardComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/DependentCardComponent.vue
@@ -858,7 +858,7 @@ export default class DependentCardComponent extends Vue {
                     @click="fetchCovid19LaboratoryTests"
                 >
                     <template #title>
-                        <div data-testid="covid19TabTitle">COVID-19</div>
+                        <div data-testid="covid19TabTitle">COVID‑19</div>
                     </template>
                     <div class="p-3">
                         <div class="d-flex justify-content-center">
@@ -898,7 +898,7 @@ export default class DependentCardComponent extends Vue {
                         borderless
                         :items="testRows"
                         class="w-100 mb-0"
-                        aria-describedby="COVID-19 Test Results"
+                        aria-describedby="COVID‑19 Test Results"
                     >
                         <b-thead>
                             <b-tr>
@@ -921,7 +921,7 @@ export default class DependentCardComponent extends Vue {
                             <b-tr v-for="(row, index) in testRows" :key="index">
                                 <b-td
                                     data-testid="dependentCovidTestDate"
-                                    class="align-middle"
+                                    class="align-middle text-nowrap"
                                 >
                                     {{ formatDate(row.test.collectedDateTime) }}
                                 </b-td>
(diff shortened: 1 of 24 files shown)


Apps/WebClient/src/ClientApp/src/components/DependentCardComponent.vue
```diff
@@ -898,7 +898,7 @@ export default class DependentCardComponent extends Vue {
                         borderless
                         :items="testRows"
                         class="w-100 mb-0"
-                        aria-describedby="COVID-19 Test Results"
+                        aria-describedby="COVID‑19 Test Results"
```
MikeLyttle, line 901: Unrelated, but this aria tag is incorrect.  The [accessibility docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) say:

> The aria-describedby attribute is very similar to [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) attribute. While the aria-labelledby lists the ids of the labels or elements that describe the essence of an object, the aria-describedby lists the ids of the descriptions or elements providing more information that the user might need. Both aria-labelledby and aria-describedby reference other elements to calculate a text alternative, but a label should be concise, while a description is intended to provide more verbose information; a label describes the essence of an object, while a description provides more information that the user might need.

So, since we have an element above that labels this table (the \<h4\>), we should be 1) using `aria-labelledby` and 2) providing it id of that element instead of text.  We'll need to update the \<h4\> to have an id.

Apps/WebClient/src/ClientApp/src/store/modules/laboratory/actions.ts
```diff
@@ -336,7 +336,7 @@ export const actions: LaboratoryActions = {
                 customBannerError.description =
                     "The information you entered does not match our records. Please try again.";
                 customBannerError.detail =
-                    "Please note that it can take up to 48 hours from the time of test before a result is available. If it has been at least 48 hours since you tested, please contact the COVID-19 Results Line (1‐833‐707‐2792) to investigate the issue.";
+                    "Please note that it can take up to 48 hours from the time of test before a result is available. If it has been at least 48 hours since you tested, please contact the COVID‑19 Results Line (1‐833‐707‐2792) to investigate the issue.";
```
MikeLyttle, line 339: Hyphens on phone number missed here.

MikeLyttle (approved):

BrianMaki (approved):

Apps/WebClient/src/ClientApp/src/components/DependentCardComponent.vue
MikeLyttle: I've created a separate bug to track this issue: AB#14443.

Apps/WebClient/src/ClientApp/src/store/modules/laboratory/actions.ts
rayking29: Actually they were already changed to non-breaking before but the Covid-19 had been missed.

PR #4477, opened 2022-12-05
Date Extending beyond Timeline Card

rayking29:
# Fixes [AB#14448](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14448)

## Description

Date Extending beyond Timeline Card

After a discussion with Mandad:
- applied the secondary text sizing of .875rem to the timeline date
- added "text-nowrap" at column level and removed from the span
- the only way I could prevent this bug's date-extending problem while using "text-nowrap" was to remove the cols="4" setting from the column
- I tried using non-breaking hyphens in the date, but it made no difference to the behavior

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

Screenshot 2022-12-05 at 2 02 26 PM

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/timeline/entryCard/EntrycardTimelineComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/timeline/entryCard/EntrycardTimelineComponent.vue
@@ -141,9 +141,9 @@ export default class EntrycardTimelineComponent extends Vue {
                                 <strong>{{ displayTitle }}</strong>
                             </span>
                         </b-col>
-                        <b-col cols="4" class="text-right">
+                        <b-col class="text-right text-nowrap">
                             <span
-                                class="text-muted text-nowrap"
+                                class="text-muted entry-card-date"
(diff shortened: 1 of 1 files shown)


Apps/WebClient/src/ClientApp/src/components/timeline/entryCard/EntrycardTimelineComponent.vue
```diff
@@ -285,4 +285,8 @@ div[class*=" row"] {
     padding: 0px;
     color: $primary;
 }
+
+.entry-card-date {
+    font-size: 0.875rem;
+}
```
MikeLyttle, line 291: This can be accomplished using a `<small>` tag instead of a `<span>`, or by applying the "small" class which avoids the need for custom CSS.

mk08sh (commented):
👍

MikeLyttle (approved):

BrianMaki (approved):

PR #4563, opened 2022-12-23
LaboratoryService refactoring

ytqsl:
# Fixes or Implements [AB#14499](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14499)

[Task 14499](https://dev.azure.com/qslvic/Health%20Gateway/_workitems/edit/14499): LaboratoryService refactoring

## Description

Refactored LaboratoryService validations

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

None

## Items to Review:


rayking29 (dismissed):

BrianMaki (dismissed):

Apps/Laboratory/src/Services/LaboratoryService.cs
```diff
@@ -86,224 +86,163 @@ public class LaboratoryService : ILaboratoryService
...
-                            loadState.BackOffMilliseconds,
-                            this.labConfig.BackOffMilliseconds);
-                    }
-                }
+                        RetryIn = Math.Max(loadState.BackOffMilliseconds, this.labConfig.BackOffMilliseconds),
+                        Loaded = !loadState.RefreshInProgress,
+                    },
+                    ActionType.Refresh,
+                    "Refresh in progress");
             }
```
MikeLyttle, line 108: The Loaded property also needs to be set when the request is successful.

BrianMaki (dismissed):

Apps/Laboratory/src/Services/LaboratoryService.cs
```diff
@@ -86,224 +86,163 @@ public class LaboratoryService : ILaboratoryService
...
```
MikeLyttle, line 147: The Loaded property also needs to be set when the request is successful.

The Queued property is no longer being populated at all.

Apps/Laboratory/src/Services/LaboratoryService.cs
```diff
@@ -86,224 +86,163 @@ public class LaboratoryService : ILaboratoryService
...
-                    ResultMessage = "Error parsing phn",
-                    ErrorCode = ErrorTranslator.InternalError(ErrorType.InvalidState),
-                };
-                return retVal;
+                return RequestResultFactory.Error<PublicCovidTestResponse>(ErrorType.InvalidState, "Error parsing phn");
             }
 
             string? accessToken = this.authenticationDelegate.AuthenticateAsSystem(this.tokenUri, this.tokenRequest).AccessToken;
-            if (string.IsNullOrEmpty(accessToken))
+            if (accessToken == null)
```
MikeLyttle, line 195: For safety, I'd recommend keeping the check as-is to preserve the same behaviour in case the value is an empty string.

Apps/Laboratory/src/Services/LaboratoryService.cs
```diff
@@ -86,224 +86,163 @@ public class LaboratoryService : ILaboratoryService
...
             {
-                this.logger.LogCritical(IsNullOrEmptyTokenErrorMessage);
```
MikeLyttle, line 254: Why was this log statement removed?

Apps/Laboratory/src/Services/LaboratoryService.cs
```diff
@@ -86,224 +86,163 @@ public class LaboratoryService : ILaboratoryService
...
+                    {
+                        RetryIn = Math.Max(loadState.BackOffMilliseconds, this.labConfig.BackOffMilliseconds),
+                        Loaded = !loadState.RefreshInProgress,
+                    },
+                    ActionType.Refresh,
+                    "Refresh in progress");
             }
 
-            if (loadState != null)
+            LabIndicatorType labIndicatorType = Enum.Parse<LabIndicatorType>(result.ResourcePayload?.Result.FirstOrDefault()?.StatusIndicator);
```
MikeLyttle, line 216: `Enum.Parse` will throw an exception if it's passed a null value, which will happen if there are no results.

MikeLyttle (changes requested):

Apps/Laboratory/src/Services/LaboratoryService.cs
ytqsl: Doesn't line 104 set the this property?

Apps/Laboratory/src/Services/LaboratoryService.cs
ytqsl: I made the check consistent, see line 168. I'd argue this should actually throw an exception from the authentication delegate as this is a very exceptional state.

Apps/Laboratory/src/Services/LaboratoryService.cs
ytqsl: made it consistent with the other checks in the other methods

Apps/Laboratory/src/Services/LaboratoryService.cs
ytqsl: So would the original -  Enum.Parse<LabIndicatorType>(payload.Select(x => x.StatusIndicator).First());
I'll make it null safe, though we need to revisit our null payload approach.

Apps/Laboratory/src/Services/LaboratoryService.cs
MikeLyttle: This still needs to be fixed.

MikeLyttle (changes requested):
One of the issues remains unresolved.

MikeLyttle (dismissed):

MikeLyttle (approved):

rayking29 (approved):

PR #4571, opened 2022-12-28
RestMedicationService refactoring

ytqsl:
# Fixes or Implements [AB#14575](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14575)

[Task 14575](https://dev.azure.com/qslvic/Health%20Gateway/_workitems/edit/14575): RestMedicationStatementService refactoring

## Description

Refactored RestMedicationService and RestMedicationStatementService 

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

None

## Items to Review:


Apps/Medication/src/Services/RestMedicationService.cs
```diff
@@ -45,52 +45,47 @@ public class RestMedicationService : IMedicationService
         /// <inheritdoc/>
         public IDictionary<string, MedicationInformation> GetMedications(IList<string> medicationDinList)
         {
-            this.logger.LogTrace("Getting list of medications...");
-            IDictionary<string, MedicationInformation> result = new Dictionary<string, MedicationInformation>();
+            var drugProducts = this.drugLookupDelegate.GetDrugProductsByDin(medicationDinList);
```
BrianMaki, line 48: Please confirm with team regarding the use of 'var'.  @MikeLyttle

Apps/Medication/src/Services/RestMedicationService.cs
```diff
@@ -45,52 +45,47 @@ public class RestMedicationService : IMedicationService
...
 
-            // Retrieve drug information from the Federal soruce
-            IList<DrugProduct> drugProducts = this.drugLookupDelegate.GetDrugProductsByDin(medicationDinList);
-            foreach (DrugProduct drugProduct in drugProducts)
+            var drugs = drugProducts.ToDictionary(m => m.DrugIdentificationNumber, m => new MedicationInformation
```
BrianMaki, line 50: Please confirm with team regarding the use of 'var'.  @MikeLyttle

Apps/Medication/src/Services/RestMedicationService.cs
```diff
@@ -45,52 +45,47 @@ public class RestMedicationService : IMedicationService
...
-            // Retrieve drug information from the Provincial source and append it to the result if previously added.
-            IList<PharmaCareDrug> pharmaCareDrugs = this.drugLookupDelegate.GetPharmaCareDrugsByDin(medicationDinList);
-            foreach (PharmaCareDrug pharmaCareDrug in pharmaCareDrugs)
-            {
-                ProvincialDrugSource provincialData = new()
-                {
-                    UpdateDateTime = pharmaCareDrug.UpdatedDateTime,
-                    PharmaCareDrug = pharmaCareDrug,
-                };
+            var pharmaCareDrugs = this.drugLookupDelegate.GetPharmaCareDrugsByDin(medicationDinList);
```
BrianMaki, line 60: Please confirm with team regarding the use of 'var'.  @MikeLyttle

Apps/Medication/src/Services/RestMedicationStatementService.cs
```diff
@@ -86,91 +82,45 @@ public async Task<RequestResult<IList<MedicationStatementHistory>>> GetMedicatio
...
-                        }
-                    }
-                    else
-                    {
-                        result.ResultError = patientResult.ResultError;
-                    }
-                }
-                else
+
+                var protectiveWordValidation = new ProtectiveWordValidator().Validate(protectiveWord);
```
BrianMaki, line 87: Please confirm with team regarding the use of 'var'.

Apps/Medication/src/Services/RestMedicationStatementService.cs
```diff
@@ -86,91 +82,45 @@ public async Task<RequestResult<IList<MedicationStatementHistory>>> GetMedicatio
...
+                if (response.ResultStatus != ResultType.Success || response.ResourcePayload == null)
                 {
-                    valid = false;
-                    errMsg = protectiveWord.Length > MaxLengthProtectiveWord ? ErrorMessages.ProtectiveWordTooLong : ErrorMessages.ProtectiveWordTooShort;
+                    return RequestResultFactory.Error<IList<MedicationStatementHistory>>(patientResult.ResultError);
                 }
-            }
 
-            return (valid, errMsg);
+                var payload = this.autoMapper.Map<IList<MedicationStatementHistory>>(response.ResourcePayload.Results);
```
BrianMaki, line 116: Please confirm with team regarding the use of 'var'.

BrianMaki (dismissed):

rayking29 (approved):

MikeLyttle (dismissed):

Apps/Medication/src/Services/RestMedicationStatementService.cs
```diff
@@ -86,91 +83,47 @@ public async Task<RequestResult<IList<MedicationStatementHistory>>> GetMedicatio
...
-            return (valid, errMsg);
+                IList<MedicationStatementHistory>? payload = this.autoMapper.Map<IList<MedicationStatementHistory>>(response.ResourcePayload.Results);
+                this.PopulateMedicationSummary(payload.Select(r => r.MedicationSummary).ToArray());
+
+                return RequestResultFactory.Success(payload, response.TotalResultCount, response.PageIndex, response.PageSize);
+            }
         }
 
-        private void PopulateMedicationSummary(List<MedicationSummary> medSummaries)
+        private void PopulateMedicationSummary(MedicationSummary[] medSummaries)
```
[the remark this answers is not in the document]
ytqsl: I agree re: IEnumerable - it should be the only list in public methods/dtos, with private it can be any, but preferable IEnumerable as well.
As for array vs list, arrays are immutable and mem contiguous so prob better than Lists IMO.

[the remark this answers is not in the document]
MikeLyttle: Under the hood, the List class uses an array to store its data, so performance is comparable.

Here's a quote from Eric Lippert's [Arrays Considered Somewhat Harmful](https://learn.microsoft.com/en-us/archive/blogs/ericlippert/arrays-considered-somewhat-harmful) article that expresses why I prefer Lists for internal implementations:

> And finally, given that arrays are mutable by design, the way an array restricts that mutability is deeply weird. All the contents of the collection are mutable, but the size is fixed.  What is up with that? Does that solve a problem anyone actually has?
> 
> For this reason alone I do almost no programming with arrays anymore. Arrays simply do not model any problem that I have at all well – I rarely need a collection which has the rather contradictory properties of being completely mutable, and at the same time, fixed in size. If I want to mutate a collection it is almost always to add something to it or remove something from it, not to change what value an index maps to.

MikeLyttle (approved):

Apps/Medication/src/Services/RestMedicationStatementService.cs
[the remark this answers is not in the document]
ytqsl: Fair enough, however, you're right that the instances in an array are mutable, but in most cases, a code that works on an array does not try to mutate the internals, but the sequence itself, and array do not have .Add() functionality.
In the spirit of functional programming, my personal preference is to end up with linq/side-effect-free functions that operate on IEnumerable.

BrianMaki (approved):

rayking29 (approved):

PR #4600, opened 2023-01-06
Refactor Problem Details for better code coverage for AB#14661.

BrianMaki:
# Fixes or Implements [AB#14661](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14661)

## Description

- Refactor Problem Details to use same implementation as Waiting Queue Web
- Updated affected unit tests
- Note: Suppressed warning on ProblemDetails due to build issue.

**Uncaught Exception:**

Screenshot 2023-01-06 at 12 24 34 PM

**Custom Exception:**

Screenshot 2023-01-06 at 12 38 15 PM

**Unit Tests:**

Screenshot 2023-01-06 at 12 48 27 PM

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:


Apps/CommonData/src/ErrorHandling/ExceptionUtility.cs
```diff
@@ -0,0 +1,46 @@
...
+// -------------------------------------------------------------------------
+namespace HealthGateway.Common.Data.ErrorHandling
+{
+    using System.Net;
+    using System.Runtime.CompilerServices;
+
+    /// <summary>
+    /// Utility for throwing exception.
+    /// </summary>
+    public static class ExceptionUtility
```
ytqsl, line 24: Should this be a factory?

Apps/CommonData/src/ErrorHandling/ProblemDetails.cs
```diff
@@ -13,25 +13,38 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 // -------------------------------------------------------------------------
-namespace HealthGateway.Common.ErrorHandling
+namespace HealthGateway.Common.Data.ErrorHandling
```
ytqsl, line 16: why move this to Common.Data?

Apps/CommonData/src/ErrorHandling/ProblemDetailsException.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace HealthGateway.Common.Data.ErrorHandling
+{
+    using System;
+    using System.Runtime.Serialization;
+
+    /// <summary>
+    /// Represents a custom api patient exception.
+    /// </summary>
+    [Serializable]
+    public class ProblemDetailsException : Exception
```
ytqsl, line 25: I don't think there should be an exception for this, the idea of problemDetail is to be able to map native exceptions or API specific states to a machine readable/actionable state.

Apps/Patient/src/Delegates/ClientRegistriesDelegate.cs
```diff
@@ -201,14 +201,14 @@ private void CheckResponseCode(string responseCode)
             {
                 // BCHCIM.GD.2.0018 Not found
                 this.logger.LogWarning("Client Registry did not find any records. Returned message code: {ResponseCode}", responseCode);
-                throw new ApiException(ErrorMessages.ClientRegistryRecordsNotFound, "ClientRegistriesDelegate.CheckResponseCode", HttpStatusCode.NotFound);
+                throw new ProblemDetailsException(ExceptionUtility.CreateProblemDetails(ErrorMessages.ClientRegistryRecordsNotFound, HttpStatusCode.NotFound, nameof(ClientRegistriesDelegate)));
```
ytqsl, line 204: Keeping the ApiException makes more sense to me, then map it in the filter to problemdetail. This leaks problemdetail concept all the way down.

ytqsl (approved):

ytqsl:
Added some comments re:usage of problemdetail and exceptions

Apps/CommonData/src/ErrorHandling/ExceptionUtility.cs
BrianMaki: @sslaws wrote this for WaitingQueueWeb.  Are you asking it to be renamed?

Apps/Patient/src/Delegates/ClientRegistriesDelegate.cs
BrianMaki: Trying to keep it consistent as it was called that over at WaitingQueueWeb.  Maybe we need a discussion with @sslaws

Apps/CommonData/src/ErrorHandling/ProblemDetails.cs
BrianMaki: ProblemDetails is a brand new POCO.  There was an ApiProblemDetails in HealthGateway.Common.ErrorHandling that has been removed as it is no longer needed for out of box botnet 7 problem details.

Apps/CommonData/src/ErrorHandling/ProblemDetailsException.cs
BrianMaki: This is used for custom thrown exceptions and is handled in the HttpResponseExceptionFilter.  It checks the exception thrown and if applicable, it maps the contents of the exception to Problem Details and adjusts the response code accordingly.

Apps/CommonData/src/ErrorHandling/ProblemDetails.cs
ytqsl: ProblemDetail existed in .net core since 2.1 AFAIK (https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.problemdetails?view=aspnetcore-2.1)
My question here is why put the ProblemDetail code in Common.Data assembly

Apps/CommonData/src/ErrorHandling/ProblemDetailsException.cs
ytqsl: Yes, I understand. My question was why not let the exceptions bubble to the filter and then map to ProblemDetail?

ytqsl (approved):
Some nitpicking comments :)

Apps/CommonData/src/ErrorHandling/ProblemDetails.cs
BrianMaki: ProblemDetails is there as it is used with ProblemDetailsException and ExceptionUtility, which are also there.  As for the name 'ProblemDetails', let's discuss with @sslaws as he created it. Do you suggest naming it something else?

Apps/CommonData/src/ErrorHandling/ProblemDetailsException.cs
BrianMaki: 1. Exceptions that bubble to the top and not intercepted in the filter, will be a 500
2. These are custom validation exceptions.  In client registry, they check for specific results in regards to codes sent back in the soap message. In the past they used the ResultError inside the RequestResult object.  Now a custom exception is thrown with  response code that can be specified like NOT FOUND, etc instead of relying on a 500 Internal Sever Error.

Apps/Patient/src/Delegates/ClientRegistriesDelegate.cs
MikeLyttle: I think we should discuss; as Yossi says, I don't think there should be references to ProblemDetails inside any of our services or delegates.

MikeLyttle (commented):
I don't understand why we removed the Hellang middleware; it seems more elegant and less confusing.  I think we should have a code camp / design session on exception handling so we can understand this better.

Apps/Patient/src/Delegates/ClientRegistriesDelegate.cs
BrianMaki: To be discussed in another story.

Apps/CommonData/src/ErrorHandling/ExceptionUtility.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="detail">The detail of the exception.</param>
+        /// <param name="statusCode">The http status code of the exception.</param>
+        /// <param name="typeName">The name of the type where the exception was generated.</param>
+        /// <param name="memberName">The member name where the exception occurred.</param>
+        /// <returns>An instance of cref="ProblemDetails".</returns>
+        public static ProblemDetails CreateProblemDetails(string detail, HttpStatusCode statusCode, string typeName, [CallerMemberName] string memberName = "")
+        {
+            return new ProblemDetails
+            {
+                ProblemType = "Waiting Queue Exception",
```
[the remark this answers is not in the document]
BrianMaki: As discussed with team reverting botnet 7 out of box implementation.

Apps/Common/src/AspNetConfiguration/Modules/ExceptionHandling.cs
```diff
@@ -36,36 +32,34 @@ public static class ExceptionHandling
...
+        {
+            app.UseExceptionHandler();
+            app.UseStatusCodePages();
 
-                    setup.MapToStatusCode<CommunicationException>(StatusCodes.Status502BadGateway);
-                });
+            if (environment.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
```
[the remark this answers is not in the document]
BrianMaki: As discussed with team reverting botnet 7 out of box implementation.

Apps/Patient/src/Program.cs
```diff
@@ -63,6 +62,8 @@ public static async Task Main(string[] args)
 
             Utility.ConfigureTracing(services, logger, configuration);
 
+            ExceptionHandling.ConfigureProblemDetails(services, environment);
+
```
MikeLyttle, line 66: Nit: this could be moved back to its original line for a commit that's less noisy.

MikeLyttle (approved):

Apps/Patient/src/Program.cs
BrianMaki: I will leave as is in case it is decided to use out of box dot net 7. If that were the case, it should bed placed at the end to ensure the proper problem details api writer is used.

BrianMaki:
> I don't understand why we removed the Hellang middleware; it seems more elegant and less confusing. I think we should have a code camp / design session on exception handling so we can understand this better.

It was moved to dot net 7 out of box as instructed.  It has since been reverted back due to team discussion.

PR #4620, opened 2023-01-11
Update inactive user report last login date time to PST for AB#14633

BrianMaki:
# Implements [AB#14633](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14633)

## Description

- Updated last login date time in inactive user report to display in Local Timezone instead of UTC via Auto Mapper - Note: local timezone is a parameter that can be turned on or off
- Added UtcLocal conversion method in DateFormatter
- Added unit for DateFormatter
- Moved GetLocalTimeZone from JobScheduler to Common.Data
- Local Timezone derived from OS
- Timezone Ids stored in configuration

**Unit Tests:**

Screenshot 2023-01-12 at 12 29 01 AM

**Report:**

Screenshot 2023-01-11 at 3 00 03 PM

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:

### DIFF
--- a/Apps/Admin/Server/Converters/UtcPstDateOutputConverter.cs
+++ b/Apps/Admin/Server/Converters/UtcPstDateOutputConverter.cs
@@ -1,0 +1,58 @@
+// -------------------------------------------------------------------------
+//  Copyright © 2019 Province of British Columbia
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------
+namespace HealthGateway.Admin.Server.Converters
+{
+    using System;
+    using System.Globalization;
+    using CsvHelper;
+    using CsvHelper.Configuration;
+    using CsvHelper.TypeConversion;
+    using HealthGateway.Common.Data.Utils;
+
+    /// <summary>
+    /// Converts UTC date to PST.
+    /// </summary>
+    public class UtcPstDateOutputConverter : ITypeConverter
+    {
+        /// <inheritdoc/>
+        public object? ConvertFromString(
+            string? text,
+            IReaderRow? row,
+            MemberMapData? memberMapData)
+        {
+            return text != null ? DateFormatter.ConvertDateTimeToPst(ParseDate(text)) : null;
+        }
+
+        /// <inheritdoc/>
+        public string ConvertToString(
+            object? value,
+            IWriterRow? row,
+            MemberMapData? memberMapData)
+        {
+            return value != null ? ConvertToPst((DateTime)value) : string.Empty;
+        }
+
+        private static string ConvertToPst(DateTime dateTime)
+        {
+            return DateFormatter.ToShortDateAndTime(DateFormatter.ConvertDateTimeToPst(dateTime));
+        }
+
+        private static DateTime ParseDate(string text)
+        {
+            return DateTime.Parse(text, CultureInfo.InvariantCulture, DateTimeStyles.None);
+        }
+    }
+}
--- a/Apps/Admin/Server/Services/InactiveUserService.cs
+++ b/Apps/Admin/Server/Services/InactiveUserService.cs
@@ -29,6 +29,7 @@ using HealthGateway.Common.AccessManagement.Authentication.Models;
 using HealthGateway.Common.Api;
 using HealthGateway.Common.Constants;
 using HealthGateway.Common.Data.Constants;
+using HealthGateway.Common.Data.Utils;
 using HealthGateway.Common.Data.ViewModels;
 using HealthGateway.Common.ErrorHandling;
 using HealthGateway.Database.Constants;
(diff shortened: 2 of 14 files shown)


Apps/CommonData/src/Utils/DateFormatter.cs
```diff
@@ -23,6 +23,8 @@ namespace HealthGateway.Common.Data.Utils;
 /// </summary>
 public static class DateFormatter
 {
+    private static readonly TimeZoneInfo PstZone = TimeZoneInfo.FindSystemTimeZoneById("Pacific Standard Time");
```
ytqsl, line 26: Unfortunately, this is not going to work in a linux container, you'll need to do something similar to [[bcgov/embc-ess-mod@master/shared/src/EMBC.Utilities/Extensions/DateTimeEx.cs](https://github.com/bcgov/embc-ess-mod/blob/master/shared/src/EMBC.Utilities/Extensions/DateTimeEx.cs)](https://github.com/bcgov/embc-ess-mod/blob/master/shared/src/EMBC.Utilities/Extensions/DateTimeEx.cs#L17)

ytqsl (changes requested):

Apps/CommonData/src/Utils/DateFormatter.cs
sslaws: We already have code that does this in JobScheduler and it can be moved into CommonData.  JobScheduler.Utils.ScheduleHelper GetLocalTimeZone

https://github.com/bcgov/healthgateway/blob/dev/Apps/JobScheduler/src/Utils/SchedulerHelper.cs

ytqsl (dismissed):

Apps/CommonData/src/Utils/DateFormatter.cs
BrianMaki: @sslaws moved GetLocalTimeZone to Common.Data.DateFormatter.cs.  Spoke with @ytqsl regarding hardcoding timezone ids instead of accessing via configuration as those values will most likely never change.

sslaws (changes requested):
Please switch back to being configurable.

Apps/Admin/Server/Converters/UtcPstDateOutputConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+    using System.Globalization;
+    using CsvHelper;
+    using CsvHelper.Configuration;
+    using CsvHelper.TypeConversion;
+    using HealthGateway.Common.Data.Utils;
+
+    /// <summary>
+    /// Converts UTC date to PST.
+    /// </summary>
+    public class UtcPstDateOutputConverter : ITypeConverter
```
MikeLyttle, line 28: We should use "Pacific Time" rather than "PST" in the comments and names of types and methods, because the code should be converting to PDT or PST as appropriate for the time of the year.

(I believe the code already works this way, even though the time zone we are looking up is labelled "Pacific Standard Time" in Windows.  The [TimeZoneInfo](https://learn.microsoft.com/en-us/dotnet/api/system.timezoneinfo?view=net-7.0) that is returned by [FindSystemTimeById](https://learn.microsoft.com/en-us/dotnet/api/system.timezoneinfo.findsystemtimezonebyid?view=net-7.0) knows about whether the time zone supports daylight savings time.)

This comment may become moot if we generalize the time zone functionality by making the time zones configurable.

BrianMaki: Spoke with @sslaws Moved GetLocalTimeZone from JobScheduler to Common.Data.  It also uses 'Pacific Standard Time' so not changing that.

BrianMaki:
> Please switch back to being configurable.

Switched back.

sslaws (dismissed):

Apps/CommonData/test/unit/CommonDataTests.csproj
```diff
@@ -15,6 +15,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="DeepEqual" Version="4.2.1" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="7.0.0" />
```
ytqsl, line 18: Can we see if we can add the abstractions package instead?

Apps/Admin/Server/Services/InactiveUserService.cs
```diff
@@ -100,7 +102,10 @@ public async Task<RequestResult<List<AdminUserProfileView>>> GetInactiveUsers(in
         // Compare inactive users in DB to users in Keycloak
         if (inactiveProfileResult.Status == DbStatusCode.Read && activeProfileResult.Status == DbStatusCode.Read)
         {
-            inactiveUsers.AddRange(this.autoMapper.Map<IEnumerable<AdminUserProfile>, IList<AdminUserProfileView>>(inactiveProfileResult.Payload));
+            inactiveUsers.AddRange(
+                inactiveProfileResult.Payload.Select(
+                        x => AdminUserProfileMapUtils.ToUiModel(x, this.configuration, true, this.autoMapper))
```
ytqsl, line 107: now the configuration must be injected here, another dependency to manage, I personally don't like this.

ytqsl (approved):
some minor comments

Apps/CommonData/test/unit/CommonDataTests.csproj
BrianMaki: @ytqsl That won't work.  I need ConfigurationBuilder, which is part of Microsoft.Extensions.Configuration and not Microsoft.Extensions.Configuration.Abstractions

Apps/Admin/Server/Services/InactiveUserService.cs
BrianMaki: Agreed. Changed.

sslaws (dismissed):

ytqsl (dismissed):

Apps/CommonData/src/Utils/DateFormatter.cs
```diff
@@ -121,21 +121,21 @@ public static bool TryParse(string dateTime, string format, out DateTime parsedD
...
         return TimeZoneInfo.ConvertTimeFromUtc(dateTime, timezone);
     }
 
     /// <summary>
-    /// Gets local timezone.
+    /// Gets local localTimezone.
     /// </summary>
     /// <param name="configuration">The configuration to use.</param>
-    /// <returns>TimeZoneInfo object representing local timezone.</returns>
+    /// <returns>TimeZoneInfo object representing local localTimezone.</returns>
```
MikeLyttle, line 127: Typos: "local localTimezone" => "local time zone".

ytqsl (approved):

ytqsl (approved):

MikeLyttle (approved):

PR #4626, opened 2023-01-13
Admin Agent Provisioning - create back end

rayking29:
# Implements [AB#14634](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14634)

## Admin Agent Provisioning - create back end

- new service 
- update IKeycloakAdminApi (refit API)

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:


Apps/Admin/Server/Services/AgentAccessService.cs
```diff
@@ -0,0 +1,212 @@
...
+            (this.tokenUri, this.tokenRequest) = this.authDelegate.GetClientCredentialsAuth(AuthConfigSectionName);
+        }
+
+        /// <inheritdoc/>
+        [SuppressMessage("Globalization", "CA1308:Normalize strings to uppercase", Justification = "KeyCloak requires lower case username")]
+        public async Task<AdminAgent> ProvisionAgentAccessAsync(AdminAgent agent)
+        {
+            JwtModel jwtModel = this.authDelegate.AuthenticateAsSystem(this.tokenUri, this.tokenRequest);
+
+            agent.Username = agent.Username.ToLowerInvariant();
```
BrianMaki, line 77: 1. Is this necessary? Doesn't keycloak automatically lowercase user?  If you go to kecyloak console and create a user with upper case or mixed case, it automatically lower cases it.  2.) If you are going to lowercase it here in the service, is it also being done on the front end for consistency?

rayking29: 1. It's not necessary for Keycloak.
2.  I have done this because Keycloak does not give back anything but a success when the user is created, so I am returning the AdminAgent object that was passed in. Which means I need to set the username to lower case to match what it actually looks like in Keycloak.

Apps/Admin/Server/Services/AgentAccessService.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            List<UserRepresentation> createdUser = await this.keycloakAdminApi.GetUserAsync(userName, jwtModel.AccessToken).ConfigureAwait(true);
+            await this.keycloakAdminApi.AddUserRolesAsync(createdUser.First().UserId.ToString(), roles, jwtModel.AccessToken).ConfigureAwait(true);
+
+            return agent;
+        }
+
+        /// <inheritdoc/>
+        public async Task<IEnumerable<AdminAgent>> GetAgentsAsync(string queryString, int? resultLimit = 25)
```
ytqsl, line 110: Looks like a security risk, does it make sense to pass user input directly from the controller to KC?

Apps/Admin/Server/Services/AgentAccessService.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            const int firstRecord = 0;
+
+            JwtModel jwtModel = this.authDelegate.AuthenticateAsSystem(this.tokenUri, this.tokenRequest);
+
+            List<UserRepresentation> users = await this.keycloakAdminApi.GetUsersQueryAsync(queryString, firstRecord, resultLimit.GetValueOrDefault(), jwtModel.AccessToken).ConfigureAwait(true);
+
+            List<AdminAgent> adminAgents = new();
+            foreach (UserRepresentation user in users)
```
ytqsl, line 119: How many results are we expecting here? If large, there may be a N+1 issue here; the loop can also leverage Parallel.ForEach or similar to load all user roles in parallel.

Apps/Admin/Server/Services/AgentAccessService.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                List<RoleRepresentation> userRoles = await this.keycloakAdminApi.GetUserRolesAsync(user.UserId.ToString(), jwtModel.AccessToken).ConfigureAwait(true);
+                string userName = user.Username.Split('@').First();
```
ytqsl, line 122: Duplicate split? Also a potential null/not found exception.

Apps/Admin/Server/Services/AgentAccessService.cs
sslaws: I think this is safe for a few reasons:
1) The application audience generally has equivalent admin access in Keycloak
2) The sanitation of the input should be handled by the service provider (Keycloak) as it is their consumable API.
3) What would make it safe?  We could strip out characters but which ones are invalid for Keycloak:
image

ytqsl: Ahh, I traced the actual Refit map, it's not a query string, it's a search param, then yes, not a risk. I'd recommend renaming queryString=>searchCriteria to make it clear this is not just forwarding a query string to the API call

Apps/Admin/Server/Services/AgentAccessService.cs
rayking29: I don't think we normally expect to get very many, but the limit is 25, which is the default. We tell Keycloak to only give back the max amount given. Does that answer the question?
I will look into the Parallel.ForEach

ytqsl: It does, then most likely this is good enough

Apps/Admin/Server/Services/AgentAccessService.cs
rayking29: Fixing

Apps/Admin/Server/Services/AgentAccessService.cs
rayking29: Removed the ToLowerInvariant and return a new AdminAgent instead since it needs the user id from Keycloak to be set.

Apps/Common/src/Api/IKeycloakAdminApi.cs
```diff
@@ -33,7 +33,7 @@ public interface IKeycloakAdminApi
         /// <param name="token">The bearer token to authorize the call.</param>
         /// <returns>A resulting UserRepresentation object.</returns>
         [Get("/users?briefRepresentation=true&username={username}&exact=true")]
-        Task<UserRepresentation> GetUserAsync(string username, [Authorize] string token);
+        Task<List<UserRepresentation>> GetUserAsync(string username, [Authorize] string token);
```
MikeLyttle, line 36: Would recommend renaming to `GetUsersByUsernameAsync` since it's returning a collection.

Apps/Common/src/Api/IKeycloakAdminApi.cs
```diff
@@ -46,6 +46,63 @@ public interface IKeycloakAdminApi
...
+
+        /// <summary>
+        /// Adds realm roles to user.
+        /// </summary>
+        /// <param name="id">The user's id.</param>
+        /// <param name="roles">List of RoleRepresentations to delete from the user.</param>
+        /// <param name="token">The bearer token to authorize the call.</param>
+        /// <returns>Returns true when roles deleted.</returns>
+        [Delete("/users/{id}/role-mappings/realm")]
+        Task DeleteUserRolesAsync(string id, [Body] IEnumerable<RoleRepresentation> roles, [Authorize] string token);
```
MikeLyttle, line 96: The `id` parameters in these declarations can be changed from `string` to `Guid` to save a lot of `ToString()` calls in the service.

Apps/Common/src/Api/IKeycloakAdminApi.cs
```diff
@@ -46,6 +46,63 @@ public interface IKeycloakAdminApi
...
+
+        /// <summary>
+        /// Get realm all roles.
+        /// </summary>
+        /// <param name="token">The bearer token to authorize the call.</param>
+        /// <returns>A RoleRepresentations of the roles.</returns>
```
MikeLyttle, line 102: Typo: should be "A List of RoleRepresentations...".

Apps/Common/src/Api/IKeycloakAdminApi.cs
```diff
@@ -46,6 +46,63 @@ public interface IKeycloakAdminApi
...
+        /// </summary>
+        /// <param name="id">The user's id.</param>
+        /// <param name="roles">List of RoleRepresentations to delete from the user.</param>
+        /// <param name="token">The bearer token to authorize the call.</param>
+        /// <returns>Returns true when roles deleted.</returns>
+        [Delete("/users/{id}/role-mappings/realm")]
+        Task DeleteUserRolesAsync(string id, [Body] IEnumerable<RoleRepresentation> roles, [Authorize] string token);
+
+        /// <summary>
+        /// Get realm all roles.
```
MikeLyttle, line 99: Typo: should be "Get all realm roles."

Apps/Common/src/Api/IKeycloakAdminApi.cs
```diff
@@ -46,6 +46,63 @@ public interface IKeycloakAdminApi
...
+        /// </summary>
+        /// <param name="id">The user's id.</param>
+        /// <param name="roles">List of RoleRepresentations to add to user.</param>
+        /// <param name="token">The bearer token to authorize the call.</param>
+        /// <returns>Returns true when roles added.</returns>
+        [Post("/users/{id}/role-mappings/realm")]
+        Task AddUserRolesAsync(string id, [Body] IEnumerable<RoleRepresentation> roles, [Authorize] string token);
+
+        /// <summary>
+        /// Adds realm roles to user.
```
MikeLyttle, line 89: Typo: should be "Removes realm roles from user."

MikeLyttle (approved):

MikeLyttle (approved):

Apps/Admin/Server/Services/AgentAccessService.cs
```diff
@@ -0,0 +1,199 @@
...
+
+                this.logger.LogError(e, "Keycloak API call failed");
+                throw;
+            }
+
+            List<UserRepresentation> getUserResponse = await this.keycloakAdminApi.GetUsersByUsernameAsync(user.Username, jwtModel.AccessToken).ConfigureAwait(true);
+            UserRepresentation createdUser = getUserResponse.First();
+            await this.keycloakAdminApi.AddUserRolesAsync(createdUser.UserId.GetValueOrDefault(), roles, jwtModel.AccessToken).ConfigureAwait(true);
+
+            string[] splitString = createdUser.Username.Split('@');
```
ytqsl, line 105: Is it worth asserting there are 2 elements post split?

Apps/Admin/Server/Services/AgentAccessService.cs
ytqsl: I still think it is worth renaming queryString=>searchString

Apps/Common/src/Api/IKeycloakAdminApi.cs
```diff
@@ -44,7 +44,64 @@ public interface IKeycloakAdminApi
...
+        /// <summary>
+        /// Look up User accounts by username, first or last name, or email.
+        /// </summary>
+        /// <param name="searchCriteria">The text string to search by.</param>
+        /// <param name="first">The first record to return.</param>
+        /// <param name="max">The maximum results to return.</param>
+        /// <param name="token">The bearer token to authorize the call.</param>
+        /// <returns>A List of UserRepresentation objects.</returns>
+        [Get("/users?briefRepresentation=true&first={first}&max={max}&search={searchCriteria}")]
+        Task<List<UserRepresentation>> GetUsersSearchAsync(string searchCriteria, int first, int max, [Authorize] string token);
```
ytqsl, line 58: Maybe rename to GetUsers or SearchUsers?

ytqsl (approved):

PR #4627, opened 2023-01-13
AB#14701 Cache refactoring

ytqsl:
# Fixes or Implements [AB#14701](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14701)

[Task 14701](https://dev.azure.com/qslvic/Health%20Gateway/_workitems/edit/14701): Refactor ICacheProvider and DI registration

## Description

Refactored the existing memory and redis cache providers with a distributed cache provider

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:


Apps/Common/test/unit/Services/CommunicationServiceTests.cs
```diff
@@ -240,52 +238,45 @@ public void ShouldProcessChangeEvent(string action, Scenario scenario, bool cach
...
-                        Assert.NotEqual(communication.Id, cacheResult?.ResourcePayload?.Id);
+                        Assert.NotEqual(communication.Id, cacheResult!.ResourcePayload!.Id);
                     }
 
                     break;
 
                 case Scenario.Expired:
-                    Assert.Equal(ResultType.Success, cacheResult?.ResultStatus);
-                    Assert.Equal(0, cacheResult?.TotalResultCount);
+                    Assert.Equal(ResultType.Success, cacheResult!.ResultStatus);
```
sslaws, line 268: The null conditional is a more appropriate than null suppression as cacheResult is nullable.  This applies to additional Asserts.

sslaws (changes requested):

Apps/Common/test/unit/Services/CommunicationServiceTests.cs
ytqsl: I'd agree from a consuming client point of view, but I'd argue that unit tests must be deterministic - there shouldn't be 'if/maybe/perhaps' in assertions as it would make them unreliable.

sslaws: Fair enough and if the unit test blows up things will still fail so no harm :)

BrianMaki (approved):

Apps/Common/test/unit/Services/CommunicationServiceTests.cs
ytqsl: Exactly, if it blows up, then the arrange is incorrect of the logic is incorrect

PR #4657, opened 2023-01-23
Backend changes for AB#14643

ytqsl:
# Fixes or Implements [AB#14735](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14735) and [AB#14737](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14737)

## Description

Backend work to enable dependent admin queries

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:

### DIFF
--- a/Apps/Common/src/MapProfiles/UserProfileProfile.cs
+++ b/Apps/Common/src/MapProfiles/UserProfileProfile.cs
@@ -19,6 +19,7 @@ namespace HealthGateway.Common.MapProfiles
     using AutoMapper;
     using HealthGateway.Common.Data.Models;
     using HealthGateway.Common.Data.ViewModels;
+    using HealthGateway.Database.Models;
 
     /// <summary>
     /// An AutoMapper profile class which defines mapping between DB Model UserProfile and API Model UserProfileModel.
@@ -39,6 +40,11 @@ namespace HealthGateway.Common.MapProfiles
             this.CreateMap<UserProfile, SupportUser>()
                 .ForMember(dest => dest.Hdid, opt => opt.MapFrom(src => src.HdId))
                 .ReverseMap();
+
+            this.CreateMap<ResourceDelegate, SupportUser>()
+                .ForMember(d => d.Hdid, opts => opts.MapFrom(s => s.ResourceOwnerHdid))
+                .ForMember(d => d.PersonalHealthNumber, opts => opts.Ignore())
+                .ForMember(d => d.LastLoginDateTime, opts => opts.Ignore());
         }
     }
 }
--- a/Apps/Common/src/MapProfiles/UserProfileProfile.cs
+++ b/Apps/Common/src/MapProfiles/UserProfileProfile.cs
@@ -42,7 +42,7 @@ namespace HealthGateway.Common.MapProfiles
                 .ReverseMap();
 
             this.CreateMap<ResourceDelegate, SupportUser>()
-                .ForMember(d => d.Hdid, opts => opts.MapFrom(s => s.ResourceOwnerHdid))
+                .ForMember(d => d.Hdid, opts => opts.MapFrom(s => s.ProfileHdid))
                 .ForMember(d => d.PersonalHealthNumber, opts => opts.Ignore())
                 .ForMember(d => d.LastLoginDateTime, opts => opts.Ignore());
         }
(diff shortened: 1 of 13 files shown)


Apps/Common/src/Services/PatientService.cs
```diff
@@ -94,34 +95,91 @@ public async Task<RequestResult<string>> GetPatientPhn(string hdid)
...
 
+        /// <inheritdoc/>
+        public async Task<string> GetPatientHdid(string phn)
+        {
+            using Activity? activity = Source.StartActivity();
+            RequestResult<PatientModel> patientResult = await this.GetPatient(phn, PatientIdentifierType.Phn).ConfigureAwait(true);
+            if (patientResult.ResultStatus != ResultType.Success || patientResult.ResourcePayload == null)
+            {
+                throw new ProblemDetailsException(
+                    new ProblemDetails
```
BrianMaki, line 106: You can use ExceptionUtility.CreateProblemDetails instead of doing new ProblemDetails. Pass in 'nameof<Your Class>' for typeName.

BrianMaki (changes requested):

Apps/Common/src/Services/PatientService.cs
ytqsl: fixed

rayking29 (approved):

Apps/Common/src/MapProfiles/UserProfileProfile.cs
```diff
@@ -39,6 +40,11 @@ public UserProfileProfile()
             this.CreateMap<UserProfile, SupportUser>()
                 .ForMember(dest => dest.Hdid, opt => opt.MapFrom(src => src.HdId))
                 .ReverseMap();
+
+            this.CreateMap<ResourceDelegate, SupportUser>()
+                .ForMember(d => d.Hdid, opts => opts.MapFrom(s => s.ResourceOwnerHdid))
```
MikeLyttle, line 45: The ResourceOwnerHdid is for the dependent; the ProfileHdid is for the delegate.

Apps/Common/src/Services/IPatientService.cs
```diff
@@ -32,6 +32,13 @@ public interface IPatientService
         /// <returns>The patient phn.</returns>
         Task<RequestResult<string>> GetPatientPhn(string hdid);
 
+        /// <summary>
+        /// Gets the patient hdid by phn.
+        /// </summary>
+        /// <param name="phn">The patient phn.</param>
+        /// <returns>The patient hdid.</returns>
```
MikeLyttle, line 39: Nit: HDID and PHN should be capitalized in comments.

Apps/Database/src/Delegates/IResourceDelegateDelegate.cs
```diff
@@ -67,5 +68,33 @@ public interface IResourceDelegateDelegate
...
+        public string? ByDelegateHdid { get; set; }
+    }
+
+    public record ResourceDelegateQueryResult
+    {
+        /// <summary>
+        /// gets or sets the found items.
+        /// </summary>
+        public IEnumerable<ResourceDelegate> Items { get; set; } = Array.Empty<ResourceDelegate>();
     }
```
MikeLyttle, line 99: We generally only define one type per file.  I'm not sure if adding these records is all that beneficial in general; maybe we can discuss the pattern with the team.

Apps/CommonData/src/Utils/DateFormatter.cs
```diff
@@ -13,128 +13,131 @@
 //  See the License for the specific language governing permissions and
 //  limitations under the License.
 // -------------------------------------------------------------------------
-namespace HealthGateway.Common.Data.Utils;
```
MikeLyttle, line 16: We allow both styles of namespace definitions and I prefer the file-scoped namespace for new files, so I'd recommend leaving this as-is.  Without the extra noise, it will make it easier to tell if there are any other changes to this file in this PR.

MikeLyttle (changes requested):

MikeLyttle (approved):

BrianMaki (approved):

PR #4670, opened 2023-01-26
Add auth handler to enable direct access to ODR services AB#14689

MikeLyttle:
# Implements AB#14689

## Description

Adds an OdrAuthorizationHandler to attach a client certificate and basic authorization header to requests to the ODR service.

The client certificate and authorization header are only added to requests when the appropriate configuration settings are enabled, and they are disabled by default.  The URLs for the API calls will continue to use the prefix needed by the proxy server unless the client certificate configuration setting is enabled.  So, the Medication and Encounter services should continue to function the same as they currently do until the configuration settings are changed.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:


Apps/Medication/src/Api/IOdrApi.cs
```diff
@@ -29,15 +29,15 @@ public interface IOdrApi
         /// </summary>
         /// <param name="request">The request to query.</param>
         /// <returns>The MedicationHistory.</returns>
-        [Post("/odr/patientProfile")]
```
ytqsl, line 32: I think this should not be changed, otherwise it would break current proxy usage

Apps/Common/src/AspNetConfiguration/Modules/OdrConfiguration.cs
```diff
@@ -0,0 +1,80 @@
...
+    public static class OdrConfiguration
+    {
+        /// <summary>
+        /// Configures the Refit API for accessing ODR services and adds it to the services collection.
+        /// </summary>
+        /// <param name="services">The service collection provider.</param>
+        /// <param name="configuration">The configuration to use for values.</param>
+        /// <typeparam name="TApi">The Refit API to be configured.</typeparam>
+        public static void AddOdrRefitClient<TApi>(IServiceCollection services, IConfiguration configuration)
+            where TApi : class
```
ytqsl, line 42: why is this needed and why is the method generic? HG services should be able to communicate to ODR without changing anything but the url of the remote service and a little configuration?

I'd expect a different options class to modify the httpclient handler that refit will use when connecting to the remote endpoint

ytqsl (commented):
some comments, maybe worth a quick discussion

Apps/Medication/src/Api/IOdrApi.cs
MikeLyttle: When the Refit client is added, the appropriate prefix is added depending on the value of the ClientCertificate.Enabled configuration setting.

Apps/Common/src/AspNetConfiguration/Modules/OdrConfiguration.cs
MikeLyttle: This was a refactor to avoid duplicating code in Encounter and Medication.  It's generic because we split the ODR calls into separate Refit APIs.  We could easily merge the Refit APIs and move them to Common instead.

BrianMaki (approved):

Apps/Common/src/AspNetConfiguration/Modules/OdrConfiguration.cs
ytqsl: I think it is better. Even better, move ODR to its own assembly including the adapter/delegate so the usable unit is the assembly, not code.

ytqsl:
@MikeLyttle is this PR supposed to go to DEV or to a feature branch?

ytqsl (approved):

PR #4599, opened 2023-01-06
Update agent access controller to latest standard AB#14660

MikeLyttle:
# Implements [AB#14660](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14660)

## Description

- updates agent access controller and service interface to latest standard
- enables use of problem details in Admin.Server

## Testing

- [ ] Unit Tests Updated
- [ ] Functional Tests Updated
- [x] Not Required

## Items to Review:

-   [General PR Guidelines](https://github.com/bcgov/healthgateway/wiki/PRguidance)

PR #4673, opened 2023-01-26
Dependent Endpoint for PHSA - update delegate

rayking29:
# Implements AB#14738

## Dependent Endpoint for PHSA - update delegate

- Updated delegate
- refactored page and pageSize to defaults on controller and not nullable in service
- renamed pageNumber to page
- fixed paging bug on GetDependents(hdid, page, pageSize) method

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/GatewayApi/src/Controllers/PhsaController.cs
+++ b/Apps/GatewayApi/src/Controllers/PhsaController.cs
@@ -19,7 +19,6 @@ namespace HealthGateway.GatewayApi.Controllers
     using System;
     using System.Collections.Generic;
     using HealthGateway.Common.AccessManagement.Authorization.Policy;
-    using HealthGateway.Common.Data.ViewModels;
     using HealthGateway.GatewayApi.Models;
     using HealthGateway.GatewayApi.Services;
     using Microsoft.AspNetCore.Authorization;
@@ -32,7 +31,7 @@ namespace HealthGateway.GatewayApi.Controllers
     [ApiVersion("1.0")]
     [Route("[controller]")]
     [ApiController]
-    public class PhsaController
+    public class PhsaController : ControllerBase
     {
         private readonly IDependentService dependentService;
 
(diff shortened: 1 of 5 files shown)


Apps/GatewayApi/src/Controllers/PhsaController.cs
```diff
@@ -59,9 +58,9 @@ public PhsaController(IDependentService dependentService)
         [HttpGet]
         [Authorize(Policy = UserProfilePolicy.Read)]
         [Route("dependents/{hdid}")]
-        public RequestResult<IEnumerable<DependentModel>> GetAll(string hdid)
+        public ActionResult<IEnumerable<DependentModel>> GetAll(string hdid)
         {
-            return this.dependentService.GetDependents(hdid);
+            return this.Ok(this.dependentService.GetDependents(hdid));
         }
```
MikeLyttle, line 64: I had done [something similar](https://github.com/bcgov/healthgateway/pull/4599/commits/a071f02d1e6a72a3a258c42a5ff4133322f34221) for the AgentAccessController but reverted it once I learned it was not needed and it made it more difficult to determine the response type, both manually and in Swagger.  (I think you still have a RequestResult involved here but it is not complaining that the types don't match.)

BrianMaki (approved):

Apps/GatewayApi/src/Controllers/PhsaController.cs
ytqsl: I respectfully disagree :) IMO this is a better way to control response code, types and swagger documentation. @MikeLyttle let's discuss, I'd be interested to know why you think implicit return types are better.

ytqsl (approved):

Apps/GatewayApi/src/Controllers/PhsaController.cs
MikeLyttle: I linked the wrong part of my sentence.  The linked commit I referenced was the revert and it was initiated [at your request](https://github.com/bcgov/healthgateway/pull/4599#discussion_r1063804055).

MikeLyttle: It doesn't check the type when returning ActionResult.

PR #4677, opened 2023-01-27
Dependent Endpoint for PHSA - unit tests

rayking29:
# Implements [AB#14741](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14741)

## Dependent Endpoint for PHSA - unit tests

- Added PhsaControllerTests.cs and new test to DependentServiceTests.cs
- Changed the PhsaController method returns to have the RequestResult included and only return the service method result so that I could create the PhsaController unit tests.  This matches the pattern in CommentController. Discuss.

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

Screenshot 2023-01-27 at 2 20 11 PM

Screenshot 2023-01-27 at 2 20 29 PM

## Items to Review:

### DIFF
--- a/Apps/GatewayApi/src/Controllers/PhsaController.cs
+++ b/Apps/GatewayApi/src/Controllers/PhsaController.cs
@@ -19,6 +19,7 @@ namespace HealthGateway.GatewayApi.Controllers
     using System;
     using System.Collections.Generic;
     using HealthGateway.Common.AccessManagement.Authorization.Policy;
+    using HealthGateway.Common.Data.ViewModels;
     using HealthGateway.GatewayApi.Models;
     using HealthGateway.GatewayApi.Services;
     using Microsoft.AspNetCore.Authorization;
@@ -58,9 +59,9 @@ namespace HealthGateway.GatewayApi.Controllers
         [HttpGet]
         [Authorize(Policy = UserProfilePolicy.Read)]
         [Route("dependents/{hdid}")]
-        public ActionResult<IEnumerable<DependentModel>> GetAll(string hdid)
+        public ActionResult<RequestResult<IEnumerable<DependentModel>>> GetAll(string hdid)
         {
-            return this.Ok(this.dependentService.GetDependents(hdid));
+            return this.dependentService.GetDependents(hdid);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Apps/GatewayApi/src/Controllers/PhsaController.cs
```diff
@@ -58,9 +59,9 @@ public PhsaController(IDependentService dependentService)
         [HttpGet]
         [Authorize(Policy = UserProfilePolicy.Read)]
         [Route("dependents/{hdid}")]
-        public ActionResult<IEnumerable<DependentModel>> GetAll(string hdid)
+        public ActionResult<RequestResult<IEnumerable<DependentModel>>> GetAll(string hdid)
```
ytqsl, line 62: NIT: I'm not sure why we insist using RequestResult, I thought we are trying to move away from it

Apps/GatewayApi/test/unit/Controllers.Test/PhsaControllerTests.cs
```diff
@@ -0,0 +1,129 @@
...
+    using HealthGateway.GatewayApi.Models;
+    using HealthGateway.GatewayApi.Services;
+    using Microsoft.AspNetCore.Mvc;
+    using Moq;
+    using Xunit;
+
+    /// <summary>
+    /// PhsaController's Unit Tests.
+    /// </summary>
+    public class PhsaControllerTests
```
ytqsl, line 33: NIT: Do we test controllers, or their underlying services?

ytqsl (approved):

Apps/GatewayApi/test/unit/Services.Test/DependentServiceTests.cs
```diff
@@ -74,6 +76,26 @@ public void GetDependents()
             }
         }
 
+        /// <summary>
+        /// GetDependents by date - Happy Path.
+        /// </summary>
+        [Fact]
+        public void GetDependentsByDate()
```
MikeLyttle, line 83: You'll get a few more lines of coverage if you include a test for the path that returns an error.

BrianMaki (approved):

MikeLyttle (approved):

Apps/GatewayApi/test/unit/Controllers.Test/PhsaControllerTests.cs
MikeLyttle: We often don't test (simple) controllers, but it looks like this is not uniform across all our projects.  We may wish to discuss in retro as I think we should do away with these types of tests.

rayking29: Yeah it really isn't testing much.

PR #4707, opened 2023-02-02
AdminWebClient Support Lookup Broken

rayking29:
# Fixes [AB#14841](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14841)

## AdminWebClient Support Lookup Broken

- added IResourceDelegateDelegate to Startup.cs for SupportService.cs changes
- added Dependent to UserQueryType.cs to match back end
- changed GetSupportUsers to an async task in the SupportController.cs to match change to GetUsers in SupportService.cs

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/AdminWebClient/src/Server/Startup.cs
+++ b/Apps/AdminWebClient/src/Server/Startup.cs
@@ -104,6 +104,7 @@ namespace HealthGateway.Admin
             services.AddTransient<IVaccineProofDelegate, VaccineProofDelegate>();
             services.AddTransient<IAdminUserProfileDelegate, DbAdminUserProfileDelegate>();
             services.AddTransient<IAuthenticationDelegate, AuthenticationDelegate>();
+            services.AddTransient<IResourceDelegateDelegate, DbResourceDelegateDelegate>();
 
             // Configure SPA
             services.AddControllersWithViews();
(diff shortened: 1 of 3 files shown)


Apps/AdminWebClient/src/Server/Startup.cs
```diff
@@ -161,7 +162,6 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                         endpoints.MapToVueCliProxy(
                             "{*path}",
                             new SpaOptions { SourcePath = "ClientApp" },
-                            "serve",
```
BrianMaki, line 164: What was this change for?

ytqsl (approved):

Apps/AdminWebClient/src/Server/Startup.cs
rayking29: I didn't realize that had happened until I was looking at this PR. I think it was a code cleanup thing. I looked at it and the default value for that item is "serve" so I think the code cleanup just removed it.

BrianMaki (approved):

PR #4713, opened 2023-02-04
Admin Support Last Login in PST

rayking29:
# Implements [AB#14872](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14872)

## Admin Support Last Login in PST

- added new map utils to convert to Pacific time using the configuration values
- convert to Pacific time in the SupportService using new map utils
- removed the conversion to local time from Support.vue in AdminWebClient since it is now done in the service

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

Screenshot 2023-02-07 at 1 46 12 PM

## UI Changes

**UserProfile table data:**
Screenshot 2023-02-03 at 5 10 36 PM

**MessagingVerfication table data:**
Screenshot 2023-02-03 at 5 15 20 PM

**Converted Times:**
Screenshot 2023-02-03 at 5 12 34 PM

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Components/MessageVerificationTable.razor
+++ b/Apps/Admin/Client/Components/MessageVerificationTable.razor
@@ -40,12 +40,10 @@
     <RowTemplate>
         <MudTd data-testid="messaging-verification-table-email-sms" DataLabel="Email/SMS">@context.EmailOrSms</MudTd>
         <MudTd data-testid="messaging-verification-verified" DataLabel="Verified">@context.Verified</MudTd>
-        <MudTd data-testid="messaging-verification-table-date" DataLabel="Verification Date">@DateFormatter.ToShortDateAndTime(@context.VerificationDate)</MudTd>
+        <MudTd data-testid="messaging-verification-table-date" DataLabel="Verification Date">@DateFormatter.ToShortDateAndTime(context.VerificationDate.ToLocalTime())</MudTd>
         <MudTd data-testid="messaging-verification-table-code" DataLabel="Verification Code">@context.VerificationCode</MudTd>
     </RowTemplate>
     <PagerContent>
         <MudTablePager />
     </PagerContent>
 </MudTable>
-
-
--- a/Apps/Admin/Client/Components/MessageVerificationTable.razor
+++ b/Apps/Admin/Client/Components/MessageVerificationTable.razor
@@ -40,10 +40,12 @@
     <RowTemplate>
         <MudTd data-testid="messaging-verification-table-email-sms" DataLabel="Email/SMS">@context.EmailOrSms</MudTd>
         <MudTd data-testid="messaging-verification-verified" DataLabel="Verified">@context.Verified</MudTd>
-        <MudTd data-testid="messaging-verification-table-date" DataLabel="Verification Date">@DateFormatter.ToShortDateAndTime(context.VerificationDate.ToLocalTime())</MudTd>
+        <MudTd data-testid="messaging-verification-table-date" DataLabel="Verification Date">@DateFormatter.ToShortDateAndTime(@context.VerificationDate)</MudTd>
         <MudTd data-testid="messaging-verification-table-code" DataLabel="Verification Code">@context.VerificationCode</MudTd>
     </RowTemplate>
     <PagerContent>
         <MudTablePager />
     </PagerContent>
 </MudTable>
+
+
--- a/Apps/Admin/Client/Pages/SupportPage.razor
+++ b/Apps/Admin/Client/Pages/SupportPage.razor
@@ -76,7 +76,7 @@
                         </MudTableSortLabel>
                     </MudTh>
                     <MudTh>
-                        <MudTableSortLabel SortBy="new Func<SupportUserRow, object>(x => x.LastLoginDateTime ?? DateTime.MinValue)"
+                        <MudTableSortLabel SortBy="new Func<SupportUserRow, object>(x => x.LastLoginDateTime)"
                                            InitialDirection="@SortDirection.Descending">
                             Last Login
                         </MudTableSortLabel>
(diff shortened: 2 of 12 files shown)


Apps/Admin/Client/Pages/SupportPage.razor.cs
```diff
@@ -207,7 +207,7 @@ public SupportUserRow(ExtendedSupportUser model)
 
             public string PersonalHealthNumber { get; }
 
-            public DateTime? LastLoginDateTime { get; }
+            public DateTime LastLoginDateTime { get; }
```
ytqsl, line 210: I think this will break the dependent query as it doesn't return a login date for the delegates

Apps/CommonData/src/ViewModels/SupportUser.cs
```diff
@@ -35,6 +35,6 @@ public class SupportUser
         /// <summary>
         /// Gets or sets the user's last login date time.
         /// </summary>
-        public DateTime? LastLoginDateTime { get; set; }
```
ytqsl, line 38: IMO this should stay nullable and the view methods should handle null dates

ytqsl (commented):
Some comments about datetime?->datetime changes

Apps/Admin/Client/Components/MessageVerificationTable.razor
```diff
@@ -40,12 +40,10 @@
     <RowTemplate>
         <MudTd data-testid="messaging-verification-table-email-sms" DataLabel="Email/SMS">@context.EmailOrSms</MudTd>
         <MudTd data-testid="messaging-verification-verified" DataLabel="Verified">@context.Verified</MudTd>
-        <MudTd data-testid="messaging-verification-table-date" DataLabel="Verification Date">@DateFormatter.ToShortDateAndTime(@context.VerificationDate)</MudTd>
+        <MudTd data-testid="messaging-verification-table-date" DataLabel="Verification Date">@DateFormatter.ToShortDateAndTime(context.VerificationDate.ToLocalTime())</MudTd>
```
BrianMaki, line 43: Using ToLocalTime() is not going to work.  There is no guarantee, the server the code will be run is Pacific time.

The conversion to PST from UTC should be done in an auto mapper util via SupportService.GetMessageVerifications.  See AdminUserProfileMapUtils.

Apps/Admin/Client/Pages/SupportPage.razor
```diff
@@ -86,7 +86,9 @@
                 <RowTemplate>
                     <MudTd DataLabel="HDID" data-testid=@($"user-table-hdid-{context.Hdid}")>@context.Hdid</MudTd>
                     <MudTd DataLabel="PHN" data-testid=@($"user-table-phn-{context.Hdid}")>@context.PersonalHealthNumber</MudTd>
-                    <MudTd DataLabel="Last Login" data-testid=@($"user-table-last-login-{context.Hdid}")>@DateFormatter.ToShortDateAndTime(context.LastLoginDateTime)</MudTd>
+                    <MudTd DataLabel="Last Login" data-testid=@($"user-table-last-login-{context.Hdid}")>
+                        @DateFormatter.ToShortDateAndTime(context.LastLoginDateTime.ToLocalTime())
```
BrianMaki, line 90: LastLoginDateTIme has already been converted to PST from UTC in changes made in AdminUserProfileMapUtils so is it necessary to use LastLoginDateTIme.ToLocalTime().  Also there is no guarantee the server will be running in PST.

BrianMaki (changes requested):
I think the conversion to PST from UTC should be handled in an auto mapper util and we should not depend on ToLocalTime().  Please see comments.

https://learn.microsoft.com/en-us/dotnet/api/system.datetime.tolocaltime?view=net-7.0

Timezone is configurable.  See AdminUserProfileMapUtils.

Apps/Admin/Server/Models/AdminUserProfileView.cs
```diff
@@ -60,6 +60,6 @@ public class AdminUserProfileView
         /// <summary>
         /// Gets or sets the user's last login date time from DB AdminUserProfile.
         /// </summary>
-        public DateTime? LastLoginDateTime { get; set; }
+        public DateTime LastLoginDateTime { get; set; }
```
BrianMaki, line 63: Why is this file being changed? Is the Support Page using this view? I believe it’s being used by inactive users report?

Apps/Admin/Server/MapUtils/AdminUserProfileMapUtils.cs
```diff
@@ -41,9 +41,7 @@ public static AdminUserProfileView ToUiModel(AdminUserProfile adminUserProfile,
                 adminUserProfile,
                 opts =>
                     opts.AfterMap(
-                        (_, dest) => dest.LastLoginDateTime = dest.LastLoginDateTime != null
-                            ? TimeZoneInfo.ConvertTimeFromUtc((DateTime)dest.LastLoginDateTime, timezone)
-                            : dest.LastLoginDateTime));
+                        (_, dest) => dest.LastLoginDateTime = TimeZoneInfo.ConvertTimeFromUtc(dest.LastLoginDateTime, timezone)));
```
BrianMaki, line 44: Why is this file being changed? The support page is using SupportUser model.

Apps/Admin/Tests/Services/CsvExportServiceTests.cs
```diff
@@ -133,7 +133,6 @@ public void TestMapperAdminUserProfileViewFromUserRepresentation()
             {
                 AdminUserProfileId = null,
                 UserId = userRepresentation.UserId,
-                LastLoginDateTime = null,
```
BrianMaki, line 136: Why is this test being changed? This is not related to the support page. It is dealing with System Analytics; looks like keycloak to Admin user profile view.  Inactive users report.

BrianMaki (approved):

MikeLyttle (approved):

ytqsl (approved):

PR #4771, opened 2023-02-22
refactored quicklinks to dataset feature toggles

ytqsl:
# Fixes or Implements [AB#14948](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14948)

Task 14948: Refactor quick links

## Description

Quicklinks to use new feature toggles config

image

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/constants/entryType.ts
+++ b/Apps/WebClient/src/ClientApp/src/constants/entryType.ts
@@ -116,7 +116,7 @@ entryTypeMap.set(EntryType.MedicationRequest, {
     icon: "file-medical",
     component: "MedicationRequestComponent",
     eventName: "special_authority",
-    enabled: (config) => isDatasetEnabled(config, "Medication"),
+    enabled: (config) => isDatasetEnabled(config, "SpecialAuthority"),
 });
 
 entryTypeMap.set(EntryType.ClinicalDocument, {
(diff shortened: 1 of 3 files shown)


Apps/WebClient/src/ClientApp/src/constants/entryType.ts
```diff
@@ -97,6 +116,7 @@ entryTypeMap.set(EntryType.MedicationRequest, {
     icon: "file-medical",
     component: "MedicationRequestComponent",
     eventName: "special_authority",
+    enabled: (config) => isDatasetEnabled(config, "Medication"),
```
MikeLyttle, line 119: This should be passing the dataset named "SpecialAuthorityRequest".  Once we've finished getting rid of the modules, we can update the enum values so these overrides won't be needed and we can just pass the EntryType as-is.

MikeLyttle (changes requested):

Apps/WebClient/src/ClientApp/src/constants/entryType.ts
ytqsl: We have a dataset called `specialAuthority`, is this the one?

MikeLyttle: Yes, though I think it would be better if the dataset name included request.

MikeLyttle (approved):

BrianMaki (commented):
What about functional tests?

PR #4789, opened 2023-02-27
Add new Services page skeleton AB#15008

BronzBierd:
# Fixes or Implements [AB#15008](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15008)

## Description

Add new services page, 
skeleton page 
with routing and toggle

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes
New services view
New sidebar nav link for services with hand holding medical icon

image

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
@@ -350,6 +350,7 @@ export default class SidebarComponent extends Vue {
                             </b-col>
                             <b-col
                                 v-show="isFullyOpen"
+                                data-testid="servicesLabel"
                                 class="button-text pl-3"
                             >
                                 <span>Services</span>
--- a/Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
@@ -26,10 +26,10 @@ library.add(
     faAngleDoubleLeft,
     faCheckCircle,
     faClipboardList,
+    faHandHoldingMedical,
     faHome,
     faStream,
-    faUserFriends,
-    faHandHoldingMedical
+    faUserFriends
 );
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -334,7 +334,7 @@ export default class SidebarComponent extends Vue {
                     <hg-button
                         v-show="isServicesEnabled && userIsActive"
                         id="menuBtnServices"
-                        data-testid="menu-btn-Services-link"
+                        data-testid="menu-btn-services-link"
                         to="/services"
                         variant="nav"
                         class="my-3 px-3 px-md-4"
--- a/Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
@@ -4,6 +4,7 @@ import {
     faAngleDoubleLeft,
     faCheckCircle,
     faClipboardList,
+    faHandHoldingMedical,
     faHome,
     faStream,
     faUserFriends,
@@ -27,7 +28,8 @@ library.add(
     faClipboardList,
     faHome,
     faStream,
-    faUserFriends
+    faUserFriends,
+    faHandHoldingMedical
 );
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -184,6 +186,10 @@ export default class SidebarComponent extends Vue {
         return this.$route.path == "/covid19";
     }
 
+    private get isServices(): boolean {
+        return this.$route.path == "/services";
+    }
+
     private get isTermsOfService(): boolean {
         return this.$route.path == "/profile/termsOfService";
     }
@@ -198,6 +204,10 @@ export default class SidebarComponent extends Vue {
 
     private get isDependentEnabled(): boolean {
         return this.config.featureToggleConfiguration.dependents.enabled;
+    }
+
+    private get isServicesEnabled(): boolean {
+        return this.config.featureToggleConfiguration.services.enabled;
     }
 
     private get isDependents(): boolean {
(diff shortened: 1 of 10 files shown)


Apps/WebClient/src/ClientApp/src/components/navmenu/SidebarComponent.vue
```diff
@@ -320,6 +330,33 @@ export default class SidebarComponent extends Vue {
                             </b-col>
                         </b-row>
                     </hg-button>
+                    <!-- Services button -->
+                    <hg-button
+                        v-show="isServicesEnabled && userIsActive"
+                        id="menuBtnServices"
+                        data-testid="menu-btn-Services-link"
```
MikeLyttle, line 337: Nit: recommend lowercasing to "menu-btn-services-link".

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
```diff
@@ -0,0 +1,46 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const options: any = {
+    components: {
+        BreadcrumbComponent,
+        LoadingComponent,
+    },
+};
+@Component(options)
+export default class ServicesView extends Vue {
+    private isLoading = true;
```
MikeLyttle, line 18: Recommend making isLoading into a getter since that will allow easy updating if it needs to check multiple properties in the future:

```
get isLoading(): boolean {
    return false;
}
```

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
```diff
@@ -0,0 +1,46 @@
...
+    private breadcrumbItems: BreadcrumbItem[] = [
+        {
+            text: "Services",
+            to: "/services",
+            active: true,
+            dataTestId: "breadcrumb-services",
+        },
+    ];
```
MikeLyttle, line 26: The `private` modifier on getters and properties in Vue components should be removed as we add new files and update old ones, to prevent Volar reporting problems of the form:

`Property 'breadcrumbItems' is private and only accessible within class 'ServicesView'`

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
```diff
@@ -0,0 +1,46 @@
...
+        this.isLoading = false;
+    }
+}
+</script>
+
+<template>
+    <div>
+        <BreadcrumbComponent :items="breadcrumbItems" />
+        <LoadingComponent :is-loading="isLoading" />
+        <page-title title="Services"></page-title>
```
MikeLyttle, line 38: Can be simplified:

`<page-title title="Services" />`

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
```diff
@@ -0,0 +1,46 @@
...
+        <h5 class="my-3">
+            You can check and update your Organ Donor Registry information here.
+            More health services will be added in future.
+        </h5>
+    </div>
+</template>
+
+<style scoped></style>
```
MikeLyttle, line 46: The `<style>` tag can be removed if it's not needed.

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
```diff
@@ -0,0 +1,46 @@
...
+
+<template>
+    <div>
+        <BreadcrumbComponent :items="breadcrumbItems" />
+        <LoadingComponent :is-loading="isLoading" />
+        <page-title title="Services"></page-title>
+        <h5 class="my-3">
+            You can check and update your Organ Donor Registry information here.
+            More health services will be added in future.
+        </h5>
```
MikeLyttle, line 42: I'd suggest making this a `<p>`, but that's up to UX and the product owner.

Testing/functional/tests/cypress/integration/ui/services/menu.js
```diff
@@ -0,0 +1,68 @@
...
+function login(isMobile) {
+    if (isMobile) {
+        cy.viewport("iphone-6"); // Set viewport to 375px x 667px
+    }
+    cy.login(
+        Cypress.env("keycloak.username"),
+        Cypress.env("keycloak.password"),
+        AuthMethod.KeyCloak
+    );
+    cy.checkTimelineHasLoaded();
```
MikeLyttle, line 20: You can simplify this by removing the call to `checkTimelineHasLoaded` and specifying the 4th parameter of .login() to load a different page instead of the timeline, e.g. "/home".

Testing/functional/tests/cypress/integration/ui/services/menu.js
```diff
@@ -0,0 +1,68 @@
...
+        toggleServices(false);
+    });
+    it("Side bar does not contain services nav link when services is disabled", () => {
+        cy.configureSettings({
+            services: {
+                enabled: false,
+            },
+        });
+        login(false);
+        cy.get("[data-testid=menu-btn-Services-link]").should("not.be.visible");
```
MikeLyttle, line 65: The "not.exist" assertion should be used when the element should not be present in the DOM.

However, it's also important to note that assertions for non-existence near the start of tests will happily return true and march onward if they execute before everything is loaded.  That can be prevented by first checking that a reliable element is visible.  Cypress will wait for that element to appear (or for the timeout value to elapse).

Testing/functional/tests/cypress/integration/ui/services/view.js
```diff
@@ -0,0 +1,30 @@
...
+    });
+
+    afterEach(() => {
+        Cypress.session.clearAllSavedSessions();
+    });
+
+    // Minimum to test skeleton of services page
+    it("View should show title and header", () => {
+        cy.get("h1#subject").should("have.text", "Services");
+        cy.get("h5.my-3").should("be.visible");
```
MikeLyttle, line 27: We don't usually check the existence or content of static page elements in the functional tests.

MikeLyttle (changes requested):
I left feedback for some nits and requests for small changes.

Since the PR includes new UI changes, the description should include screenshots (unless it's already been demoed).

Testing/functional/tests/cypress/integration/ui/services/view.js
BronzBierd: This was just for skeleton page's UI test, I went along with the joke in the morning meeting.

There isn't anything to test on the page but to ensure that the requirements of the task were facilitated.

Should I remove this?

MikeLyttle: I'd recommend removing them; the selector for the second one is very brittle, so if we did want to check that element, we would assign it a data-testid.

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
BronzBierd: I followed your implementation from the dependents view:

https://github.com/bcgov/healthgateway/blob/dev/Apps/WebClient/src/ClientApp/src/views/DependentsView.vue#L162

Would you like me to change them both?

Testing/functional/tests/cypress/integration/ui/services/menu.js
BronzBierd: I've shuffled a few elements around to check for more reliable test structures.

The disabling of features doesn't remove the element from the DOM but set opacity to 0. So I need to stick with `not.be.visible`

I appreciate the insight in the loading caveats though haven't done testing like this in a long time so this will come in handy.

MikeLyttle: Good catch; I didn't notice it was a `v-show` rather than a `v-if`.

MikeLyttle (approved):

Apps/WebClient/src/ClientApp/src/views/ServicesView.vue
BronzBierd: Hi, @MikeLyttle,

I followed up with @mk08sh, regarding the styling etc. he's not worried about the current selection of styling as the text will be removed when more than one service is loaded. So I'm going to leave as is.

mk08sh (commented):
👍

rayking29 (approved):

ytqsl (approved):

PR #4824, opened 2023-03-07
Dependent DOB Picker - Future

rayking29:
# Fixes [AB#15049](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15049)

## Dependent DOB Picker - Future

- changed date picker to have max date for add dependent only, so user can't pick future date
- added message for future dated DoB when typed in to show 'Invalid date'
- added validator for server DoB future date check
- added unit test for new validator

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

Screenshot 2023-03-06 at 4 12 13 PM

## UI Changes

Screenshot 2023-03-06 at 4 52 38 PM

Screenshot 2023-03-06 at 4 53 26 PM

Screenshot 2023-03-06 at 4 53 57 PM

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
@@ -21,18 +21,14 @@ library.add(faCalendar);
 export default class DatePickerComponent extends Vue {
     @Model("change", { type: String }) public model!: string;
     @Prop({ default: undefined }) state?: boolean;
-    @Prop({ default: undefined }) maxPickerDate?: Date;
+    @Prop({ default: new DateWrapper("2100-01-01") }) maxDate!: DateWrapper;
     @Ref("datePicker") datePicker!: BFormDatepicker;
 
     private value = "";
     private inputValue = "";
-    private maxDate = new Date("2100-01-01");
 
     private mounted(): void {
         this.value = this.model;
-        if (this.maxPickerDate) {
-            this.maxDate = this.maxPickerDate;
-        }
     }
 
     private onFocus(): void {
@@ -101,9 +97,7 @@ export default class DatePickerComponent extends Vue {
                     ),
                 maxValue: (value: string) =>
                     !value ||
-                    DateWrapper.fromStringFormat(value).isBefore(
-                        new DateWrapper("2100-01-01")
-                    ),
+                    DateWrapper.fromStringFormat(value).isBefore(this.maxDate),
             },
         };
     }
@@ -131,7 +125,7 @@ export default class DatePickerComponent extends Vue {
             <b-form-datepicker
                 ref="datePicker"
                 v-model="value"
-                :max="maxDate"
+                :max="maxDate.toJSDate()"
                 menu-class="datepicker-style"
                 button-only
                 right
--- a/Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
@@ -21,13 +21,18 @@ library.add(faCalendar);
 export default class DatePickerComponent extends Vue {
     @Model("change", { type: String }) public model!: string;
     @Prop({ default: undefined }) state?: boolean;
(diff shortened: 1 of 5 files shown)


Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
```diff
@@ -126,6 +131,7 @@ export default class DatePickerComponent extends Vue {
             <b-form-datepicker
                 ref="datePicker"
                 v-model="value"
+                :max="maxDate"
```
BrianMaki, line 134: I see you've added maxDate to :max in the b-form-datepicker.  

How does this interact with the current validation routine already in place from lines 94 to 113. Wouldn't the prop added on line 24 be used to override the max value defined on line 105?

Screenshot 2023-03-06 at 5 08 50 PM

Does this change have any affect on the date picker used in the timeline filter, report filter and add notes?

rayking29: That's a good point, maybe it should be using the value of maxDate on line 29 as the validator maxValue it is checking against. The maxDate is set to a default value that is the same as the maxValue in the validator, then in the mounted() function if there is a maxPickerDate passed in it sets maxDate to match.

There is no affect on other places that date picker is used since the maxPickerDate is nullable and therefore optional value and is only used if it has been set. So I didn't set the maxPickerDate anywhere else but in Add Dependent so it will continue to work the same as before.

Apps/WebClient/src/ClientApp/src/components/modal/NewDependentComponent.vue
```diff
@@ -250,12 +252,22 @@ export default class NewDependentComponent extends Vue {
...
+                                    <b-form-invalid-feedback
+                                        :state="
+                                            $v.dependent.dateOfBirth.$dirty
+                                                ? $v.dependent.dateOfBirth
+                                                      .maxValue
+                                                : undefined
+                                        "
+                                    >
+                                        Invalid date
+                                    </b-form-invalid-feedback>
```
MikeLyttle, line 270: If the `required` or `minLength` validations fail, there still needs to be feedback displayed for them.  I think they can be covered here under the designation of "Invalid date":

```
$v.dependent.dateOfBirth.$dirty
      ? $v.dependent.dateOfBirth.required || $v.dependent.dateOfBirth.minLength || $v.dependent.dateOfBirth.maxValue
      : undefined
```

Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
```diff
@@ -21,13 +21,18 @@ library.add(faCalendar);
...
 
     private value = "";
     private inputValue = "";
+    private maxDate = new Date("2100-01-01");
 
     private mounted(): void {
         this.value = this.model;
+        if (this.maxPickerDate) {
+            this.maxDate = this.maxPickerDate;
+        }
```
MikeLyttle, line 35: This could be simplified to:

```
@Prop({ default: new Date("2100-01-01") }) maxPickerDate: Date;
```

Nit: I'd suggest renaming the prop to `maxDate`.

mk08sh (commented):
👍

Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
rayking29: Fixed so that the maxDate is used for the validation.

Apps/WebClient/src/ClientApp/src/components/modal/NewDependentComponent.vue
rayking29: Fixed

Apps/WebClient/src/ClientApp/src/components/DatePickerComponent.vue
rayking29: Fixed

MikeLyttle (approved):

PR #4858, opened 2023-03-10
Add cypress config setup for services AB#15120

BronzBierd:
# Fixes or Implements [AB#15120](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15120)

## Description

1. Generalised the `configureSettings`'s  supporting functions.
  * Generalised function names
  * Added default arguments for key and value property names for merging an object array
2. Added configuration flow for services similar to datasets
3. Removed unnecessary clear sessions call in view.js for services page.

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes
image

## Notes
None

## Items to Review:

### DIFF
--- a/Testing/functional/tests/cypress/integration/ui/services/view.js
+++ b/Testing/functional/tests/cypress/integration/ui/services/view.js
@@ -1,25 +1,30 @@
+const { AuthMethod } = require("../../../support/constants");
+
 const servicesTestsConstants = {
     servicesUrl: "/services",
+    unauthorized: "/unauthorized",
 };
-const { AuthMethod } = require("../../../support/constants");
 
 describe("Authenticated Services View", () => {
     beforeEach(() => {
+        cy.login(
+            Cypress.env("keycloak.username"),
+            Cypress.env("keycloak.password"),
+            AuthMethod.KeyCloak,
+            "/services"
+        );
+    });
+
+    it("The url should be the services url if services enabled", () => {
         cy.configureSettings({
             services: {
                 enabled: true,
             },
         });
-        cy.login(
-            Cypress.env("keycloak.username"),
-            Cypress.env("keycloak.password"),
-            AuthMethod.KeyCloak,
-            "/home"
-        );
-        cy.visit(servicesTestsConstants.servicesUrl);
+        cy.url().should("include", servicesTestsConstants.servicesUrl);
     });
 
-    it("The page should load correctly", () => {
-        cy.url().should("include", servicesTestsConstants.servicesUrl);
+    it("The url should be the unauthorized url if services is disabled", () => {
+        cy.url().should("include", servicesTestsConstants.unauthorized);
     });
 });
--- a/Testing/functional/tests/cypress/integration/ui/services/view.js
+++ b/Testing/functional/tests/cypress/integration/ui/services/view.js
@@ -25,6 +25,7 @@ describe("Authenticated Services View", () => {
     });
 
     it("The url should be the unauthorized url if services is disabled", () => {
+        cy.configureSettings({});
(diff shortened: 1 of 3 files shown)


Testing/functional/tests/cypress/integration/ui/services/view.js
```diff
@@ -19,10 +19,6 @@ describe("Authenticated Services View", () => {
         cy.visit(servicesTestsConstants.servicesUrl);
```
BrianMaki, line 19: In cy.login you can use /services instead of /home?  Is that not possible?

Screenshot 2023-03-10 at 2 05 34 PM

Testing/functional/tests/cypress/integration/ui/services/view.js
```diff
@@ -19,10 +19,6 @@ describe("Authenticated Services View", () => {
...
     });
 
-    afterEach(() => {
-        Cypress.session.clearAllSavedSessions();
-    });
-
     it("The page should load correctly", () => {
```
BrianMaki, line 22: Should there be a test where services enabled = false?  What will the VUE router do when services enabled = false?

BrianMaki (changes requested):
I think one more functional test is required.  See comments.

BrianMaki (changes requested):
Have you confirmed ui/errorAlerts and ui/tooManyRequestsAlerts have passed?

Testing/functional/tests/cypress/integration/ui/services/view.js
BronzBierd: I'll add one it currently goes to the unauthorized page, which is the same behaviour as dependents.

Testing/functional/tests/cypress/integration/ui/services/view.js
BronzBierd: I'll add one it currently goes to the unauthorized page, which is the same behaviour as dependents.

BronzBierd: image

Added the test as suggested

BronzBierd:
@sslaws Please see comments above and below, regarding the need to now update the openshift dev's configuration to match the required structures. This is in line with the current implementation however the configuration structures could be improved further to make it more robust but the then the testing will not highlight missing configuration structures when disabling and patching the configuration.

cc: @BrianMaki 

Below is the context from teams and the argument that the config now requires updating:
>without the config change on openshift this is now expected to fail similar to the configureObject code.
>
>it highlights mandatory setting structure.
> 
>if you delete dependents.datasets you'll see the same error:
>image
>This time you can see what config is failing

BronzBierd:
@BrianMaki Running from a clean pull of the branch, all going through.

image

BrianMaki (changes requested):
I have updated configuration in openshift for dev. There is still one functional test error to fix:

- Authenticated Services View:

Screenshot 2023-03-10 at 3 43 45 PM

https://cloud.cypress.io/projects/ofnepc/runs/2773/overview?roarHideRunsWithDiffGroupsAndTags=1

BronzBierd:
@BrianMaki, working after the latest push local run with the expected structure

image

BronzBierd:
https://dev.azure.com/qslvic/Health%20Gateway/_build/results?buildId=31955&view=logs&j=9d906afe-3eb5-5d79-316d-86f3f3fa360c&t=345a5767-382b-5ba7-a23d-cff1ca5fc31a

Only the clinical documents test failed, so it looks to be in the clear.

BrianMaki (approved):

rayking29 (approved):

PR #4876, opened 2023-03-14
Create database tables and delegate for AB15116

BrianMaki:
# Implements [AB#15116](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15116)

## Description

- Created Dependent table with Hdid as PK
- Created AllowedDelegation table with DependentHdid and DelegateHdid as PK with DependentHdid as FK to Dependent table
- Created DelegationDelegate:

1. Get Dependents with option to retrieve AllowedDelegations
2. Insert Dependent
3. Update Dependent
4. Insert Allowed Delegation
5. Delete Allowed Delegation

**Migration:**

Screenshot 2023-03-14 at 3 09 18 PM

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:


Apps/Database/src/Delegates/DbDelegationDelegate.cs
```diff
@@ -0,0 +1,168 @@
...
+                    result.Status = DbStatusCode.Error;
+                    result.Message = e.Message;
+                }
+            }
+
+            return result;
+        }
+
+        /// <inheritdoc/>
+        public DbResult<IList<Dependent>> GetDependents(string hdid, bool includeAllowedDelegation = false)
```
BronzBierd, line 72: The `GetDependents` name is a little misleading. n the query below, you are getting a single dependent record and then including the delegations.

Would this not be better as just `GetDependent`

Apps/Database/src/Delegates/IDelegationDelegate.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Fetches the Dependents from the database.
+        /// </summary>
+        /// <param name="hdid">The dependent hdid to query on.</param>
+        /// <param name="includeAllowedDelegation">
+        /// Indicates whether allowed delegation should be included in the returned
+        /// dependent.
+        /// </param>
+        /// <returns>A DB result which encapsulates the return objects and status.</returns>
+        DbResult<IList<Dependent>> GetDependents(string hdid, bool includeAllowedDelegation = false);
```
BronzBierd, line 46: As per the previous comment you're retrieving a dependent and their allowed delegates. suggestion: `GetDependent`

BronzBierd (changes requested):
Only a naming suggestion on the method name for retrieving a dependent.

Apps/Database/src/Delegates/DbDelegationDelegate.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            return new()
+            {
+                Payload = this.dbContext.Dependent
+                    .Where(d => d.HdId == hdid)
+                    .OrderByDescending(d => d.CreatedDateTime)
+                    .ToList(),
+                Status = DbStatusCode.Read,
+            };
+        }
```
MikeLyttle, line 96: The query could be built bit-by-bit before executing to avoid repetition in the branching paths.

```
IQueryable<Dependent> query = this.dbContext.Dependent
    .Where(d => d.HdId == hdid);

if (includeAllowedDelegation)
{
    query = query.Include(d => d.AllowedDelegations);
}

return new()
{
    Payload = query.OrderByDescending(d => d.CreatedDateTime).ToList(),
    Status = DbStatusCode.Read,
};
```

Apps/Database/src/Delegates/DbDelegationDelegate.cs
BronzBierd: I think a better way to discuss this is that the dependent table's pk is the HDID thus its guaranteed to only be a single entry query.

Should this also not only be `DbResult<Dependent>`.

Apps/Database/src/Delegates/IDelegationDelegate.cs
MikeLyttle, line 1: I don't think the docs need to include information on what happens with the audit properties (e.g. UpdatedDateTime). 

 If we want to include the information about the concurrency exception, we should do that in the xmldoc via something like:

```<exception cref="ConcurrencyException">Version is not set.</exception>```

Apps/Database/src/Delegates/DbDelegationDelegate.cs
BrianMaki: Correct it should have been GetDependent not GetDependents.  Fixed.  Returns a single dependent.

Apps/Database/src/Delegates/DbDelegationDelegate.cs
BrianMaki: Modified as per suggestion.

Apps/Database/src/Delegates/IDelegationDelegate.cs
BrianMaki: Removed the information regarding UpdatedDateTime in docs.

MikeLyttle (approved):

rayking29 (approved):

PR #4877, opened 2023-03-14
Add patientData store AB#15060

BronzBierd:
# Fixes or Implements [AB#15060](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15060)

## Description

1. Add patientData store infrastructure
    * Models
    * Actions, getters and mutations
    * Adding to container
    * Initializing service
2. Add patientData retrieval code
3. Add patientData file retrieval code.

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

## UI Changes
None

## Notes
None

## Items to Review:


Apps/WebClient/src/ClientApp/src/services/interfaces.ts
```diff
@@ -43,6 +44,10 @@ import UserRating from "@/models/userRating";
 import VaccinationStatus from "@/models/vaccinationStatus";
 import { RootState } from "@/store/types";
 
+export interface IHttpDelegateService {
```
BronzBierd, line 47: This initialize signature is declared 21 times in this file without variation, hence the generalized interface. I would like to apply it to all of the interfaces, but can be done after delivery as this doesn't not directly impact the delivery of this work such as the PatientData refactor yesterday.

Att: @sslaws, @ytqsl

Apps/WebClient/src/ClientApp/src/store/modules/patientData/actions.ts
```diff
@@ -0,0 +1,150 @@
...
+                            reject(new Error("No patient data was returned"));
+                        } else {
+                            context.commit("setPatientDataFile", {
+                                fileId: params.fileId,
+                                file: data,
+                            });
+                            resolve(data);
+                        }
+                    })
+                    .catch((error: HttpError) => {
```
MikeLyttle, line 52: This error would actually be a ResultError. (Applies throughout file.)

All HttpErrors encountered by services should be converted to ResultError by ErrorTranslator on the way out.

As an aside, while it can be useful to have a type annotation here, it doesn't do as much as we'd want - we can't set up different catch blocks for different errors, for example.  All rejections go to the first catch block, regardless of any type mismatches.

Apps/WebClient/src/ClientApp/src/store/modules/patientData/types.ts
```diff
@@ -0,0 +1,98 @@
...
+    ): void;
+    setPatientDataError(
+        state: PatientDataState,
+        payload: { hdid: string; error: HttpError }
+    ): void;
+
+    setPatientDataFileError(
+        state: PatientDataState,
+        payload: { fileId: string; error: HttpError }
+    ): void;
```
MikeLyttle, line 89: Nit: inconsistent newlines.  Recommend removing to match other store interfaces.

Apps/WebClient/src/ClientApp/src/store/modules/patientData/types.ts
MikeLyttle, line 1: The references to HttpError in here should all be ResultError.

Apps/WebClient/src/ClientApp/src/store/modules/patientData/getters.ts
```diff
@@ -0,0 +1,27 @@
...
+    PatientDataState,
+} from "@/store/modules/patientData/types";
+import { getPatientDataRecordState } from "@/store/modules/patientData/utils";
+
+export const getters: PatientDataGetters = {
+    patientDataFiles(
+        state: PatientDataState
+    ): Dictionary<PatientDataFileState> {
+        return state.patientDataFiles;
+    },
```
MikeLyttle, line 16: I suspect it would be more useful to have getters similar to the ones for patientData/isPatientDataLoading that instead use the fileId to retrieve details of a particular file.

Apps/WebClient/src/ClientApp/src/store/modules/patientData/actions.ts
```diff
@@ -0,0 +1,150 @@
...
+                resolve(patientDataFile);
+            } else {
+                logger.debug("Retrieving patient data file");
+                context.commit("setPatientDataFileRequested", params.fileId);
+                patientDataService
+                    .getFile(params.hdid, params.fileId)
+                    .then((data) => {
+                        if (data === undefined) {
+                            reject(new Error("No patient data was returned"));
+                        } else {
```
MikeLyttle, line 44: This case can be handled with the other errors in the catch without any special treatment once the service is changed to reject requests when the appropriate configuration setting is not enabled. (Applies throughout file.)

MikeLyttle (changes requested):

Apps/WebClient/src/ClientApp/src/plugins/inversify.config.ts
```diff
@@ -21,7 +21,7 @@ import {
     ILaboratoryService,
     ILogger,
     IMedicationService,
-    INotificationService,
+    INotificationService, IPatientDataService,
```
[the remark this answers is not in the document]
BronzBierd: I rarely add these manually, not sure why the saving didn't auto format this correctly. While I'm making the changes I'll shuffle this.

Apps/WebClient/src/ClientApp/src/models/patientData.ts
```diff
@@ -0,0 +1,18 @@
+﻿export default class PatientData {
+    items: PatientHealthOptions[] | undefined;
```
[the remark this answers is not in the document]
BronzBierd: This shouldn't have been left in on the final submit, thanks for catching that.

Apps/WebClient/src/ClientApp/src/store/modules/patientData/types.ts
BronzBierd: The added whitespace with the IDE annotations messes with my perception, I'll remove this its an eyesore I agree.

Apps/WebClient/src/ClientApp/src/store/modules/patientData/getters.ts
BronzBierd: I was following clinical docs which just returns it all, but I guess that's for a very different screen, I'll amend.

MikeLyttle (approved):

BrianMaki (approved):

PR #4935, opened 2023-03-27
Dependent Exports - Implement export component in new tab

rayking29:
# Implements [AB#14965](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/14965)

## Dependent Exports - Implement export component in new tab
 - add Export Records tab to Dependents page
 - Record Type available list according to dependent data sets
 - advanced filter available
 - No Records Found same as Export Records
 - no preview data shown
 - all download types are available
 - Immunizations report is dependent version

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

Screenshot 2023-03-27 at 3 26 00 PM

Screenshot 2023-03-27 at 3 26 24 PM

Screenshot 2023-03-27 at 3 32 53 PM

Screenshot 2023-03-27 at 3 27 03 PM

Screenshot 2023-03-27 at 3 28 19 PM

Screenshot 2023-03-27 at 3 29 05 PM

Screenshot 2023-03-27 at 3 29 17 PM

Dependent Imms Report
Screenshot 2023-03-27 at 3 29 53 PM

User Imms Report from Export Records

Screenshot 2023-03-27 at 5 44 09 PM

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
@@ -261,19 +261,16 @@ export default class ImmunizationHistoryReportComponent extends Vue {
 
 <template>
     <div>
-        <div
-            v-if="isRecommendationEmpty && isEmpty && !isLoading"
-            class="mt-2 d-md-none"
-        >
+        <div v-if="isRecommendationEmpty && isEmpty && !isLoading">
             No records found.
         </div>
-        <section v-else-if="!isDependent">
-            <b-row class="d-none d-md-block">
+        <section v-else-if="!isDependent" class="d-none d-md-block">
+            <b-row>
                 <b-col>
                     <h4>Immunization History</h4>
                 </b-col>
             </b-row>
-            <b-row v-if="isEmpty && !isLoading" class="d-none d-md-block">
+            <b-row v-if="isEmpty && !isLoading">
                 <b-col>No records found.</b-col>
             </b-row>
             <b-table
@@ -283,7 +280,7 @@ export default class ImmunizationHistoryReportComponent extends Vue {
                 :items="immunizationItems"
                 :fields="immunizationFields"
                 data-testid="immunization-history-report-table"
-                class="table-style d-none d-md-table"
+                class="table-style"
             >
                 <!-- A custom formatted header cell for field 'name' -->
                 <template #head(agents)>
@@ -311,12 +308,12 @@ export default class ImmunizationHistoryReportComponent extends Vue {
             </b-table>
             <b-row class="mt-3">
                 <b-col class="col-7">
-                    <b-row class="d-none d-md-block">
+                    <b-row>
                         <b-col>
                             <h4>Recommended Immunizations</h4>
                         </b-col>
                     </b-row>
-                    <b-row class="d-none d-md-block">
+                    <b-row>
                         <b-col>
                             <div id="disclaimer">
                                 <p>
@@ -338,10 +335,7 @@ export default class ImmunizationHistoryReportComponent extends Vue {
                             </div>
                         </b-col>
                     </b-row>
-                    <b-row
-                        v-if="isRecommendationEmpty && !isLoading"
-                        class="mt-2 d-none d-md-block"
-                    >
+                    <b-row v-if="isRecommendationEmpty && !isLoading">
                         <b-col>No recommendations found.</b-col>
                     </b-row>
                     <b-table
@@ -352,7 +346,7 @@ export default class ImmunizationHistoryReportComponent extends Vue {
                         :items="recomendationItems"
                         :fields="recomendationFields"
                         data-testid="recommendation-history-report-table"
-                        class="mt-2 table-style d-none d-md-table"
+                        class="mt-2 table-style"
                     >
                         <template #table-busy>
                             <content-placeholders>
--- a/Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
@@ -277,7 +277,7 @@ export default class ImmunizationHistoryReportComponent extends Vue {
                 <b-col>No records found.</b-col>
             </b-row>
             <b-table
-                v-if="(!isEmpty || isLoading) && !isDependent"
+                v-if="!isDependent && (!isEmpty || isLoading)"
                 striped
                 :busy="isLoading"
                 :items="immunizationItems"
--- a/Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
+++ b/Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
@@ -38,6 +38,8 @@ export default class ImmunizationHistoryReportComponent extends Vue {
 
     @Prop() private filter!: ReportFilter;
 
+    @Prop({ default: false }) isDependent!: boolean;
+
     @Getter("immunizationsAreDeferred", { namespace: "immunization" })
     immunizationsAreDeferred!: (hdid: string) => boolean;
 
@@ -199,7 +201,9 @@ export default class ImmunizationHistoryReportComponent extends Vue {
                 records: this.immunizationItems,
                 recommendations: this.recomendationItems,
             },
-            template: TemplateType.Immunization,
+            template: this.isDependent
+                ? TemplateType.DependentImmunization
+                : TemplateType.Immunization,
             type: reportFormatType,
         });
     }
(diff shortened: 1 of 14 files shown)


mk08sh (commented):
👍✅

Apps/WebClient/src/ClientApp/src/components/report/ReportsComponent.vue
```diff
@@ -294,6 +368,13 @@ export default class ReportsComponent extends Vue {
...
+        if (this.isDependent) {
+            SnowPlow.trackEvent({
+                action: "download_report",
+                text: `Dependent_${this.reportComponentName}`,
+            });
+        }
+
         this.isGeneratingReport = true;
 
         this.trackDownload();
```
MikeLyttle, line 380: This should be integrated into `trackDownload()` so that only one event is issued when dowloading a dependent report.  You should also probably use the same report names it uses (but with "Dependent " or "Dependent_" in front).

Apps/WebClient/src/ClientApp/src/components/report/ReportsComponent.vue
```diff
@@ -45,6 +46,14 @@ const medicationRequestReport = "medication-request-report";
 const noteReport = "note-report";
 const laboratoryReport = "laboratory-report";
 const hospitalVisitReport = "hospital-visit-report";
+const medicationsText = "Medications";
```
BronzBierd, line 49: The report and the text can be merged into a single object map with the EntryType values this would allow you to loop through the configuration and tests each key against the config.

https://stackoverflow.com/questions/52296113/using-enum-as-object-keys

```JS
{
  [EntryType.Medication]: {
    value: "medication-report",
    text: "Medications"
    }, 
    // etc.
}
```

you could use Object.keys  to loop through reliable values to test against 

`ConfigUtil.isDependentDatasetEnabled(key)`
and
`ConfigUtil.isDatasetEnabled(key)`

This is much easier to maintain.

BronzBierd (changes requested):

Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
MikeLyttle, line 1: Can you provide screenshots comparing the appearance when no records are found for dependent immunizations vs. user immunizations, both desktop and mobile?  It's a bit harder to follow the logic in this template.

MikeLyttle (changes requested):

Apps/WebClient/src/ClientApp/src/components/report/ReportsComponent.vue
```diff
@@ -45,6 +46,14 @@ const medicationRequestReport = "medication-request-report";
...
+const healthVisitsText = "Health Visits";
+const covid19TestResultsText = "COVID‑19 Test Results";
+const immunizationsText = "Immunizations";
+const specialAuthorityRequestsText = "Special Authority Requests";
+const myNotesText = "My Notes";
+const laboratoryTestsText = "Laboratory Tests";
+const hospitalVisitsText = "Hospital Visits";
```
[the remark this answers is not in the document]
MikeLyttle: After checking, I noticed some slight differences between the names we were using here and the standard names we use to refer to the datasets throughout the app (i.e. the ones in entryTypeMap), so we should confirm that using the standard names in the report dropdown is a welcome change. Some functional tests might need to be updated to use the new values when interacting with the dropdown.

Differences:

- standard name for "Laboratory Tests" is "Lab Results"
- standard name for "COVID‑19 Test Results" is "COVID-19 Tests"
- standard name for "Special Authority Requests" is "Special Authority"

Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
MikeLyttle: There is a difference in what is shown on the dependents tab compared to the export page, given the same data.

Screenshot 2023-03-28 113507

Screenshot 2023-03-28 113549

I think we should always display the simple "No records found." when both are missing.  The bigger preview should only appear when there are immunizations or recommendations (and you're in desktop mode on the export page).

This would be a good time to simplify the logic for all of the exports.

```
<b-row v-if="isEmpty && !isLaboratoryLoading">
    <b-col>No records found.</b-col>
</b-row>
<b-table
    v-if="!isDependent && (!isEmpty || isLaboratoryLoading)"
```

This could be simplified to:

```
<div v-if="isEmpty && !isLoading">No records found.</div>
<b-table
    v-else-if="!isDependent"
```

The immunization one would be similar, except it would need a new "No records found." element and the v-else-if would have to be a new div that contains everything that's currently inside the \<section\>.

Apps/WebClient/src/ClientApp/src/components/report/ReportsComponent.vue
```diff
@@ -575,6 +588,7 @@ export default class ReportsComponent extends Vue {
                 ref="report"
                 :hdid="hdid"
                 :filter="reportFilter"
+                :is-dependent="isDependent"
```
MikeLyttle, line 591: The parent \<div\> should not have the "preview" class added for dependent reports, which don't have previews.  It's adding scrollbars & unnecessarily increasing the height of the tab.

Line 584:

```
class="preview"
```

Should be:

```
:class="{ preview: !isDependent }"
```

Apps/WebClient/src/ClientApp/src/components/report/ReportsComponent.vue
BronzBierd: Hi Ray, you can choose to take this or not, but it doesn't seem my comment made was fully understood, I'll add some detail.

My intention was something along the lines of:
```TypeScript
  loadReportOptions(isEnabledPredicate: (entryType: string) => boolean) {

        for (var entryType in EntryType) {
            if (isEnabledPredicate(entryType)) {
                this.reportTypeOptions.push({
                    // This could be where mike's suggestion comes in for the use of the entryTypeMap values or the config option I suggested.
                    // value: reportOptionConfig[entryType].component, or value: entryTypeMap[entryType].reports.componentName
                    // text: reportOptionConfig[entryType].component, or text: entryTypeMap[entryType].reports.optionText
                });
            }
        }
    }

    created(): void {
        this.logger = container.get<ILogger>(SERVICE_IDENTIFIER.Logger);

        const isEnabledPredicate: (dataSetName: EntryType) => boolean = this.isDependent
        ? ConfigUtil.isDatasetEnabled
        : ConfigUtil.isDependentDatasetEnabled;
        
        this.loadReportOptions(isEnabledPredicate);
    }
```

Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
rayking29: Or I could just make the dependent one work the same as the regular one when there are no imms records found. I see what you mean now, my changes actually change how the no records found case works in dependent as compared to regular.
Otherwise we will need to talk to Nino and Mandad about changing how they both work.

Apps/WebClient/src/ClientApp/src/components/report/ReportsComponent.vue
rayking29: I just went with Mike's suggestion and it works with less changes. However I see what you are suggesting and looks like a very good way to do it.

BronzBierd (approved):

Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
rayking29: Simplified as suggested and agreed by Nino and Mandad.

Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
```diff
@@ -257,7 +261,13 @@ export default class ImmunizationHistoryReportComponent extends Vue {
 
 <template>
     <div>
-        <section>
+        <div
+            v-if="isRecommendationEmpty && isEmpty && !isLoading"
+            class="mt-2 d-md-none"
```
MikeLyttle, line 266: This doesn't need these classes.

Apps/WebClient/src/ClientApp/src/components/report/ImmunizationHistoryReportComponent.vue
```diff
@@ -257,7 +261,13 @@ export default class ImmunizationHistoryReportComponent extends Vue {
...
+        >
+            No records found.
+        </div>
+        <section v-else-if="!isDependent">
```
MikeLyttle, line 270: I think you can simplify the display logic now by adding the `d-none d-md-block` classes to this element and removing all of the `d-none d-md-block` and `d-none d-md-table` classes from its descendants.

MikeLyttle (approved):

PR #4939, opened 2023-03-28
Admin protect dependent functional tests for AB#15118

BrianMaki:
# Implements [AB#15118](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15118)

## Description

Add functional tests for admin protect dependent:

- Updated seed script to load dependent guardian relationships in resource delegate table
- Added functional test to verify protect/cancel button(s) functionality
- Added functional test to verify protect/unprotect functionality including select to remove checkbox option

Screenshot 2023-03-27 at 6 56 41 PM

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

No

## Items to Review:

### DIFF
--- a/Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
+++ b/Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
@@ -1,6 +1,7 @@
 const dependentWithoutGuardian = { phn: "9874307168" };
 const dependentWithGuardian = { phn: "9874307175", guardianPhn: "9735353315" };
 const dependentExceedingAgeCutoff = { phn: "9735353315" };
+const protectDependent = "9872868095";
 
 function performSearch(phn) {
     cy.get("[data-testid=query-input]").clear().type(phn);
@@ -12,7 +13,7 @@ function getTableRows(tableSelector) {
     return cy.get(`${tableSelector} tbody`).find("tr");
 }
 
-describe("Delegation", () => {
+describe("Delegation Search", () => {
     beforeEach(() => {
         cy.login(
             Cypress.env("keycloak_username"),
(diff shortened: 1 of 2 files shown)


rayking29 (dismissed):

Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
```diff
@@ -93,3 +94,131 @@ describe("Delegation", () => {
...
+            "false"
+        );
+    });
+
+    it("Verify protect/unprotect dependent toggle, select to remove, save button and confirmation button are selected.", () => {
+        cy.get("[data-testid=query-input]").clear().type(protectDependent);
+        cy.get("[data-testid=search-button]").click();
+
+        // Confirm delegate table
+        getTableRows("[data-testid=delegate-table]").should("have.length", 2);
```
BronzBierd, line 168: Testing this makes the test brittle, without value.

For the dependent timeline test on another ticket I'm going to be adding a new dependent to the seed file, thus testing for 2 will break but it also doesn't add any value. I would argue that extracting the number of rows x (testing should check that its > 0) and then the result of removing the dependent should be equal to x - 1 will provide your result without being brittle.

BronzBierd (changes requested):

Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
```diff
@@ -93,3 +94,131 @@ describe("Delegation", () => {
...
+            .clear()
+            .type(dependentWithGuardian.phn);
+        cy.get("[data-testid=search-button]").click();
+
+        // Protect dependent toggle
+        cy.get("[data-testid=dependent-protected-switch]").should(
+            "have.attr",
+            "aria-checked",
+            "false"
+        );
```
MikeLyttle, line 118: Can we use the same style of assertions for the switch as the delegation search tests above?

```
cy.get("[data-testid=dependent-protected-switch]").should(
    "not.be.checked"
);
```

Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
```diff
@@ -1,6 +1,7 @@
 const dependentWithoutGuardian = { phn: "9874307168" };
 const dependentWithGuardian = { phn: "9874307175", guardianPhn: "9735353315" };
 const dependentExceedingAgeCutoff = { phn: "9735353315" };
+const protectDependent = "9872868095";
```
MikeLyttle, line 4: Nit: would suggest naming "dependentToProtect".  I also like the object syntax used for the others since it lets us group any related data we need (and not have to include PHN/HDID/etc. in the variable name).

Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
```diff
@@ -93,3 +94,131 @@ describe("Delegation", () => {
...
+            });
+
+        // Save button
+        cy.get("[data-testid=save-button]").click();
+
+        // Confirmation button
+        cy.get("[data-testid=confirm-button]").click();
+
+        // Confirm delegate table
+        getTableRows("[data-testid=delegate-table]").should("have.length", 2);
```
MikeLyttle, line 201: Shouldn't there only be 1 row once the protection is turned on, saved, and confirmed?

BrianMaki: Changed it.  Test passed with 2 but you're right.

MikeLyttle (approved):

BronzBierd (approved):

PR #4984, opened 2023-04-06
Fix updated by to use current user login for AB#15321

BrianMaki:
# Fixes [AB#15321](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15321)

## Description

Fix protect call to use current logged in user when populating 'UpdatedBy' in Dependent table.

Screenshot 2023-04-06 at 11 50 51 AM

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:


BronzBierd (approved):
Unusual that the previous version didn't work, it looks like perfectly normal initialization to me.

rayking29 (approved):

MikeLyttle:
> Unusual that the previous version didn't work, it looks like perfectly normal initialization to me.

The difference is the previous code only set the value for new Dependent rows.  Existing Dependents retrieved from the database would not be affected.

PR #4987, opened 2023-04-08
Create dependent audit tables for AB#15252

BrianMaki:
# Implements [AB#15252](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15252)

## Description

- Created dependent audit table as standalone
- Created dependent audit operation code table to be linked to dependent audit operation code.
- Created migration script to create dependent audit and dependent audit operation code tables. Migration script also seeds data to dependent adult operation table

Screenshot 2023-04-11 at 1 28 57 PM

Screenshot 2023-04-11 at 1 29 57 PM

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:


sslaws (changes requested):
Audit data should not be deleted

sslaws (changes requested):
audit information needs to persist deletes.  audit information should only reside in audit table.

Apps/Database/src/Context/GatewayDbContext.cs
```diff
@@ -387,6 +388,33 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                 .HasForeignKey(ad => ad.DependentHdId)
                 .OnDelete(DeleteBehavior.Cascade);
 
+            // Create Foreign key for DependentAudit to Dependent
+            modelBuilder.Entity<DependentAudit>()
+                .HasOne<Dependent>(da => da.Dependent)
+                .WithMany(d => d.DependentAudits)
+                .HasForeignKey(da => da.DelegateHdId)
+                .OnDelete(DeleteBehavior.SetNull);
```
BronzBierd, line 396: I'm not sure this qualifies to block the PR, however similarly with deleting the record entirely. I'm personally not a fan of losing the detail here, we would be disconnecting the action from the original target. it may feel redundant but an extra column for "TargetHdid" would ensure we don't lose the integrity of the data. And would allow easier tracing when viewing the data in an investigation post dependent deletion.

I argue, treat the table as a "warehouse table" remove the foreign key constraint (living with the performance tradeoff, potentially use an index to mitigate) and leave the data as is on deletion, or add the "duplicated" non-FK column as mentioned above.

CC: @sslaws, @ytqsl

Apps/Database/src/Migrations/GatewayDbContextModelSnapshot.cs
```diff
@@ -1384,6 +1384,126 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                    b.Property<string>("CreatedBy")
+                        .IsRequired()
+                        .HasMaxLength(60)
+                        .HasColumnType("character varying(60)");
+
+                    b.Property<DateTime>("CreatedDateTime")
+                        .HasColumnType("timestamp with time zone");
+
+                    b.Property<string>("DelegateHdId")
+                        .IsRequired()
```
BronzBierd, line 1407: If we're allowing nullable values this should not be `IsRequired()`

Apps/Database/src/Migrations/GatewayDbContextModelSnapshot.cs
```diff
@@ -3428,6 +3548,23 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .IsRequired();
                 });
 
+            modelBuilder.Entity("HealthGateway.Database.Models.DependentAudit", b =>
+                {
+                    b.HasOne("HealthGateway.Database.Models.Dependent", "Dependent")
+                        .WithMany("DependentAudits")
+                        .HasForeignKey("DelegateHdId")
+                        .OnDelete(DeleteBehavior.SetNull)
+                        .IsRequired();
```
BronzBierd, line 3557: Similar to above, is it right to set IsRequired() if nullable will be a valid value on deleting dependent?

BronzBierd (commented):
Hi Brian, some items may require discussion with the team as a whole to determine the convention moving forward.

There would be wiggle room to leave this type of change/refactor towards the end of the story if you want to move on, I'll bring it up for tomorrow's stand up.

let me know how you want to progress

Apps/Database/src/Migrations/GatewayDbContextModelSnapshot.cs
BrianMaki: Table is standalone now.

Apps/Database/src/Context/GatewayDbContext.cs
BrianMaki: This is no longer an issue. Discussed with @sslaws  This will be a standalone table.

BrianMaki:
> Hi Brian, some items may require discussion with the team as a whole to determine the convention moving forward.
> 
> There would be wiggle room to leave this type of change/refactor towards the end of the story if you want to move on, I'll bring it up for tomorrow's stand up.
> 
> let me know how you want to progress

This is no longer an issue.  Discussed with @sslaws

Apps/Database/src/Migrations/20230411193034_CreateDependentAuditTables.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            migrationBuilder.CreateTable(
+                name: "DependentAudit",
+                schema: "gateway",
+                columns: table => new
+                {
+                    DependentAuditId = table.Column<Guid>(type: "uuid", nullable: false),
+                    DelegateHdId = table.Column<string>(type: "character varying(52)", maxLength: 52, nullable: false),
+                    AgentUsername = table.Column<string>(type: "character varying(255)", maxLength: 255, nullable: false),
+                    ProtectedReason = table.Column<string>(type: "character varying(500)", maxLength: 500, nullable: false),
```
BronzBierd, line 59: Nit: A more `Reason` would be a better futureproof property name, as currently we're only capturing protection but that doesn't mean we won't have other audited actions in the future.

BronzBierd (dismissed):

Apps/Database/src/Models/DependentAudit.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <summary>
+    /// The dependent audit model.
+    /// </summary>
+    public class DependentAudit : AuditableEntity
+    {
+        /// <summary>
+        /// Gets or sets the id.
+        /// </summary>
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
```
sslaws, line 33: DB Column should be mapped to DependentAuditId, only the C# model should be Id.

BronzBierd (approved):

sslaws (changes requested):

sslaws (approved):

PR #5017, opened 2023-04-14
Extend configuration controller to return tour configuration AB#15331

BronzBierd:
# Fixes or Implements [AB#15331](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15331)

## Description

1. Extend Configuration returned by WebClient.Server.ConfigurationController
2. Add new TourConfiguration model, populated by DB's ApplicationSetting table.
3. Add caching to reduce communication with database on configuration requests.

## Testing

- [X] Unit Tests Updated
- [ ] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/Server/Models/TourConfiguration.cs
+++ b/Apps/WebClient/src/Server/Models/TourConfiguration.cs
@@ -31,15 +31,32 @@ namespace HealthGateway.WebClient.Server.Models
         public DateTime? LatestChangeDateTime { get; set; }
     }
 
-#pragma warning disable SA1600
-#pragma warning disable SA1602
-    internal static class TourSettingsMapper
+    /// <summary>
+    /// Contains the ApplicationSettings values and mapping logic for the TourConfiguration.
+    /// </summary>
+    public static class TourSettingsMapper
     {
+        /// <summary>
+        /// The application name.
+        /// </summary>
         public const string Application = "WEB";
+
+        /// <summary>
+        /// The component name.
+        /// </summary>
         public const string Component = "Tour";
+
+        /// <summary>
+        /// The latest change date time key.
+        /// </summary>
         public const string LatestChangeDateTime = "latestChangeDateTime";
 
-        public static TourConfiguration Map(IList<ApplicationSetting> settings)
+        /// <summary>
+        /// Transforms a list of ApplicationSetting to a TourConfiguration.
+        /// </summary>
+        /// <param name="settings">list of ApplicationSetting models.</param>
+        /// <returns cref="TourConfiguration">a populated TourConfiguration.</returns>
+        internal static TourConfiguration Map(IList<ApplicationSetting> settings)
         {
             TourConfiguration config = new();
             foreach (ApplicationSetting setting in settings)
@@ -57,6 +74,4 @@ namespace HealthGateway.WebClient.Server.Models
             return config;
         }
     }
-#pragma warning restore SA1600
-#pragma warning restore SA1602
 }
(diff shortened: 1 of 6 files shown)


Apps/WebClient/src/Server/Services/ConfigurationService.cs
```diff
@@ -66,5 +81,21 @@ public MobileConfiguration GetMobileConfiguration()
...
+        private TourConfiguration? GetTourConfiguration()
+        {
+            TourConfiguration? tourConfig = this.cacheProvider.GetItem<TourConfiguration>(this.tourSettingsCacheKey);
+            if (tourConfig == null)
+            {
+                IList<ApplicationSetting>? settings = this.applicationSettingsDelegate.GetApplicationSettings(TourSettingsMapper.Application, TourSettingsMapper.Component);
+                if (settings.Any())
+                {
+                    tourConfig = TourSettingsMapper.Map(settings);
+                    this.cacheProvider.AddItem(this.tourSettingsCacheKey, tourConfig);
```
ytqsl, line 94: 1. there's a getorset read through cache function that can be used
2. for how long this is going to be cached?

Apps/WebClient/src/Server/Models/TourConfiguration.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Gets or sets the tour update date time.
+        /// </summary>
+        public DateTime? LatestChangeDateTime { get; set; }
+    }
+
+    /// <summary>
+    /// Contains the ApplicationSettings values and mapping logic for the TourConfiguration.
+    /// </summary>
+    public static class TourSettingsMapper
```
ytqsl, line 37: Feels like this is an overkill to me, can you consider adding the application setting key to be passed to the delegate (maybe optional)? It can also be made generic and parse/cast the type in the service or delegate call?

Apps/WebClient/src/Server/Services/ConfigurationService.cs
BronzBierd: @ytqsl , Thanks for the getorset hint, I see its not often used in the code, I'll keep it in mind for future implementations.

There hasn't been any discussion on the expiry for the cache, however we could set it quite high as we would only set this value when we release thus in the sprint cycle. So I would argue once a day should be fine.

Do we clear cache on production deployment? or would we need to target this particular key on deployments?

Apps/WebClient/src/Server/Models/TourConfiguration.cs
BronzBierd: I'm using the delegate as it was built. And add this for extensibility. I'll consider a simplification.

Apps/WebClient/src/Server/Models/TourConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// <summary>
+    /// The Tour Configuration.
+    /// </summary>
+    public record TourConfiguration
+    {
+        /// <summary>
+        /// Gets or sets the tour update date time.
+        /// </summary>
+        public DateTime? LatestChangeDateTime { get; set; }
```
BrianMaki, line 28: Is it necessary to have the property defined as DateTime?  Are we gong to have more than one release per day?  Would DateOnly be simpler?

BronzBierd: As per the discussion the time will allow for a more accurate UX.

BrianMaki (approved):

BronzBierd:
Moving ApplicationSetting delegate to UserProfile endpoint to perform server side decision on whether to show the tour indicator or not,

separating the DB call from static configuration and moving it to behind the auth wall.

BronzBierd:
Replaced by #5034

PR #5022, opened 2023-04-15
Add functional test for dependent audit for AB#15269

BrianMaki:
# Implements AB#15269

## Description

- Added functional test to verify dependent audit can be shown and not shown after PHN query on delegation page.

Screenshot 2023-04-14 at 5 45 49 PM

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

No

## Items to Review:


rayking29 (approved):

Apps/Admin/Tests/Functional/cypress/integration/e2e/delegation.cy.js
```diff
@@ -89,6 +90,58 @@ describe("Delegation Search", () => {
             });
     });
 
+    it("Verify response contains dependent audit in descending datetime order as per seeded data.", () => {
```
BronzBierd, line 93: I'll still accept the PR with this, but this fixed index approach makes this test brittle. We should be searching for state's we're interested and not a fixed ordering.

This would be more resilient if we rather loop through the results  `[data-testid=delegation-change-{index}]` record the results in a test map and expect against an expected collection to be present

```
expected = {
  "admin": true,
  "support": true,
  "reviewer": true,
}
```
with the expected test outcome we will still conform to the intentional nature that tests need to be written in without relying on developers to remember not to affect the order of seed data.

BronzBierd (approved):

PR #5036, opened 2023-04-21
Fix unreliable idle detection AB#13081

MikeLyttle:
# Fixes [AB#13081](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/13081)

## Description

When tabs are inactive or the browser is minimized on a mobile device, our idle timeouts don't execute as desired.  I've introduced a ReliableTimer class, which stores the time it was started, then every second, compares the difference between the current time and that original time to determine if it has expired.  This means when the app is resumed, it will immediately detect that the timer has expired.

I also replaced idle-vue with an IdleDetector class that uses ReliableTimer, so if the app resumes and it's been longer than the combined idle time + modal countdown length since the last interaction, it can log out without displaying the modal and waiting for the countdown.  Hypothetically, it should also handle the case where you come back to the app at the point where it should be partway through the countdown: the modal should appear with that much time remaining.

I tested locally and did a couple runs in BrowserStack which seemed to behave as expected.

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/package-lock.json
+++ b/Apps/WebClient/src/ClientApp/package-lock.json
@@ -25,7 +25,6 @@
                 "file-saver": "^2.0.5",
                 "html2canvas": "^1.4.1",
                 "https-browserify": "^1.0.0",
-                "idle-vue": "^2.0.5",
                 "image-webpack-loader": "^8.1.0",
                 "inversify": "^6.0.1",
                 "keycloak-js": "^18.0.1",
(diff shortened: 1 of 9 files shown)


BronzBierd (approved):
I have a suspicion, that under computer load when tabs are put to sleep by chrome, that a user can bypass the logout by closing the tab (not focused) and then reopening it, its a little hard to test cause there are conditions that need to be met before a tab is put to sleep and trying to test it with 10-15 second window didn't fail. 

This is a major improvement and solves many of the issues originally listed so I'll leave it as a suspicion and test more thoroughly when I have the time to pick this up.

BrianMaki (approved):

PR #5038, opened 2023-04-21
Add Functional tests for app tour AB#15333

BronzBierd:
# Fixes or Implements AB#15333

## Description

1. Add test for new registration tour show.
2. Add tests for UI elements of tour modal.

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## Notes
Requires PR#5035

## Items to Review:


BrianMaki (approved):

Testing/functional/tests/cypress/integration/ui/user/appTour.js
```diff
@@ -0,0 +1,87 @@
...
+        cy.get("[data-testid=app-tour-next]").should("be.visible").click();
+
+        cy.get("[data-testid=app-tour-modal]").then((modal) => {
+            if (modal.find("[data-testid=app-tour-back]").length) {
+                cy.get("[data-testid=app-tour-back]")
+                    .should("be.visible")
+                    .click();
+            }
+            cy.log("No back button found");
+        });
```
MikeLyttle, line 85: This part of the test isn't doing much.
```suggestion
        cy.get("[data-testid=app-tour-back]")
            .should("be.visible")
            .should("be.enabled")
            .click();

        cy.get("[data-testid=app-tour-back]")
            .should("be.visible")
            .should("be.disabled");
```

BronzBierd: This is based on the configuration of the slides, if there is per se only 2 slides then there'll never be an active back button.

its conditional to manage the changing nature of the slides, if it exists it checks the availability and the click, if it doesn't a better check would be to ensure a done button is available.

I'll update the test to better reflect this but your suggestion would break the test if there is only 2 slides, not likely but possible.

MikeLyttle: I don't think we need to worry about there being only 2 slides. If there were only 2 slides, we wouldn't need a back button at all and this test could be removed.

Testing/functional/tests/cypress/integration/ui/user/appTour.js
```diff
@@ -0,0 +1,101 @@
...
+            // last slide app-tour-done should be visible
+            cy.get("[data-testid=app-tour-done]").should("be.visible").click();
+
+            // modal should be closed
+            cy.get("[data-testid=app-tour-modal]").should("not.exist");
+        });
+    });
+
+    // This greatly depends on the tour content, and isn't a great test but will test if the content is there
+    it("User should be able to navigate forwards and backwards if ", () => {
```
BronzBierd, line 75:
```suggestion
    it("User should be able to navigate forwards and backwards if greater than 2 slides", () => {
```

Testing/functional/tests/cypress/integration/ui/user/appTour.js
BronzBierd: I've amended to test the requirements more robustly.

MikeLyttle (approved):
Generally, `wait()` statments are discouraged in Cypress tests, but they will make it possible to see the slide content when the tests are run.

PR #5139, opened 2023-05-16
Add Refit and FluxorStore for BlockAccess AB#15410

BronzBierd:
# Fixes or Implements AB#15410

## Description

1. Add new Refit endpoint on ISupportAPI for BlockAccess
2. Add FluxorStore for new BlockAccess actions.
3. Update Verifications Fluxor store to properly represent the new PatientSupportDetails context.
4. Correct AgentAudit query to always filter by the appropriate Hdid not only if group is also set.

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

## Notes
All Unit tests still passing and view has been tested for new Verification message changes.
image

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
+++ b/Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
@@ -1,0 +1,88 @@
+﻿// -------------------------------------------------------------------------
+//  Copyright © 2019 Province of British Columbia
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------
+namespace HealthGateway.Admin.Client.Store.BlockAccess
+{
+    using System;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using Fluxor;
+    using HealthGateway.Admin.Client.Api;
+    using HealthGateway.Admin.Client.Store.PatientDetails;
+    using HealthGateway.Admin.Client.Store.PatientSupport;
+    using HealthGateway.Admin.Client.Utils;
+    using HealthGateway.Admin.Common.Models;
+    using Microsoft.Extensions.Logging;
+    using Refit;
+
+#pragma warning disable CS1591, SA1600
+    public class BlockAccessEffects
+    {
+        public BlockAccessEffects(ISupportApi api, ILogger<BlockAccessEffects> logger, IState<PatientSupportState> patientSupportState)
+        {
+            this.Api = api;
+            this.Logger = logger;
+            this.PatientSupportState = patientSupportState;
+        }
+
+        private ISupportApi Api { get; }
+
+        private ILogger<BlockAccessEffects> Logger { get; }
+
+        private IState<PatientSupportState> PatientSupportState { get; }
+
+        [EffectMethod]
+        public async Task HandleSetBlockAccessAction(BlockAccessActions.SetBlockAccessAction action, IDispatcher dispatcher)
+        {
+            this.Logger.LogInformation("Setting block access");
+            try
+            {
+                string? patientHdid = this.PatientSupportState.Value.Result?[0].Hdid;
+                if (patientHdid == null)
+                {
+                    RequestError error = new() { Message = "No patient selected" };
+                    dispatcher.Dispatch(new BlockAccessActions.SetBlockAccessFailureAction(error));
+                    return;
+                }
+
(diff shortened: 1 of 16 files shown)


Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private ISupportApi Api { get; }
+
+        private ILogger<BlockAccessEffects> Logger { get; }
+
+        private IState<PatientSupportState> PatientSupportState { get; }
+
+        [EffectMethod]
+        public async Task HandleSetBlockAccessAction(BlockAccessActions.SetBlockAccessAction action, IDispatcher dispatcher)
```
MikeLyttle, line 47: Rename to HandleBlockAccessAction.

Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
```diff
@@ -0,0 +1,88 @@
...
+            catch (Exception e) when (e is ApiException or HttpRequestException)
+            {
+                this.Logger.LogError("Error setting block access: {Exception}", e.ToString());
+                RequestError error = StoreUtility.FormatRequestError(e);
+                dispatcher.Dispatch(new BlockAccessActions.SetBlockAccessFailureAction(error));
+            }
+        }
+
+        [EffectMethod(typeof(BlockAccessActions.SetBlockAccessSuccessAction))]
+        public Task HandleSetBlockSuccessAction(IDispatcher dispatcher)
```
MikeLyttle, line 73: Rename to HandleBlockAccessSuccessAction.

BrianMaki (approved):

Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
```diff
@@ -0,0 +1,88 @@
...
+            this.Logger.LogInformation("Setting block access");
+            try
+            {
+                string? patientHdid = this.PatientSupportState.Value.Result?[0].Hdid;
+                if (patientHdid == null)
+                {
+                    RequestError error = new() { Message = "No patient selected" };
+                    dispatcher.Dispatch(new BlockAccessActions.SetBlockAccessFailureAction(error));
+                    return;
+                }
```
MikeLyttle, line 58: The HDID should instead be passed as a property of the action.  We may have retrieved multiple patients in the query.

This will remove the need for the `if` statement as well.

Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
```diff
@@ -0,0 +1,88 @@
...
+        [EffectMethod(typeof(BlockAccessActions.SetBlockAccessSuccessAction))]
+        public Task HandleSetBlockSuccessAction(IDispatcher dispatcher)
+        {
+            string? patientHdid = this.PatientSupportState.Value.Result?[0].Hdid;
+            if (patientHdid == null)
+            {
+                RequestError error = new() { Message = "No patient selected" };
+                dispatcher.Dispatch(new BlockAccessActions.SetBlockAccessFailureAction(error));
+                return Task.CompletedTask;
+            }
```
MikeLyttle, line 81: As above, the HDID should instead be passed as a property of the action.

Apps/Admin/Client/Store/PatientDetails/PatientSupportDetailsReducers.cs
```diff
@@ -33,27 +33,29 @@ public static MessageVerificationState ReduceLoadAction(MessageVerificationState
...
             {
                 IsLoading = false,
                 Result = action.Data,
                 Error = null,
-                Data = data,
+                MessagingVerifications = messageVerifications,
+                BlockedDataSources = action.Data?.BlockedDataSources,
+                AgentActions = action.Data?.AgentActions,
             };
         }
```
MikeLyttle, line 55: This reducer preserves existing messaging verifications for other patients, but doesn't support the same for blocked data sources or agent actions.  Since we perform a new request for patient details on every page load, we may as well make messaging verifications behave the same as the others, so the entire state is concerned with the most recently retrieved patient.

MikeLyttle (commented):
I suggest merging the BlockAccess store and the PatientDetails store, so there is just PatientDetails.

On the naming, the action that's being performed is "block[ing] access", or "updat[ing] data source access", so I prefer sticking with that terminology instead of using "set[ting] block[ed] access".

Apps/Admin/Client/Store/BlockAccess/BlockAccessEffects.cs
BronzBierd: This isn't the way it was done in the delegation store, it would retrieve the delegate from the state, but its not guaranteed that we clicked on the first patient, I'll rework this.

Apps/Admin/Client/Store/PatientDetails/PatientSupportDetailsReducers.cs
BronzBierd: I hope I understood your comment fully here. I removed the filtering to be inline with the other details values.

MikeLyttle: Yeah, that's what I was getting at.

MikeLyttle (approved):

BrianMaki (approved):

PR #5194, opened 2023-05-30
Update components to composition API (Group 2) AB#15553

MikeLyttle:
# Implements [AB#15553](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15553)

## Description

- updates components to composition API (Group 2)
   - refactors ReportsComponent to avoid subcomponent aliases
- removes extension recommendation for functionality now built in to vscode

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/constants/entryType.ts
+++ b/Apps/WebClient/src/ClientApp/src/constants/entryType.ts
@@ -1,3 +1,15 @@
+import { Component } from "vue";
+
+import ClinicalDocumentTimelineComponent from "@/components/timeline/entryCard/ClinicalDocumentTimelineComponent.vue";
+import Covid19LaboratoryOrderTimelineComponent from "@/components/timeline/entryCard/Covid19LaboratoryOrderTimelineComponent.vue";
+import DiagnosticImagingTimelineComponent from "@/components/timeline/entryCard/DiagnosticImagingTimelineComponent.vue";
+import EncounterTimelineComponent from "@/components/timeline/entryCard/EncounterTimelineComponent.vue";
+import HospitalVisitTimelineComponent from "@/components/timeline/entryCard/HospitalVisitTimelineComponent.vue";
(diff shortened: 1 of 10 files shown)


Apps/WebClient/src/ClientApp/src/constants/entryType.ts
```diff
@@ -18,7 +30,7 @@ export class EntryTypeDetails {
     name!: string;
     description!: string;
     icon!: string;
-    component!: string;
+    component!: Component;
```
BronzBierd, line 33: Nice way of dealing with this, will clean up any imports in the future 👍

Apps/WebClient/src/ClientApp/src/views/LandingView.vue
BronzBierd, line 1: I don't see the reasoning/requirement for the larger changes to this file, was this discussed in a private chat?

BronzBierd (approved):

Apps/WebClient/src/ClientApp/src/constants/entryType.ts
MikeLyttle: It also resolves an issue when these were being used.  The composition API SFCs don't seem to automatically register components referenced dynamically as strings, so they would have needed to be registered explicitly, by adding a separate `<script>` tag and exporting an object that lists all of the components.  It looked a bit messy.

Apps/WebClient/src/ClientApp/src/views/LandingView.vue
MikeLyttle: The main changes here are:

- removing references to vaccination banner in the template that are no longer in use because `isVaccinationBannerEnabled` was permanently set to false
- adding a typed string enum for the various preview devices instead of using strings
- replacing the `v-b-tooltip` directive with a `<b-tooltip>` to avoid having to call `this.$root.$emit("bv::hide::tooltip", ...)`

The first two are just cleanup / improvements made during the migration while the third was necessary.

PR #5196, opened 2023-05-30
Set communications page date and times to PST

BronzBierd:
# Fixes or Implements [AB#15558](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15558)

## Description

1. Add BaseTableFluxourComponent to centralise date time conversion methods for table components.
2. Consume conversion methods in the broadcast and communications table.

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

## Notes
UI remains unchanged and the time zone for the date conversions are centralised through the application's configuration.

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Components/AgentAudit/AgentAuditHistory.razor.cs
+++ b/Apps/Admin/Client/Components/AgentAudit/AgentAuditHistory.razor.cs
@@ -52,7 +52,7 @@ namespace HealthGateway.Admin.Client.Components.AgentAudit
         /// Gets or sets the application's configuration.
         /// </summary>
         [Inject]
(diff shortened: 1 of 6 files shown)


BrianMaki (changes requested):
There appears to be a DEV PR build error.

Screenshot 2023-05-30 at 2 12 27 PM

MikeLyttle (approved):
This is a good step but I'm curious now if there's a way we could accomplish this without inheritance - maybe we could create a utility/service has IConfiguration injected in its constructor and then inject that utility into any components we want.

BrianMaki (approved):

PR #5218, opened 2023-06-08
Guard Dataset API Calls with BlockedAccess for AB#15457

BrianMaki:
# Implements [AB#15457](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15457)

## Description

Guarding API calls for the following datasets:

- ClinicalDocument,
- Immunization
- SpecialAuthorityRequest
- Covid19TestResult
- LabResult
- HealthVisit
- HospitalVisit
- Medication,
- Note
- DiagnosticImaging
- OrganDonorRegistration

Added to and updated existing unit tests.

Screenshot 2023-06-08 at 10 51 50 AM

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

No

## Items to Review:

### DIFF
--- a/Apps/AccountDataAccess/src/Patient/PatientRepository.cs
+++ b/Apps/AccountDataAccess/src/Patient/PatientRepository.cs
@@ -85,6 +85,20 @@ namespace HealthGateway.AccountDataAccess.Patient
         }
 
         /// <inheritdoc/>
+        public async Task<bool> CanAccessDataSourceAsync(string hdid, DataSource dataSource, CancellationToken ct = default)
+        {
+            IEnumerable<DataSource> blockedDataSources = await this.GetDataSources(hdid, ct);
+            this.logger.LogDebug("Blocked data sources for hdid: {Hdid} - {DataSources}", hdid, blockedDataSources);
+
+            if (blockedDataSources.Any(ds => ds == dataSource))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <inheritdoc/>
         public async Task BlockAccess(BlockAccessCommand command, CancellationToken ct = default)
         {
             string authenticatedUserId = this.authenticationDelegate.FetchAuthenticatedUserId() ?? UserId.DefaultUser;
--- a/Apps/AccountDataAccess/src/Patient/PatientRepository.cs
+++ b/Apps/AccountDataAccess/src/Patient/PatientRepository.cs
@@ -90,12 +90,7 @@ namespace HealthGateway.AccountDataAccess.Patient
             IEnumerable<DataSource> blockedDataSources = await this.GetDataSources(hdid, ct);
             this.logger.LogDebug("Blocked data sources for hdid: {Hdid} - {DataSources}", hdid, blockedDataSources);
 
-            if (blockedDataSources.Any(ds => ds == dataSource))
-            {
-                return false;
-            }
-
-            return true;
+            return !blockedDataSources.Contains(dataSource);
         }
 
         /// <inheritdoc/>
--- a/Apps/AccountDataAccess/test/unit/PatientRepositoryTests.cs
+++ b/Apps/AccountDataAccess/test/unit/PatientRepositoryTests.cs
@@ -272,6 +272,44 @@ namespace AccountDataAccessTest
         }
 
         /// <summary>
+        /// Can access data source.
+        /// </summary>
+        /// <param name="dataSource">The data source to check for access.</param>
+        /// <param name="canAccessDataSource">The value indicates whether the data source can be accessed or not.</param>
+        /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
+        [Theory]
+        [InlineData(DataSource.Note, true)]
+        [InlineData(DataSource.Medication, false)]
+        public async Task CanAccessDataSource(DataSource dataSource, bool canAccessDataSource)
+        {
+            // Arrange
+            string hdid = Hdid;
+
+            HashSet<DataSource> dataSources = new()
+            {
+                DataSource.Immunization,
+                DataSource.Medication,
+            };
+
+            BlockedAccess blockedAccess = new()
+            {
+                Hdid = hdid,
+                DataSources = new HashSet<DataSource>
+                {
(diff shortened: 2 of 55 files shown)


Apps/Patient/src/Services/PatientDataService.cs
```diff
@@ -17,30 +17,43 @@ namespace HealthGateway.Patient.Services
...
 
         public async Task<PatientDataResponse> Query(PatientDataQuery query, CancellationToken ct)
         {
+            List<PatientDataType> patientDataTypes = query.PatientDataTypes.ToList();
+
+            if (!await this.CanAccessDataSourceAsync(query.Hdid, patientDataTypes).ConfigureAwait(true))
+            {
+                return new PatientDataResponse(Enumerable.Empty<PatientData>());
+            }
+
```
BronzBierd, line 54: NIT:

This could be simplified to still use the patientRepository's `CanAccessDataSourceAsync` by using it within a linq `Where`. The only caveat which you can use to determine if this is worthwhile or not, would be that you should either create an Async extension for `Where` or force the method into operate syncrhonously using `.Result`.

```CSharp
List<PatientDataType> patientDataTypes = query.PatientDataTypes
                .Where(p => this.patientRepository.CanAccessDataSourceAsync(query.Hdid, this.mapper.Map<DataSource>(p), ct).Result)
                .ToList();

if (!patientDataTypes.Any())
{
    return new PatientDataResponse(Enumerable.Empty<PatientData>());
}
```

You would not need the additional function below.
`private async Task<bool> CanAccessDataSourceAsync(string hdid, IList<PatientDataType> patientDataTypes)`

It does not affect the outcome of the tests:
image

Apps/AccountDataAccess/src/Patient/PatientRepository.cs
```diff
@@ -84,6 +84,20 @@ public async Task<PatientQueryResult> Query(PatientQuery query, CancellationToke
...
+        {
+            IEnumerable<DataSource> blockedDataSources = await this.GetDataSources(hdid, ct);
+            this.logger.LogDebug("Blocked data sources for hdid: {Hdid} - {DataSources}", hdid, blockedDataSources);
+
+            if (blockedDataSources.Any(ds => ds == dataSource))
+            {
+                return false;
+            }
+
+            return true;
```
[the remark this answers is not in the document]
BrianMaki: Changed.

Apps/Patient/src/Services/PatientDataService.cs
BrianMaki: PatientDataService.CanAccessDataSourceAsync is required as it removes any PatientDataType that have been blocked.

PatientDataTypes is list of PatientDataType so it can have both OrganDonorRegistrationStatus and DiagnosticImaging.  One is blocked while the other is not.  

Line 57 -  PatientDataQueryResult result = await this.patientDataRepository.Query(healthQuery, ct)
                .ConfigureAwait(true);

This queries based on the list.

So if one PatientDataType  from the list is blocked but the other is not, the above code logic will allow it still query.

Apps/Patient/src/Services/PatientDataService.cs
```diff
@@ -49,13 +62,29 @@ public async Task<PatientDataResponse> Query(PatientDataQuery query, Cancellatio
...
+                DataSource dataSource = this.mapper.Map<DataSource>(patientDataTypes[i]);
+                if (!await this.patientRepository.CanAccessDataSourceAsync(hdid, dataSource).ConfigureAwait(true))
+                {
+                    // Remove patient data type from query if access has been blocked.
+                    patientDataTypes.RemoveAt(i);
+                }
+            }
+
+            return patientDataTypes.Count != 0;
+        }
```
MikeLyttle, line 86: I like Declan's idea, but I think the async stuff might be too fancy (if trying to avoid the the blocking call to .Result).

Your method is doing 2 things - filtering the data types and counting them - when it could focus on just the filtering.

```
private static async Task<IList<PatientDataType>> GetUnblockedPatientDataTypesAsync(string hdid, IEnumerable<PatientDataType> patientDataTypes)
{
    List<PatientDataType> unblockedPatientDataTypes = new();
    foreach (PatientDataType patientDataType in patientDataTypes)
    {
        DataSource dataSource = this.mapper.Map<DataSource>(patientDataType);
        if (await CanAccessDataSourceAsync(hdid, dataSource).ConfigureAwait(true))
        {
            unblockedPatientDataTypes.Add(patientDataType);
        }
    }

    return unblockedPatientDataTypes;
}
```

Usage:

```
public async Task<PatientDataResponse> Query(PatientDataQuery query, CancellationToken ct)
{
    IList<PatientDataType> unblockedPatientDataTypes = await GetUnblockedPatientDataTypesAsync(query.PatientDataTypes).ConfigureAwait(true));
    if (!unblockedPatientDataTypes.Any())
    {
        return new PatientDataResponse(Enumerable.Empty<PatientData>());
    }
```

Apps/AccountDataAccess/test/unit/PatientRepositoryTests.cs
```diff
@@ -271,6 +271,44 @@ public async Task ShouldBlockAccess()
...
 
+        /// <summary>
+        /// Can access data source.
+        /// </summary>
+        /// <param name="dataSource">The data source to check for access.</param>
+        /// <param name="canAccessDataSource">The value indicates whether the data source can be accessed or not.</param>
+        /// <returns>A <see cref="Task"/> representing the result of the asynchronous operation.</returns>
+        [Theory]
+        [InlineData(DataSource.Note, true)]
+        [InlineData(DataSource.Medication, false)]
```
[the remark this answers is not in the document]
BrianMaki: Note is not blocked.  Note is not part of the data sources that are blocked.  So this tests if the data source can be accessed, which it can.  Medication is blocked.  So this data source cannot be accessed.

Apps/Patient/src/Services/PatientDataService.cs
BrianMaki: Changed

MikeLyttle (approved):

PR #5236, opened 2023-06-15
Use midnight pacific time not utc on default date values for AB#15718

BrianMaki:
# Fixes [AB#15718](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15718)

## Description

Their reason why there was a date discrepancy when running functional tests at 6:30 am was because the default date being used in the date picker was based on a UTC value and then converted back to PST.  

Example, when running the admin functional tests at 6:30 AM, the date picker uses UTC Now and then converts it PST for the default dates it uses for effective and expiry dates.

So UTC June 15 at 6:36 AM is actually June 14 11:36 PM in PST.

As a result, changed both Communication and Broadcast Dialogs to use PST as the default when initializing the date picker.

**See the screenshot.  You will notice the notifications were created the day before.**

Screenshot 2023-06-15 at 1 49 17 PM

**Communication functional tests:**

Screenshot 2023-06-15 at 1 43 03 PM

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Components/Communications/BroadcastDialog.razor.cs
+++ b/Apps/Admin/Client/Components/Communications/BroadcastDialog.razor.cs
@@ -157,8 +157,6 @@ public partial class BroadcastDialog : FluxorComponent
         {
             this.Broadcast.ActionUrl = null;
         }
-
-        Console.WriteLine($"Scheduled Date UTC: {this.Broadcast.ScheduledDateUtc}");
     }
 
     private void HandleClickCancel()
--- a/Apps/Admin/Client/Components/Communications/BroadcastDialog.razor.cs
+++ b/Apps/Admin/Client/Components/Communications/BroadcastDialog.razor.cs
@@ -122,7 +122,7 @@ public partial class BroadcastDialog : FluxorComponent
     {
         if (this.IsNewBroadcast)
         {
-            DateTime now = this.DateConversionService.ConvertFromUtc(DateTime.UtcNow);
+            DateTime now = this.DateConversionService.ConvertTime(DateTime.Now);
             DateTime tomorrow = now.AddDays(1);
             this.EffectiveDate = now.Date;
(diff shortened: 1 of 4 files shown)


MikeLyttle (dismissed):
I don't quite follow how this works differently from the old code.

When running at 6:30 AM our time, UtcNow should not be giving a time of "6:30 AM UTC" but giving the UTC equivalent of the local time.  If the local time was "6:30AM PDT", UtcNow should yield "1:30 PM UTC".

BronzBierd (approved):
I'm also not entirely sure, why this will fix the issue.

my understanding of the conversions:
[anyzone now] -> [pst now]
[utc now] -> [pst now] 

however they're both [pst now] in the end and should in theory be equivalent.

BronzBierd (approved):

BrianMaki:
> I don't quite follow how this works differently from the old code.
> 
> When running at 6:30 AM our time, UtcNow should not be giving a time of "6:30 AM UTC" but giving the UTC equivalent of the local time. If the local time was "6:30AM PDT", UtcNow should yield "1:30 PM UTC".

Original code:

1. Gets UTC Now = June 15, 630 AM
2. Converts it to Pacific = June 15 11:30 PM for display only
3. The DateTime object is still UTC.  Gets saved as UTC.

Change:

1. Gets DateTime Now with no timezone - June 15, 6:30 AM
2. 

> I'm also not entirely sure, why this will fix the issue.
> 
> my understanding of the conversions: [anyzone now] -> [pst now] [utc now] -> [pst now]
> 
> however they're both [pst now] in the end and should in theory be equivalent.

This is what's happening with current code:

Screenshot 2023-06-15 at 3 03 17 PM

MikeLyttle (approved):

sslaws (approved):

PR #5242, opened 2023-06-16
Stores - navbar translated AB#15704

BronzBierd:
# Fixes or Implements [AB#15704](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15704)

## Description

1. navbar translated

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/NewClientApp/src/stores/navbar.ts
+++ b/Apps/WebClient/src/NewClientApp/src/stores/navbar.ts
@@ -14,12 +14,7 @@ export const useNavbarStore = defineStore("navbar", () => {
     const isSidebarAnimating = ref<boolean>(false);
     const isHeaderShown = ref<boolean>(true);
 
-    const isSidebarOpen = computed(() => {
-        if (isSidebarOpenField.value === null) {
-            return !appStore.isMobile;
-        }
-        return isSidebarOpenField.value;
-    });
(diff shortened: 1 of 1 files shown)


Apps/WebClient/src/NewClientApp/src/stores/navbar.ts
```diff
@@ -0,0 +1,48 @@
...
+    const isSidebarOpenField = ref<boolean>();
+    const isSidebarAnimating = ref<boolean>(false);
+    const isHeaderShown = ref<boolean>(true);
+
+    const isSidebarOpen = computed(() => {
+        if (isSidebarOpenField.value === null) {
+            return !appStore.isMobile;
+        }
+        return isSidebarOpenField.value;
+    });
```
MikeLyttle, line 22: Since the type of the field is now `boolean | undefined` rather than `boolean | null`, the comparison should be against `undefined` rather than `null`.  It also means this can be simplified.

```suggestion
    const isSidebarOpen = computed(() => isSidebarOpenField.value ?? !appStore.isMobile);
```

BronzBierd: Good suggestion.

MikeLyttle (approved):

BrianMaki (approved):

PR #5269, opened 2023-06-22
Apply PR feedback for AB#15711

MikeLyttle:
- replace Bootstrap margins with equivalent Material margins
- fix preview device tooltips not disappearing
- add `text-body-1` class to all paragraphs
- simplify structure
- greatly simplify appearance of offline mode


BronzBierd (approved):
I agree that the you've simplified the implementation and kudos on fixing the tooltip issue, it was annoyingly inconsistent. Personally not a fan of using the template #slot methodology over the built-in components such as v-card-title tags to keep the template resistant to breaking changes from Vuetify updates.

I am not focused on paying back technical debt right now and focusing on a faithful conversion, I'll look out for the bootstrap classes vs material to be sure to remove them. I can't promise I'm going to tweak everything like you have, I personally think it takes too much time before receiving feedback from Mandad.

PR #5325, opened 2023-07-07
Convert FeedbackComponent

BronzBierd:
# Fixes or Implements AB#nnnnn

## Description

1. Convert FeedbackComponent
2. Introduce dialog method naming for:
    * AppTourComponent
    * RatingComponent
3. Updated calling component.
4. Reduced Sidebar icon complexity
5. Added feedback button in navigation drawer append slot

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

#### Open
image 

#### Rail
image

### Initial:
image

### Input:
image

### Validation:
image

### Submissions:
image
<br />
image
<br />
image

## Notes
Single buttons are now centrally aligned.

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
+++ b/Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
@@ -1,17 +1,17 @@
 <script setup lang="ts">
 import { useVuelidate } from "@vuelidate/core";
 import { maxLength, minLength, required } from "@vuelidate/validators";
-import { computed, ref } from "vue";
+import { computed, ref, unref } from "vue";
 
 import HgButtonComponent from "@/components/common/HgButtonComponent.vue";
 import HgIconButtonComponent from "@/components/common/HgIconButtonComponent.vue";
 import { container } from "@/ioc/container";
 import { SERVICE_IDENTIFIER } from "@/ioc/identifier";
 import { ResultError } from "@/models/errors";
-import User from "@/models/user";
 import { IUserFeedbackService } from "@/services/interfaces";
 import { useErrorStore } from "@/stores/error";
 import { useUserStore } from "@/stores/user";
+import ValidationUtil from "@/utility/validationUtil";
 
 defineExpose({
     showDialog,
@@ -39,19 +39,19 @@ const validations = computed(() => ({
 
 const v$ = useVuelidate(validations, { comment });
 
-const user = computed<User>(() => userStore.user);
-const hasEmail = computed<boolean>(
+const user = computed(() => userStore.user);
+const hasEmail = computed(
     () => user.value.verifiedEmail && user.value.hasEmail
 );
-const isInvalid = computed(() => ValidationUtil.isValid(v$.value) !== true);
-const resultTitle = computed<string>(() => {
+const isValid = computed(() => ValidationUtil.isValid(v$.value));
+const resultTitle = computed(() => {
     if (hasSubmitted.value) {
         return isSuccess.value ? "Received" : "Sorry";
     }
 
     return "";
 });
-const resultDescription = computed<string>(() => {
+const resultDescription = computed(() => {
     if (hasSubmitted.value) {
         return isSuccess.value
             ? "Your message has been sent successfully!"
@@ -60,8 +60,17 @@ const resultDescription = computed<string>(() => {
 
     return "";
 });
+const commentErrors = computed(() =>
+    v$.value.comment.$errors.map((e) => unref(e.$message))
+);
+const isSuccessWithoutEmail = computed(
+    () => isSuccess.value && !hasEmail.value
+);
+const isSuccessWithEmail = computed(() => isSuccess.value && hasEmail.value);
+const hasFailed = computed(() => !isSuccess.value && hasSubmitted.value);
 
 function showDialog(): void {
+    resetFeedback();
     visible.value = true;
 }
 
@@ -71,7 +80,7 @@ function hideDialog(): void {
 
 function onSubmit(): void {
     isLoading.value = true;
-    if (isInvalid.value) {
+    if (isValid.value !== true) {
         v$.value.$touch();
         isLoading.value = false;
         return;
@@ -96,20 +105,6 @@ function onSubmit(): void {
         });
 }
 
-const commentErrors = computed(() =>
-    v$.value.comment.$errors.map((e) => unref(e.$message))
-);
-
-const isSuccessWithoutEmail = computed<boolean>(
-    () => isSuccess.value && !hasEmail.value
-);
-
-const isSuccessWithEmail = computed<boolean>(
-    () => isSuccess.value && hasEmail.value
-);
-
-const hasFailed = computed(() => !isSuccess.value && hasSubmitted.value);
-
 function resetFeedback(): void {
     hasSubmitted.value = false;
     isSuccess.value = false;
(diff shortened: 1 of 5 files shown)


ytqsl (approved):

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+
+const v$ = useVuelidate(validations, { comment });
+
+const user = computed<User>(() => userStore.user);
+const hasEmail = computed<boolean>(
+    () => user.value.verifiedEmail && user.value.hasEmail
+);
+const isInvalid = computed<boolean>(
+    () => !v$.value.$dirty || (v$.value.$dirty && v$.value.$invalid)
+);
```
MikeLyttle, line 49:
```suggestion
const isInvalid = computed(() => ValidationUtil.isValid(v$.value) !== true);
```
This can be simplified with the ValidationUtil, however I'd suggest going further:

The property can be changed to `isValid` and return the result of `ValidationUtil.isValid()` directly.  This is how it is commonly used elsewhere.  Then places where `isInvalid.value` was used can check `isValid.value !== true` instead.

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+    comment: {
+        required,
+        minLength: minLength(5),
+        maxLength: maxLength(500),
+    },
+}));
+
+const v$ = useVuelidate(validations, { comment });
+
+const user = computed<User>(() => userStore.user);
```
MikeLyttle, line 43: The type parameter should not often be required.  (Applies throughout.)

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+            hasSubmitted.value = true;
+            isLoading.value = false;
+        });
+}
+
+const commentErrors = computed(() =>
+    v$.value.comment.$errors.map((e) =>
+        typeof e.$message === "string" ? e.$message : e.$message.value
+    )
+);
```
MikeLyttle, line 106:
```suggestion
const commentErrors = computed(() =>
    v$.value.comment.$errors.map((e) => unref(e.$message))
);
```
This can be simplified.  Note: `unref` will have to be improted from "vue".

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+function resetFeedback(): void {
+    visible.value = false;
+    // begin closing before changing state
+    setTimeout(() => {
+        hasSubmitted.value = false;
+        isSuccess.value = false;
+        comment.value = "";
+        v$.value.$reset();
+    }, 500);
+}
```
MikeLyttle, line 127:
```suggestion
function resetFeedback(): void {
    hasSubmitted.value = false;
    isSuccess.value = false;
    comment.value = "";
    v$.value.$reset();
}
```
It looks like the timeout was added because the dialog re-renders with changes before it is hidden (due to the animation of the dialog).

Rather than using a timeout, I suggest calling `resetState` inside `openDialog`, and changing everything that called `resetState` previously to instead call `closeDialog`.  Resetting the state before the dialog is displayed avoids any visual inconsistencies.

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+                <v-card-title class="px-0">
+                    <v-toolbar title="Feedback" density="compact" color="white">
+                        <HgIconButtonComponent
+                            data-testid="messageModalCloseButton"
+                            icon="close"
+                            @click="hideDialog"
+                        />
+                    </v-toolbar>
+                </v-card-title>
+                <v-card-text v-if="!hasSubmitted" class="text-body-1">
```
MikeLyttle, line 153:
```suggestion
                <v-card-text v-if="!hasSubmitted" class="pa-4 text-body-1">
```
Suggest adding this padding (which only affects the top) for consistency with `<MessageModalComponent />` and other dialogs.

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+                        auto-grow
+                        rows="5"
+                        data-testid="feedbackCommentInput"
+                        placeholder="Describe your suggestion or idea..."
+                        :error-messages="commentErrors"
+                        :disabled="isSuccess || isLoading"
+                        @input="v$.comment.$touch()"
+                    />
+                </v-card-text>
+                <v-card-text v-else>
```
MikeLyttle, line 171:
```suggestion
                <v-card-text v-else class="pa-4">
```

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+                    </template>
+                    <!-- Single button States (centered) -->
+                    <template v-else>
+                        <HgButtonComponent
+                            v-if="isSuccessWithEmail"
+                            data-testid="hasEmailResetFeedbackBtn"
+                            text="Got it!"
+                            variant="primary"
+                            :loading="isLoading"
+                            :disabled="isLoading"
```
MikeLyttle, line 254:
```suggestion
```

Setting the loading state also sets the disabled state.

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,272 @@
...
+        v-model="visible"
+        data-testid="feedbackContainer"
+        persistent
+        no-click-animation
+    >
+        <v-form
+            v-model="isInvalid"
+            class="d-flex justify-center"
+            @submit.prevent="onSubmit"
+        >
```
MikeLyttle, line 142: This can be changed to a `<form>` without a `v-model`.

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
BronzBierd: oh nice I didn't know about this 👍

Apps/WebClient/src/NewClientApp/src/components/site/SidebarComponent.vue
```diff
@@ -63,6 +65,10 @@ function dismiss() {
         collapsedOnDesktop.value = true;
     }
 }
+
+function handleSendFeedbackClick() {
+    feedbackDialog.value?.showDialog();
+}
```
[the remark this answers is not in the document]
BronzBierd: I've been avoiding template inline logic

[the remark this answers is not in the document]
MikeLyttle: I think it might be preferable for simple logic.

Apps/WebClient/src/NewClientApp/src/components/site/FeedbackComponent.vue
```diff
@@ -0,0 +1,253 @@
...
+                            :loading="isLoading"
+                            @click="hideDialog"
+                        />
+                        <HgButtonComponent
+                            v-if="hasFailed"
+                            data-testid="tryAgainBtn"
+                            text="Try Again"
+                            variant="primary"
+                            :loading="isLoading"
+                            :disabled="isValid !== true"
```
MikeLyttle, line 244:
```suggestion
                            :loading="isLoading"
                            :disabled="isLoading || isValid !== true"
```

When both apply, the disabled prop (unfortunately?) overrides the disabled styling applied by the loading prop.

BronzBierd: Sigh, cool thanks for that, I assumed the loading would take priority.

MikeLyttle (approved):

PR #5339, opened 2023-07-13
Migrate ProfileView & VerifySmsDialogComponent to vuetify #15807

MikeLyttle:
# Implements [AB#15807](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15807)

## Description

- improves spacing for page content, titles, and cards
- adds visible and invisible CSS classes for toggling element visibility
- fixes meta props in dependent and release note routes
- avoids relying on incomplete profile returned from close/recover account
- adds SectionHeaderComponent for simple headers on pages
- migrates ProfileView and VerifySmsDialogComponent to vuetify
  - splits ProfileView into ActiveUserProfileComponent and InactiveUserProfileComponent
  - ActiveUserProfileComponent is further split into UserProfileEmailComponent, UserProfileSmsComponent, UserProfileAddressComponent, and UserProfileManageAccountComponent

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/NewClientApp/src/components/private/profile/UserProfileAddressComponent.vue
+++ b/Apps/WebClient/src/NewClientApp/src/components/private/profile/UserProfileAddressComponent.vue
@@ -109,8 +109,7 @@ const postalAddressLabel = computed(() =>
             target="_blank"
             rel="noopener"
             >here</a
-        >
-        <span>.</span>
+        >.
     </div>
     <div v-if="!isSameAddress" class="mb-4 text-body-1">
         If either of these addresses is incorrect, update them
@@ -119,8 +118,7 @@ const postalAddressLabel = computed(() =>
             target="_blank"
             rel="noopener"
             >here</a
-        >
-        <span>.</span>
+        >.
     </div>
     <div v-if="!hasAddress" class="mb-4 text-body-1">
         To add an address, visit
@@ -129,7 +127,6 @@ const postalAddressLabel = computed(() =>
             target="_blank"
             rel="noopener"
             >this page</a
-        >
-        <span>.</span>
+        >.
     </div>
 </template>
--- a/Apps/WebClient/src/NewClientApp/src/components/private/profile/UserProfileAddressComponent.vue
+++ b/Apps/WebClient/src/NewClientApp/src/components/private/profile/UserProfileAddressComponent.vue
@@ -1,0 +1,135 @@
+<script setup lang="ts">
+import { computed } from "vue";
+
+import SectionHeaderComponent from "@/components/common/SectionHeaderComponent.vue";
+import { useUserStore } from "@/stores/user";
+
+const userStore = useUserStore();
+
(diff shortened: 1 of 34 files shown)


Apps/WebClient/src/NewClientApp/src/components/private/profile/UserProfileAddressComponent.vue
```diff
@@ -0,0 +1,135 @@
...
+    </div>
+    <div v-if="hasAddress && isSameAddress" class="mb-4 text-body-1">
+        If this address is incorrect, update it
+        <a
+            href="https://www.addresschange.gov.bc.ca/"
+            target="_blank"
+            rel="noopener"
+            >here</a
+        >
+        <span>.</span>
```
BronzBierd, line 113: Surely <span>necesssary?, is it there just so we can see it?

same below

Apps/WebClient/src/NewClientApp/src/utility/promiseUtility.ts
```diff
@@ -31,4 +34,26 @@ export default abstract class PromiseUtility {
...
     }
+
+    /**
+     * Wraps a promise so a loader is enabled beforehand and disabled afterward.
+     * @param promise The provided Promise.
+     * @param loader A loader to invoke.
+     * @param reason A reason for invoking the loader.
+     * @returns A new Promise.
+     */
+    public static async withLoader<T>(
```
BronzBierd, line 45: I would argue this could just be added directly to the loading store. Putting it here splits the tracking logic need to import 3 items in components that might both listen and set loading statuses. Moving this to the loadingStore reduces the complexity of the implementation steps.

```Typescript
loadingStore.trackPromise<T>(promise: Promise<T>, loader: Loader, reason?:string);

loadingStore.trackPromise(...yourPromiseFunctionInvocation, loader, "fancy reason");

```
Then, catch and finally works very similar to try catches and you can chain as many finally calls as needed. So you could rely on the promise's finally rather than a try catch, it will trigger regardless of promise rejections.
```Typescript
// This may need some playing but your tracking method body could be simplified to:
setIsLoading(true, loader, reason);
promise.finally(() => setIsLoading(false, loader, reason);
```

Apps/WebClient/src/NewClientApp/src/components/private/profile/UserProfileAddressComponent.vue
MikeLyttle: No, it's not necessary, so long as the period is moved directly after the ending `</a>` rather than on a new line.  Will fix.

Apps/WebClient/src/NewClientApp/src/utility/promiseUtility.ts
MikeLyttle: Done.

BronzBierd (approved):
The changes to the `loadingStore` are great 👍

PR #5343, opened 2023-07-14
Suggested PR changes for tasks/15782

MikeLyttle:
## Description

- fixes complaint about computed function not returning value
- removes unnecessary nullish coalescing operator
  - undefined and empty strings are treated the same in template
- renames reportName to reportEventName
  - the property is used for event names only, and the different pluralizations make it unsuitable for other uses
- fixes report control appearance on mobile
- hides report preview on small mobile devices
- tidies up report components

### DIFF
--- a/Apps/WebClient/src/NewClientApp/src/components/private/reports/NotesReportComponent.vue
+++ b/Apps/WebClient/src/NewClientApp/src/components/private/reports/NotesReportComponent.vue
@@ -11,7 +11,6 @@ import ReportFilter from "@/models/reportFilter";
 import ReportHeader from "@/models/reportHeader";
 import { ReportFormatType, TemplateType } from "@/models/reportRequest";
 import RequestResult from "@/models/requestResult";
-import UserNote from "@/models/userNote";
 import { ILogger, IReportService } from "@/services/interfaces";
 import { useNoteStore } from "@/stores/note";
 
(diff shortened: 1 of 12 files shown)


Apps/WebClient/src/NewClientApp/src/components/private/reports/NotesReportComponent.vue
```diff
@@ -57,34 +56,27 @@ const fields: ReportField[] = [
...
+            if (firstDate.isBefore(secondDate)) {
+                return 1;
+            }
+
+            if (firstDate.isAfter(secondDate)) {
+                return -1;
+            }
+
+            return 0;
+        })
```
BronzBierd, line 76: This is about the third time this dateWrapper sorting function is being defined, let's merge this into the current PR and I'll work to create a sorting utility to manage the different sorting methods employed, I believe there are a few in the stores that could be extracted into a encapsulated utility.

BronzBierd (approved):
Not a fan of renaming `reportName` -> `reportEventName`.

I feel the combination of the report name and the pdf makes the event name, that that the reportName is only linked to the events. Makes the property highly intentional and not as open to use elsewhere. But as we're not using anywhere else it should be fine to keep the change.

BronzBierd:
I've made the merge I'm going to add the sort utility on the main PR for the task.

PR #5348, opened 2023-07-17
Fix HgDatePickerComponent not respecting initial value AB#15831

MikeLyttle:
# Fixes [AB#15831](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15831)

## Description

Resolves issue where the initial prop value did not go through conversion, thereby causing the assignment to fail and an empty value to be assigned instead.

Also improves behaviour to handle blurring better (only when text input is left rather than when characters are entered).

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## Items to Review:


BrianMaki (approved):

BronzBierd (approved):
Looks good

PR #5373, opened 2023-07-28
Convert add dependent modal to Vue 3 for AB#15820.

BrianMaki:
# Implements [AB#15820](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15820)

## Description

- Converted new dependent component to VUE 3
- Added link to Add Dependent Dialog to Dependents 
- Updated functional tests as data-testid have been updated for some tags

**NOTE: DependentViewSelectorComponent.vue will be updated again via [AB#15821](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15821) which is currently being worked on.  Button has been added to test functionality.**

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

**Link from Dependents to Dialog:**

Screenshot 2023-07-27 at 11 02 19 PM

**Register button disabled when form validation fails:**

RegisterButtonDisabled

**Register button enabled after entering fields:**

Screenshot 2023-07-28 at 9 27 36 AM

**429 Error:**

429Error

**Data Mismatch Error:**

DataMismatch

## Notes

There seems to be a bug with the PageTitleComponent.  Both AddQuickLInkComponent  and AddDependentComponent have issues when the v-dialog is large when used in mobile mode like iPhone 4.  You are unable to scroll down on the page to see the complete v-dialog.  Created bug [AB#15871](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15871)

**iPhone 6/7/8 Mode Issue with PageTitleComponent:**

Screenshot 2023-07-27 at 8 48 25 PM

Screenshot 2023-07-27 at 10 53 28 PM

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
+++ b/Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
@@ -1,0 +1,457 @@
+<script setup lang="ts">
+import { useVuelidate } from "@vuelidate/core";
+import { helpers } from "@vuelidate/validators";
+import { minLength, required, sameAs } from "@vuelidate/validators";
+import { Duration } from "luxon";
+import { vMaska } from "maska";
+import { computed, nextTick, ref, watch } from "vue";
+
+import HgButtonComponent from "@/components/common/HgButtonComponent.vue";
+import HgDatePickerComponent from "@/components/common/HgDatePickerComponent.vue";
+import HgIconButtonComponent from "@/components/common/HgIconButtonComponent.vue";
+import LoadingComponent from "@/components/common/LoadingComponent.vue";
+import TooManyRequestsComponent from "@/components/error/TooManyRequestsComponent.vue";
+import { ActionType } from "@/constants/actionType";
+import { container } from "@/ioc/container";
+import { SERVICE_IDENTIFIER } from "@/ioc/identifier";
+import AddDependentRequest from "@/models/addDependentRequest";
+import { DateWrapper, IDateWrapper } from "@/models/dateWrapper";
+import { ResultError } from "@/models/errors";
+import { IDependentService } from "@/services/interfaces";
+import { useConfigStore } from "@/stores/config";
+import { useDependentStore } from "@/stores/dependent";
+import { useErrorStore } from "@/stores/error";
+import { useUserStore } from "@/stores/user";
+import { phnMask } from "@/utility/masks";
+import ValidationUtil from "@/utility/validationUtil";
+
+const emit = defineEmits<{
+    (e: "handle-submit"): void;
+}>();
+
+interface props {
+    disabled?: boolean;
+}
+
+withDefaults(defineProps<props>(), {
+    disabled: false,
+});
+
+defineExpose({
+    hideDialog,
+});
+
+const phnMaskaOptions = {
+    mask: phnMask,
+    eager: true,
+};
+
+const maxBirthdate = new DateWrapper();
+
+const dependentService = container.get<IDependentService>(
+    SERVICE_IDENTIFIER.DependentService
+);
+
+const configStore = useConfigStore();
+const dependentStore = useDependentStore();
+const errorStore = useErrorStore();
+const userStore = useUserStore();
+
+const isVisible = ref(false);
+const isLoading = ref(false);
+const errorMessage = ref("");
+const errorType = ref<ActionType | null>(null);
+const accepted = ref(false);
+const dependent = ref<AddDependentRequest>({
+    firstName: "",
+    lastName: "",
+    dateOfBirth: "",
+    PHN: "",
+});
+
+const hasValidationErrors = computed(() => {
+    v$.value.accepted.$touch();
+    return v$.value.$invalid;
+});
+const isError = computed(
+    () => errorType.value !== null || errorMessage.value.length > 0
+);
+const isErrorDataMismatch = computed(
+    () => errorType.value === ActionType.DataMismatch
+);
+const isErrorNoHdid = computed(() => errorType.value === ActionType.NoHdId);
+const isErrorProtected = computed(
+    () => errorType.value === ActionType.Protected
+);
+const isDependentAlreadyAdded = computed(() =>
+    dependentStore.dependents.some(
+        (d) =>
+            d.dependentInformation.PHN ===
+            dependent.value.PHN.replace(/\s/g, "")
+    )
+);
+const minBirthdate = computed<IDateWrapper>(() =>
+    new DateWrapper().subtract(
+        Duration.fromObject({ years: configStore.webConfig.maxDependentAge })
+    )
+);
+const validations = computed(() => ({
+    dependent: {
+        firstName: {
+            required: helpers.withMessage(
+                "Given names are required.",
+                required
+            ),
+        },
+        lastName: {
+            required: helpers.withMessage("Last name is required.", required),
+        },
+        dateOfBirth: {
+            required: helpers.withMessage("Invalid date.", required),
+            minLength: helpers.withMessage("Invalid date.", minLength(10)),
+            minValue: helpers.withMessage(
+                `Dependent must be under the age of
+                        ${configStore.webConfig.maxDependentAge}.`,
+                (value: string) =>
+                    new DateWrapper(value).isAfter(minBirthdate.value)
+            ),
+            maxValue: helpers.withMessage("Invalid Date.", (value: string) =>
+                new DateWrapper(value).isBefore(new DateWrapper())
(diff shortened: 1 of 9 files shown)


BronzBierd (changes requested):
I agree with the data-testid changes that would be better in line with the kebab-case convention however, Mike has set the convention if we're changing the casing to change it in both the new and the old component to allow for backwards compatibility with the Older ClientApp code for interim coverage.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+            ),
+        },
+    },
+    accepted: { isChecked: sameAs(true) },
+}));
+
+const v$ = useVuelidate(validations, { dependent, accepted });
+
+function addDependent(): void {
+    isLoading.value = true;
```
BronzBierd, line 144: This is a good component to use the loadingStore.

which works well with a single loading state. I haven't used it on the timeline components as we would be registering a dozen loaders which would trigger the loading event for a multitude of components.

this could be argued as a limitation of using enum type for loaders. however in Single loader components like page/form/non-repeating components it is a a good place to use the loadingStore.

https://github.com/bcgov/healthgateway/blob/features/15545/Apps/WebClient/src/NewClientApp/src/components/private/profile/ProfileView.vue has a good example of using the loadingStore.

In short you register to the loader through a computed property and then use the helper function to manage promises that should activate the loader, and bob's your uncle.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+        .finally(() => (isLoading.value = false));
+}
+
+function clear(): void {
+    dependent.value = {
+        firstName: "",
+        lastName: "",
+        dateOfBirth: "",
+        PHN: "",
+    };
```
BronzBierd, line 175: The empty dependent value is used twice, this can be moved to its on variable for reuse

```
const emptyDependent = { 
    //... your model here 
}

dependent.value = {...emptyDependent}
```

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+    <v-dialog
+        id="add-dependent-dialog"
+        v-model="isVisible"
+        data-testid="add-dependent-dialog"
+        scrollable
+        width="700px"
+        persistent
+        no-click-animation
+    >
+        <template #activator="slotProps">
```
BronzBierd, line 213: NO WORK, but need a decision.

I see there's some inconsistency in the implementations. We'll need to decide on a convention here.

I prefer `#activator="{ props: slotProps }"` for the definition, otherwise we'll need to consider something more "complete" if we want to get the entire value passed through like `#activator="slotObj"` and then consume as you have through `slotObj.props`

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+                                class="text-body-1"
+                            >
+                                If you continue to have issues, please contact
+                                <a href="mailto:HealthGateway@gov.bc.ca"
+                                    >HealthGateway@gov.bc.ca</a
+                                >.
+                            </span>
+                        </template>
+                    </v-alert>
+                    <v-container data-testid="newDependentModalText">
```
BronzBierd, line 297: this testid should be kebab-case for consistenc

I don't really think the v-container is necessary. If it helped with the layout though feel free to leave it.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+                        <v-row>
+                            <v-col cols="12" sm="6">
+                                <label for="firstName"
+                                    >First and Middle Names</label
+                                >
+                                <div>
```
BronzBierd, line 303: div is unnecessary here, v-text-field has display: block, thus it takes up the whole row.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+                                                v$.dependent.firstName
+                                            )
+                                        "
+                                        @blur="v$.dependent.firstName.$touch()"
+                                    />
+                                </div>
+                            </v-col>
+                            <v-col cols="12" sm="6">
+                                <label for="lastName">Last Name</label>
+                                <div>
```
BronzBierd, line 323: simliar no need for the divs, please see other instances below

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+                                    />
+                                </div>
+                            </v-col>
+                        </v-row>
+                        <v-row>
+                            <v-col>
+                                <v-checkbox
+                                    id="termsCheckbox"
+                                    v-model="accepted"
+                                    data-testid="termsCheckbox"
```
BronzBierd, line 386: another testId that could be changed to kebab-case

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,457 @@
...
+                                    color="primary"
+                                    class="text-body-1 checkbox-top"
+                                    hide-details
+                                    :error="
+                                        !ValidationUtil.isValid(v$.accepted)
+                                    "
+                                    @blur="v$.accepted.$touch()"
+                                >
+                                    <template #label>
+                                        <div class="ml-1">
```
BronzBierd, line 397: NIT: I'll leave this to your discretion:

I doubt the div is really necessary here too, similar to above, p is a block element and should take the entire width already. The margin class is unusual as the checkbox styling should already be managed by vuetify.

BronzBierd (changes requested):
Some minor changes, and a request to use the loadingStore for the consistency.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
BrianMaki: Changed.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
BrianMaki: Removed v-container.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
BrianMaki: No change.  Div makes a difference.

Screenshot 2023-07-28 at 2 14 29 PM

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
BrianMaki: Changed.

Apps/WebClient/src/NewClientApp/src/components/private/dependent/AddDependentComponent.vue
```diff
@@ -0,0 +1,445 @@
...
+const dependentStore = useDependentStore();
+const errorStore = useErrorStore();
+const loadingStore = useLoadingStore();
+const userStore = useUserStore();
+
+const isVisible = ref(false);
+const errorMessage = ref("");
+const errorType = ref<ActionType | null>(null);
+const accepted = ref(false);
+const dependent = ref<AddDependentRequest>(emptyDependent);
```
BronzBierd, line 74: This should be a copy of `emptyDependent`, so that we don't mutate the `emptyDependent` values
```suggestion
const dependent = ref<AddDependentRequest>({ ...emptyDependent });
```

BronzBierd (changes requested):

PR #5530, opened 2023-09-07
Agent Provisionning Flakey Tests 

sslaws:
# Fixes or Implements [AB#16005](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16005)

## Description
Split single Agent Provisioning Test into 4 (Create, Duplicate, Edit, Delete)
Create Test checks to see if the user FncTstUser1 exists in Keycloak and deletes it if needed
Removed the abundance of force clicks - the tests now behave like a user and click outside the select
Upgraded Cypress to V13 which has video disabled by default but had a new Replay feature

The bug mentioned other items but the test seems reliable at this point.  If not we can re-open the bug.

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes
None

## Items to Review:

### DIFF
--- a/Apps/Admin/Tests/Functional/cypress/integration/e2e/agentaccess.cy.js
+++ b/Apps/Admin/Tests/Functional/cypress/integration/e2e/agentaccess.cy.js
@@ -1,5 +1,8 @@
+import { removeUserIfExists } from "../../utilities/kcUtilities";
+
 const username = Cypress.env("idir_username");
 const password = Cypress.env("idir_password");
+const user = "FncTstUser1";
 
 describe("Provision", () => {
     beforeEach(() => {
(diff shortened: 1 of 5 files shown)


BrianMaki (approved):

Apps/Admin/Tests/Functional/cypress/integration/e2e/agentaccess.cy.js
```diff
@@ -50,62 +46,51 @@ describe("Provision", () => {
...
-        cy.get("[data-testid=save-btn]").click({ force: true });
+        cy.get("[data-testid=roles-select]").click();
+        cy.get("[data-testid=role]").contains("AdminAnalyst").click();
+        cy.get("[data-testid=save-btn]").parent().parent().click(0, 0);
+        cy.get("[data-testid=save-btn]").click();
         cy.get("[data-testid=provision-dialog-modal-text]").should("not.exist");
 
         cy.log("Validate user edit.");
-
+        const rowSelector = "[data-testid=agent-table] tbody tr.mud-table-row";
```
BronzBierd, line 88: NIT: Considering the variable was once used to reduce the duplication of the selector, this seems to go back on that decision making the variable unnecessary, shouldn't this be defined in a higher scope for all the tests to use, or just define the selector directly in the get method:

```TypeScript 
cy.get("[data-testid=agent-table] tbody tr.mud-table-row")
````

BronzBierd (approved):

Apps/Admin/Tests/Functional/cypress/integration/e2e/agentaccess.cy.js
sslaws: As discussed, I'll leave this as is as the usage is low.

MikeLyttle: Agreed - this should be moved outside the test definitions.

MikeLyttle (approved):

PR #5725, opened 2023-10-23
Implement save button for associating tags to feedback for AB#16346.

BrianMaki:
# Implements [AB#16346](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16346)

## Description

Instead of auto saving after each selection from MudSelect component, user will have to click on Save button to commit changes.

- Solution to modify the store's feedback's tags was discussed with @MikeLyttle 
- Implemented save button to associate tags to feedback for each feedback row
- Save button will be enabled if the row is dirty i.e., if any changes have been made on MudSelect component
- Dirty flag set when associating tag to feedback and dirty flag is reset in store after save is successful
- Warning will be displayed if user attempts to navigate away from Feedback page if there are any outstanding changes made to the feedback tag assocation - same implementation as DelegationPage.razor
- Date column widened to ensure there is no wrapping
- Used [SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "Team decision")] in UserFeedbackView as setter was required to set Tags.  This suppression is already being applied to other models such as ExternalConfiguration, AdminConfiguration, PhsaResult, Email, UserProfile, etc.
- Updated functional tests to use save button and to confirm save button is enabled and disabled

**Test Video:**

https://github.com/bcgov/healthgateway/assets/58790456/6dd80472-95f1-4264-8e29-839f6920bf53

**Functional Tests:**

Screenshot 2023-10-23 at 11 29 14 AM

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

Screenshot 2023-10-23 at 11 47 22 AM

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Pages/FeedbackPage.razor
+++ b/Apps/Admin/Client/Pages/FeedbackPage.razor
@@ -135,7 +135,7 @@ else
                         <MudSelect T="Guid"
                                    MultiSelection="true"
                                    SelectedValues="@context.TagIds"
-                                   SelectedValuesChanged="@(tagIds => AssociateTags(tagIds, context.Id))"
+                                   SelectedValuesChanged="@(tagIds => ChangeAssociatedTags(tagIds, context.Id))"
                                    MultiSelectionTextFunc="@DescribeTags"
                                    Disabled="@FeedbackUpdating"
                                    Variant="@Variant.Filled"
@@ -150,7 +150,7 @@ else
                         <MudIconButton OnClick="@(() => SaveAssociatedTags(context.Id))"
                                        Icon="fas fa-save"
                                        Color="@Color.Primary"
-                                       Disabled="@(!context.IsDirty)"
+                                       Disabled="@(!context.IsDirty || FeedbackUpdating)"
                                        data-testid="feedback-tag-save-button" />
                     </div>
                 </MudTd>
--- a/Apps/Admin/Client/Store/UserFeedback/UserFeedbackEffects.cs
+++ b/Apps/Admin/Client/Store/UserFeedback/UserFeedbackEffects.cs
@@ -19,30 +19,32 @@ using System;
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
+using AutoMapper;
 using Fluxor;
 using HealthGateway.Admin.Client.Api;
+using HealthGateway.Admin.Client.Models;
 using HealthGateway.Admin.Client.Utils;
 using HealthGateway.Admin.Common.Models;
 using HealthGateway.Common.Data.Constants;
 using HealthGateway.Common.Data.ViewModels;
-using Microsoft.AspNetCore.Components;
 using Microsoft.Extensions.Logging;
 using Refit;
 
 #pragma warning disable CS1591, SA1600
 public class UserFeedbackEffects
 {
-    public UserFeedbackEffects(ILogger<UserFeedbackEffects> logger, IUserFeedbackApi api)
+    public UserFeedbackEffects(ILogger<UserFeedbackEffects> logger, IUserFeedbackApi api, IMapper autoMapper)
     {
         this.Logger = logger;
         this.Api = api;
+        this.AutoMapper = autoMapper;
     }
 
-    [Inject]
-    private ILogger<UserFeedbackEffects> Logger { get; set; }
+    private ILogger<UserFeedbackEffects> Logger { get; }
 
-    [Inject]
-    private IUserFeedbackApi Api { get; set; }
+    private IUserFeedbackApi Api { get; }
+
+    private IMapper AutoMapper { get; }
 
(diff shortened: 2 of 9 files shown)


mk08sh (commented):
👍

Apps/Admin/Common/Models/UserFeedbackView.cs
```diff
@@ -83,9 +84,15 @@ public UserFeedbackView(IList<UserFeedbackTagView> tags)
     public string Email { get; set; } = string.Empty;
 
     /// <summary>
-    /// Gets the feedback admin tags.
+    /// Gets or sets a value indicating whether the feedback has been modified or not.
     /// </summary>
-    public IList<UserFeedbackTagView> Tags { get; } = new List<UserFeedbackTagView>();
+    public bool IsDirty { get; set; }
```
MikeLyttle, line 89: This property is only relevant for the front-end and should not be included in this model.

To track this in the front-end, you can create an extended model - see ExtendedDelegateInfo as the best example of this, which uses an AutoMapper profile for easy conversion.  The conversion happens in the effect method.

Apps/Admin/Common/Models/UserFeedbackView.cs
```diff
@@ -83,9 +84,15 @@ public UserFeedbackView(IList<UserFeedbackTagView> tags)
...
+
+    /// <summary>
+    /// Gets or sets the feedback admin tags.
+    /// </summary>
+    [SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "Team decision")]
+    public IList<UserFeedbackTagView> Tags { get; set; } = new List<UserFeedbackTagView>();
```
MikeLyttle, line 95: See Admin.Common/Models/DelegationInfo.cs for our current approach that avoids the warning and suppression, along with the boilerplate needed for the constructors.

Apps/Admin/Common/Models/UserFeedbackView.cs
```diff
@@ -95,4 +102,14 @@ public UserFeedbackView ShallowCopy()
...
+
+    /// <summary>
+    /// Sets the feedback admin tags and marks the instance as "dirty".
+    /// </summary>
+    /// <param name="tags">A list of feedback admin tags.</param>
+    public void SetTags(IList<UserFeedbackTagView> tags)
+    {
+        this.Tags = tags;
+        this.IsDirty = true;
+    }
```
MikeLyttle, line 114: This should move to a store action.

Apps/Admin/Client/Pages/FeedbackPage.razor
```diff
@@ -129,21 +131,28 @@ else
...
+                                   data-testid="feedback-tag-select">
+                            @foreach (AdminTagView tag in Tags)
+                            {
+                                <MudSelectItem T="Guid" Value="@tag.Id" data-testid="feedback-tag">@tag.Name</MudSelectItem>
+                            }
+                        </MudSelect>
+                        <MudIconButton OnClick="@(() => SaveAssociatedTags(context.Id))"
+                                       Icon="fas fa-save"
+                                       Color="@Color.Primary"
+                                       Disabled="@(!context.IsDirty)"
```
MikeLyttle, line 153: I think this should also be disabled when FeedbackUpdating is true.

(I was going to make a comment about setting the "Loading" property as well, but unfortunately, we don't yet have an HgIconButton custom component that supports it the same way we do for HgButton.)

Apps/Admin/Client/Store/UserFeedback/UserFeedbackEffects.cs
```diff
@@ -109,6 +109,10 @@ public async Task HandleAssociateTagsAction(UserFeedbackActions.AssociateTagsAct
             if (response is { ResourcePayload: { }, ResultStatus: ResultType.Success })
             {
                 this.Logger.LogInformation("Tags associated to user feedback successfully!");
+
+                // Associate Tags succeeded so reset IsDirty property to false.
+                response.ResourcePayload.IsDirty = false;
```
MikeLyttle, line 114: This should move to the reducer.

Apps/Admin/Common/Models/UserFeedbackView.cs
BrianMaki: Changed.

Apps/Admin/Common/Models/UserFeedbackView.cs
BrianMaki: Changed.

Apps/Admin/Client/Store/UserFeedback/UserFeedbackEffects.cs
BrianMaki: Changed.

MikeLyttle (approved):

BronzBierd (approved):

PR #5735, opened 2023-10-26
Update models & back-end for patient details dependents AB#15576

MikeLyttle:
# Implements [AB#15576](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15576) and [AB#15577](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15577)

## Description

- updates patient details model to include dependents [AB#15576](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15576)
- updates back-end and unit tests for patient details dependents [AB#15577](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/15577)
- refactors GetPatientSupportDetailsAsync to reduce complexity and coupling

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## Notes

image

## Items to Review:


Apps/Admin/Server/Services/SupportService.cs
```diff
@@ -211,6 +207,52 @@ public async Task BlockAccessAsync(string hdid, IEnumerable<DataSource> dataSour
...
+            List<PatientSupportDependentInfo> dependents = new();
+
+            IList<ResourceDelegate> resourceDelegates = await this.SearchDependents(delegateHdid);
+            foreach (ResourceDelegate resourceDelegate in resourceDelegates)
+            {
+                PatientModel? dependentPatient = await this.GetPatientAsync(PatientIdentifierType.Hdid, resourceDelegate.ResourceOwnerHdid, ct);
+                if (dependentPatient != null)
+                {
+                    PatientSupportDependentInfo dependentInfo = this.autoMapper.Map<PatientModel, PatientSupportDependentInfo>(dependentPatient);
+                    dependentInfo.ExpiryDate = resourceDelegate.ExpiryDate;
```
BrianMaki, line 221: Would this be not cleaner by creating a mapper util and pass in ExpiryDate like it is done with email via UserFeedbackMapUtils.cs?

Apps/Admin/Server/Controllers/SupportController.cs
```diff
@@ -116,6 +116,7 @@ public async Task<IEnumerable<PatientSupportResult>> GetPatients([FromQuery] Pat
                     includeEverything,
                     includeEverything,
                     includeEverything,
+                    includeEverything,
```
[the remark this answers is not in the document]
BronzBierd: https://stackoverflow.com/questions/25302178/using-async-tasks-with-the-builder-pattern

Apps/Admin/Server/Services/ISupportService.cs
BronzBierd, line 1: depending on the reception of my previous comment on the builder pattern, this could be simplified and the support service signature could be simplified.

Apps/Admin/Server/Services/SupportService.cs
BronzBierd, line 1: If the builder pattern is accepted a new class would be required to break out these requests to individual async calls that would be awaited in the builder class.

Apps/Database/src/Delegates/DbResourceDelegateDelegate.cs
```diff
@@ -208,7 +208,7 @@ public async Task<ResourceDelegateQueryResult> SearchAsync(ResourceDelegateQuery
                 dbQuery = dbQuery.Take(query.TakeAmount.Value);
             }
 
-            IEnumerable<ResourceDelegate> items = await dbQuery.ToArrayAsync().ConfigureAwait(true);
+            IList<ResourceDelegate> items = await dbQuery.ToArrayAsync().ConfigureAwait(true);
```
BronzBierd, line 211:
```suggestion
            IList<ResourceDelegate> items = await dbQuery.ToArrayAsync();
```

ConfigureAwait is no longer required.

https://devblogs.microsoft.com/dotnet/configureawait-faq/ has a great break down of the reason we would want to use the ConfigureAwait(false), however he (the author) also states: 
> Why would I want to use ConfigureAwait(true)?
You wouldn’t, unless you were using it purely as an indication that you were purposefully not using ConfigureAwait(false) (e.g. to silence static analysis warnings or the like). ConfigureAwait(true) does nothing meaningful. When comparing await task with await task.ConfigureAwait(true), they’re functionally identical. If you see ConfigureAwait(true) in production code, you can delete it without ill effect.
> 
> The ConfigureAwait method accepts a Boolean because there are some niche situations in which you want to pass in a variable to control the configuration. But the 99% use case is with a hardcoded false argument value, ConfigureAwait(false).

Apps/Admin/Server/Services/SupportService.cs
MikeLyttle: If we were performing this conversion in multiple places, I would agree, but as this is the only place, I don't think the extra overhead is worth it.

Apps/Admin/Server/Controllers/SupportController.cs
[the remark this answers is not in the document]
MikeLyttle: Alternatively, we could create a model and pass the model, e.g.

```
{
    QueryType: queryType,
    QueryString: queryString,
    IncludeMessagingVerification: userIsAdmin || userisReviewer,
    IncludeBlockedDataSources: userIsAdmin || userisReviewer,
    IncludeAgentActions: userIsAdmin || userisReviewer,
    IncludeDependents: userIsAdmin || userisReviewer,
    IncludeCovidDetails: userIsSupport,
    IncludeDependents: userIsAdmin || userisReviewer,
    RefreshVaccineDetails: refreshVaccineDetails,
}
```

Either way, I think it's worth creating a separate task since it will involve a number of changes for the unit tests.

Apps/Admin/Common/Models/PatientSupportDependentInfo.cs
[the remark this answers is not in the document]
MikeLyttle: Good catch - I forgot about that.  I'll need to query the Dependent table to retrieve it.

BrianMaki (approved):

Apps/Database/src/Delegates/DbResourceDelegateDelegate.cs
MikeLyttle: I didn't introduce this one but it can definitely be removed.

Apps/Admin/Common/Models/PatientSupportDependentInfo.cs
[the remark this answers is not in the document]
MikeLyttle: Fixed.

Apps/Database/src/Delegates/DbResourceDelegateDelegate.cs
MikeLyttle: Removed.

Apps/Admin/Server/Controllers/SupportController.cs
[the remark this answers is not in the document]
BronzBierd: I'll create a task for that

BronzBierd (approved):

PR #5761, opened 2023-11-01
Refactor SupportService for improved parallelism AB#16398

MikeLyttle:
# Implements [AB#16398](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16398)

## Description

- updates SupportService to perform external queries in parallel (GetVaccineDetails, GetCovidAssessmentDetails, patient repository queries)
  - queries involving our database are not parallelized because they all interact with the same DbContext due to our current dependency injection, and parallel operations are not supported on the same DbContext
- expands IResourceDelegateDelegate Query to be able to retrieve both ResourceDelegate and Dependent information in one query, removing the need for separate queries to retrieve each Dependent
- updates SupportService to take advantage of the unused PatientSupportDetailsMapUtils and remove the unnecessary MessagingVerificationMapUtils
  - the MessagingVerificationMapUtils method had a TimeZoneInfo as a parameter but was not using it, so the dependency on the configuration and the calculation of the time zone was also able to be removed
- tidies up SupportService
  - removes ConfigureAwait()
  - renames one of the overloads of GetPatientAsync to TryGetPatientAsync to reflect its differing functionality
  - reorders private methods to reflect the order in which they are referenced

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## Items to Review:


BrianMaki (dismissed):

Apps/Admin/Server/Services/SupportService.cs
```diff
@@ -114,57 +104,43 @@ public class SupportService : ISupportService
...
-                vaccineDetails = await this.GetVaccineDetails(patient, query.RefreshVaccineDetails).ConfigureAwait(true);
-                covidAssessmentDetails = await this.immunizationAdminApi.GetCovidAssessmentDetails(new() { Phn = patient.Phn }, this.GetAccessToken()).ConfigureAwait(true);
-            }
+            IEnumerable<AgentAction>? agentActions =
+                query.IncludeAgentActions ? await this.GetAgentActionsAsync(patient.Hdid, ct) : null;
+
+            IEnumerable<PatientSupportDependentInfo>? dependents =
+                query.IncludeDependents ? await this.GetAllDependentInfoAsync(patient.Hdid, ct) : null;
 
             PatientSupportDetails details = new()
```
BronzBierd, line 136: No Change:

I'm surprised the linter didn't suggest this to be an immediate return:

```CSharp
return new () 
{
  // ... all properties here
}
```

Apps/Admin/Server/Services/SupportService.cs
```diff
@@ -203,33 +179,32 @@ public async Task BlockAccessAsync(string hdid, IEnumerable<DataSource> dataSour
...
         }
 
-        private async Task<IList<PatientSupportDependentInfo>> GetDependentsAsync(string delegateHdid, CancellationToken ct)
+        private async Task<PatientModel> GetPatientAsync(PatientDetailsQuery query, CancellationToken ct = default)
+        {
+            PatientModel? patient = (await this.patientRepository.Query(query, ct)).Items.SingleOrDefault();
+            return patient ?? throw new ProblemDetailsException(ExceptionUtility.CreateProblemDetails(ErrorMessages.ClientRegistryRecordsNotFound, HttpStatusCode.NotFound, nameof(SupportService)));
+        }
+
+        private async Task<PatientModel?> TryGetPatientAsync(PatientIdentifierType identifierType, string queryString, CancellationToken ct)
```
BronzBierd, line 188: I have not better suggestion but with the `Try` prefix, it feels like it should be a predicate with an `out` parameter.

MikeLyttle: Yeah, I couldn't think of anything better. GetPatientOrNullAsync was my number 2 choice.

BronzBierd (approved):
Nice

Apps/Admin/Server/Services/SupportService.cs
BronzBierd: Hmm the second one is too on the nose. I'm happy to adjust my perspective for the sake of the nullable type

BrianMaki (approved):

PR #5788, opened 2023-11-11
Remove obsolete vue cli middleware and replace 

sslaws:
# Fixes or Implements [AB#16480](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16480)

## Description
Reviewing packages for the upcoming .Net 8 upgrade and found that eeparker/vueclimiddleware is marked as archived as of November 6th.  As per our normal process of only using maintained software packages, I found that the MS SPA Proxy does what we need it to do with a few tweaks.

Removed eeparker/aspnetcore-vueclimiddleware
Removed middleware configuration 
Added Microsoft.AspNetCore.SpaProxy
Added new csproj configuration
The front-end now runs on 5002 to maintain better compatibility with tests and authentication 

As the new proxy code actually switches over to the vue application, I needed to make a small change to the restConfigService to be able to flip between using the localhost:3025/configuration endpoint or the /configuration endpoint.

Bonus:  it starts up really fast now

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

## UI Changes
None

## Items to Review:


Apps/WebClient/src/ClientApp/src/services/restConfigService.ts
```diff
@@ -5,8 +5,10 @@ import { IConfigService, IHttpDelegate, ILogger } from "@/services/interfaces";
 import ErrorTranslator from "@/utility/errorTranslator";
 
 export class RestConfigService implements IConfigService {
-    private readonly CONFIG_BASE_URI: string = "/configuration";
     private logger;
+    private readonly CONFIG_BASE_URI: string = `${
+        import.meta.env.VITE_CONFIG_BASE_URI || ""
+    }/configuration`;
     private http;
```
MikeLyttle, line 12: To mirror the other services, I'd suggest separating into 2 properties.

```
    private readonly CONFIG_BASE_URI = "configuration";
    private logger;
    private http;
    private baseUri = import.meta.env.VITE_CONFIG_BASE_URI || "/";
```

This would require a change in the getConfiguration() function:

```
            .getWithCors<ExternalConfiguration>(`${this.baseUri}${this.CONFIG_BASE_URI}/`)
```

And a change to the .env files to include a trailing slash.

Since this is essentially the endpoint for the WebClient server, maybe the environment variable should be named WEB_CLIENT_BASE_URI instead of CONFIG_BASE_URI?

sslaws: I've updated the variable name but it has to be prefixed with VITE_ to be exposed.

Apps/WebClient/src/ClientApp/src/services/restConfigService.ts
```diff
@@ -5,9 +5,10 @@ import { IConfigService, IHttpDelegate, ILogger } from "@/services/interfaces";
 import ErrorTranslator from "@/utility/errorTranslator";
 
 export class RestConfigService implements IConfigService {
-    private readonly CONFIG_BASE_URI: string = "/configuration";
     private logger;
+    private readonly CONFIG_BASE_URI: string = "configuration" ;
     private http;
+    private baseUri = import.meta.env.VITE_WEB_CLIENT_BASE_URI || "/" ;
```
MikeLyttle, line 11:
```suggestion
    private readonly CONFIG_BASE_URI: string = "configuration";
    private http;
    private baseUri = import.meta.env.VITE_WEB_CLIENT_BASE_URI || "/";
```

Just removing the extra spaces.  (This should happen automatically on save when using VS Code and opening the workspace.)

MikeLyttle (approved):

BronzBierd (approved):

BrianMaki (approved):

PR #5799, opened 2023-11-15
Add delegate store and service for AB#16181

BronzBierd:
# Implements [AB#16181](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16181)

## Description
1. Add Delegate Pinia store
    * ref: all invitations
    * ref: current "new" invitation
    * action: create new invitation
    * action: reset new invitation
2. Add IDelegateService interface for Restful API
3. Add RestDelegateService implementation.

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

No tests added as back end is required for this next step.

## UI Changes
No UI implementation

## Items to Review:

### DIFF
--- a/Apps/WebClient/src/ClientApp/src/models/delegateInvitation.ts
+++ b/Apps/WebClient/src/ClientApp/src/models/delegateInvitation.ts
@@ -7,30 +7,37 @@ export interface DelegateInvitation {
     /**
      * The delegate invitation id.
      */
-    id: string;
+    id?: string;
 
     /**
      * The friendly name of the delegate.
      */
-    nickname: string;
+    nickname?: string;
 
     /**
      * The delegate invitation status.
+     * Pending, Active, Declined, Locked, InviteExpired, AccessExpired
      */
-    status: string;
+    status?:
+        | "Pending"
+        | "Active"
+        | "Declined"
+        | "Locked"
+        | "InviteExpired"
+        | "AccessExpired";
 
     /**
      * The email address to send the invitation.
      */
-    email: string;
+    email?: string;
 
     /**
      * The expiry date of access.
      */
-    expiryDate: string;
+    expiryDate?: string;
 
     /**
      * The list of data sources the delegate has access to.
      */
-    dataSources: DataSource[];
+    dataSources?: DataSource[];
 }
--- a/Apps/WebClient/src/ClientApp/src/models/delegateInvitation.ts
+++ b/Apps/WebClient/src/ClientApp/src/models/delegateInvitation.ts
@@ -1,0 +1,36 @@
+import { DataSource } from "@/constants/dataSource";
+
+/**
+ * Represents the delegate invitation model
+ */
+export interface DelegateInvitation {
+    /**
+     * The delegate invitation id.
+     */
+    id: string;
+
+    /**
+     * The friendly name of the delegate.
+     */
+    nickname: string;
+
+    /**
+     * The delegate invitation status.
+     */
+    status: string;
+
+    /**
+     * The email address to send the invitation.
+     */
+    email: string;
+
+    /**
+     * The expiry date of access.
+     */
+    expiryDate: string;
+
+    /**
+     * The list of data sources the delegate has access to.
+     */
+    dataSources: DataSource[];
+}
--- a/Apps/WebClient/src/ClientApp/src/models/sharing/createDelegateInvitationRequest.ts
+++ b/Apps/WebClient/src/ClientApp/src/models/sharing/createDelegateInvitationRequest.ts
@@ -1,14 +1,10 @@
 import { DataSource } from "@/constants/dataSource";
+import { DelegateInvitationStatus } from "@/constants/delegateInvitationStatus";
 
 /**
- * Represents the delegate invitation model
+ * Represents the request delegate invitation model.
  */
-export interface DelegateInvitation {
-    /**
-     * The delegate invitation id.
-     */
-    id?: string;
-
+export interface CreateDelegateInvitationRequest {
     /**
      * The friendly name of the delegate.
      */
@@ -16,15 +12,8 @@ export interface DelegateInvitation {
 
     /**
      * The delegate invitation status.
-     * Pending, Active, Declined, Locked, InviteExpired, AccessExpired
      */
-    status?:
-        | "Pending"
-        | "Active"
-        | "Declined"
-        | "Locked"
-        | "InviteExpired"
-        | "AccessExpired";
+    status?: DelegateInvitationStatus;
 
     /**
      * The email address to send the invitation.
--- a/Apps/WebClient/src/ClientApp/src/models/sharing/createDelegateInvitationRequest.ts
+++ b/Apps/WebClient/src/ClientApp/src/models/sharing/createDelegateInvitationRequest.ts
@@ -1,5 +1,4 @@
 import { DataSource } from "@/constants/dataSource";
-import { DelegateInvitationStatus } from "@/constants/delegateInvitationStatus";
 
 /**
  * Represents the request delegate invitation model.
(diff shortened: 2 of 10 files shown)


Apps/WebClient/src/ClientApp/src/models/delegateInvitation.ts
```diff
@@ -0,0 +1,43 @@
...
+     * The delegate invitation status.
+     * Pending, Active, Declined, Locked, InviteExpired, AccessExpired
+     */
+    status?:
+        | "Pending"
+        | "Active"
+        | "Declined"
+        | "Locked"
+        | "InviteExpired"
+        | "AccessExpired";
```
MikeLyttle, line 27: This should be a separate type - perhaps a string enum.

Apps/WebClient/src/ClientApp/src/services/restDelegateService.ts
```diff
@@ -0,0 +1,53 @@
...
+    private baseUri;
+
+    constructor(
+        logger: ILogger,
+        http: IHttpDelegate,
+        config: ExternalConfiguration
+    ) {
+        this.logger = logger;
+        this.http = http;
+        this.baseUri = config.serviceEndpoints["Delegate"];
```
MikeLyttle, line 24:
```suggestion
        this.baseUri = config.serviceEndpoints["GatewayApi"];
```

Apps/WebClient/src/ClientApp/src/models/delegateInvitation.ts
```diff
@@ -0,0 +1,43 @@
+import { DataSource } from "@/constants/dataSource";
+
+/**
+ * Represents the delegate invitation model
+ */
+export interface DelegateInvitation {
```
MikeLyttle, line 6: After chatting with Brian about the back-end implementation, we thought it might be best to split up the models so that CreateDelegateInvitationRequest will be different from the eventual UpdateDelegateInvitationRequest and returned DelegateInvitation.

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
```diff
@@ -0,0 +1,88 @@
...
+                errorType,
+                ErrorSourceType.Delegate,
+                incomingError.traceId
+            );
+        }
+    }
+
+    return {
+        createInvitation,
+        resetInvitation: resetNewInvitation,
```
MikeLyttle, line 84: I suspect one of these was supposed to be renamed.

BronzBierd: vsCode refactor, I'll correct

Apps/WebClient/src/ClientApp/src/models/sharing/createDelegateInvitationRequest.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface CreateDelegateInvitationRequest {
+    /**
+     * The friendly name of the delegate.
+     */
+    nickname?: string;
+
+    /**
+     * The delegate invitation status.
+     */
+    status?: DelegateInvitationStatus;
```
MikeLyttle, line 16: Status is not needed in the create model.

Apps/WebClient/src/ClientApp/src/models/sharing/delegateInvitation.ts
```diff
@@ -0,0 +1,33 @@
+import { DataSource } from "@/constants/dataSource";
+import { DelegateInvitationStatus } from "@/constants/delegateInvitationStatus";
+
+export interface DelegateInvitation {
+    /**
+     * The delegate invitation id.
+     */
+    id: string;
+    /**
```
MikeLyttle, line 9:
```suggestion
    id: string;
    
    /**
```

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
```diff
@@ -0,0 +1,125 @@
...
+    nickname?: string;
+    email?: string;
+    expiryDate?: string;
+    dataSources?: DataSource[];
+}
+
+const defaultInvitationDialogState: DelegateInvitationDialogState = {
+    mode: "Create",
+    step: DelegateInvitationDialogStep.contact,
+};
```
MikeLyttle, line 36: I think it would be good to consider where we keep track of the dialog being closed.  I'd suggest that occur in the store as well so the view doesn't have to send instructions to the dialog directly.

Then, if there's a default state, it should be closed (null or however we want to represent that).  The state can be populated as appropriate in the actions that open the dialog.  The state should also only need to be reset to the closed state internally within the store as a result of another action (closing or cancelling the dialog).

This could be handled when the dialog itself is implemented, though.

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
```diff
@@ -0,0 +1,125 @@
...
+                status: DelegateInvitationStatus.Pending,
+                nickname: invitationDialogState.value.nickname,
+                email: invitationDialogState.value.email,
+                expiryDate: invitationDialogState.value.expiryDate,
+                dataSources: invitationDialogState.value.dataSources,
+            });
+        }
+        // TODO: Remove this once all paths are covered
+        return Promise.resolve();
+    }
```
MikeLyttle, line 72: It will be easiest if each page of the dialog has a separate store action to submit (since they will have different parameters).  This one is set up nicely for the final "review" page, but the naming implies there is only one way to submit the dialog and there won't be other submit actions.

I'd also suggest we may want to throw an error for states that are unsupported instead of returning a successful promise.

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
```diff
@@ -0,0 +1,125 @@
...
+        return delegateService
+            .createInvitation(invitation)
+            .catch((error: ResultError) => {
+                handleError(error, ErrorType.Create, "sharing-dialog");
+            })
+            .then((result) => {
+                // TODO: retrieve all invitations, when invitation is successfully created
+                invitationsAreLoading.value = false;
+                resetInvitationDialogState();
+                return result;
```
MikeLyttle, line 88: Technically, the dialog is not done after an invitation is created, because there is an additional page where the sharing code is displayed.  Instead of resetting the state, I think we should move the sharingCode into DelegateInvitationDialogState, then the state can be updated so as to display the new page and the code won't need to be returned.

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
BronzBierd: Alright I'll go with the following approach (it may mature as I implement but for discussion sake):

```Typescript
dialogState = null -> dialog is closed (default)
function startNewInvitation => dialogState = { mode: "Create", step: DelegateInvitationDialogStep.contact }
function modifyInvitationExpiry(invite) = { mode: "Modify", step: DelegateInvitationDialogStep.expiry }
fucntion cancelNewInvitation => dialogState = null
```

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
BronzBierd: I've made some changes but any further changes at this point will be too much optimization before the implementation, I'll keep this in mind, the recent change already aligns to this so idea so it should follow through into the implementation.

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
BronzBierd: I'll continue to develop this as I implement the front end.

Apps/WebClient/src/ClientApp/src/stores/delegate.ts
```diff
@@ -0,0 +1,140 @@
...
+            mode: "Create",
+            step: DelegateInvitationDialogStep.contact,
+        };
+    }
+
+    function submitInvitationDialog(): Promise<string | void> {
+        if (
+            invitationDialogState.value != null &&
+            invitationDialogState.value.mode === "Create"
+        ) {
```
BronzBierd, line 77:
```suggestion
        if (invitationDialogState.value?.mode === "Create") {
```

MikeLyttle (approved):

BrianMaki (approved):

PR #5822, opened 2023-11-23
Apply C# 12 improvements in Admin projects AB#16500

MikeLyttle:
# Implements [AB#16500](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16500)

## Description

- migrate to collection expressions in Admin
- migrate to primary constructors for injected services in Admin
- fix controller documentation in Admin.Server

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## Items to Review:

### DIFF
--- a/Apps/Admin/Client/Components/Delegation/DependentTable.razor.cs
+++ b/Apps/Admin/Client/Components/Delegation/DependentTable.razor.cs
@@ -17,7 +17,6 @@ namespace HealthGateway.Admin.Client.Components.Delegation
 {
     using System;
     using System.Collections.Generic;
-    using System.Linq;
     using System.Threading.Tasks;
     using Fluxor;
     using Fluxor.Blazor.Web.Components;
@@ -51,7 +50,7 @@ namespace HealthGateway.Admin.Client.Components.Delegation
 
         private bool InEditMode => this.DelegationState.Value.InEditMode;
 
-        private IEnumerable<DependentRow> Rows => new[] { this.Data }.Select(d => new DependentRow(d));
+        private IEnumerable<DependentRow> Rows => [new DependentRow(this.Data)];
 
         private void SetEditMode(bool enabled)
         {
--- a/Apps/Admin/Client/Components/Delegation/DependentTable.razor.cs
+++ b/Apps/Admin/Client/Components/Delegation/DependentTable.razor.cs
@@ -51,7 +51,7 @@ namespace HealthGateway.Admin.Client.Components.Delegation
 
         private bool InEditMode => this.DelegationState.Value.InEditMode;
 
-        private IEnumerable<DependentRow> Rows => new List<DependentInfo> { this.Data }.Select(d => new DependentRow(d));
+        private IEnumerable<DependentRow> Rows => new[] { this.Data }.Select(d => new DependentRow(d));
 
         private void SetEditMode(bool enabled)
         {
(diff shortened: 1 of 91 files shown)


Apps/Admin/Client/Components/Delegation/DependentTable.razor.cs
```diff
@@ -51,7 +51,7 @@ public partial class DependentTable : FluxorComponent
 
         private bool InEditMode => this.DelegationState.Value.InEditMode;
 
-        private IEnumerable<DependentRow> Rows => new List<DependentInfo> { this.Data }.Select(d => new DependentRow(d));
+        private IEnumerable<DependentRow> Rows => new[] { this.Data }.Select(d => new DependentRow(d));
```
BronzBierd, line 54:
```suggestion
        private IEnumerable<DependentRow> Rows => [ new DependentRow(this.Data ) ];
```
Isn't this equivalent?

Apps/Admin/Client/Pages/DashboardPage.razor
```diff
@@ -123,7 +123,7 @@
                 <MudChart
                     data-testid="year-of-birth-chart"
                     ChartType="ChartType.Bar"
-                    ChartSeries="@(new List<ChartSeries> { GetYearOfBirthCountSeries() })"
+                    ChartSeries="[GetYearOfBirthCountSeries()]"
```
BronzBierd, line 126: Sorry don't have time to test this, but is it correct to drop the @ to ensure its executed as a C# expression?

if its fine please just close this.

BronzBierd (approved):
Wow its a lot to get follow, looks good, I made two comments not important from what I can see, they just looked out of place.

regarding the primary constructor it really looks like they've dropped true "injection" it looks more like sharing a global variable, weird.

Apps/Admin/Client/Pages/DashboardPage.razor
BronzBierd: I did see this in another template as well so I don't think my concerns are really an issue.

MikeLyttle: The only reason I dropped the @ was because the formatter was being weird and adding an extra space character at the start of the string when it was there.  I did test it out to be sure it would still work.

Apps/Admin/Client/Components/Delegation/DependentTable.razor.cs
MikeLyttle: Good catch.  I had tried a direct swap out of the List creation with the new collection expression - `[this.Data].Select(d => new DependentRow(d))` - but it wasn't able to interpret the type it should create when the Select() was involved.  I hadn't noticed that the Select() was largely unnecessary so I'll update it per your suggestion.

Apps/Admin/Common/Models/CovidSupport/VaccineDetails.cs
```diff
@@ -24,26 +24,6 @@ namespace HealthGateway.Admin.Common.Models.CovidSupport
...
-        /// <summary>
-        /// Initializes a new instance of the <see cref="VaccineDetails"/> class.
-        /// </summary>
-        /// <param name="doses">The list of doses.</param>
-        /// <param name="vaccineStatusResult">The patient's vaccine status.</param>
-        [JsonConstructor]
-        public VaccineDetails(IList<VaccineDose> doses, VaccineStatusResult vaccineStatusResult)
-        {
-            this.Doses = doses;
-            this.VaccineStatusResult = vaccineStatusResult;
```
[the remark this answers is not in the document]
MikeLyttle: No, but the constructor is not needed when all properties are settable.  Changing the type of the collections (to readonly collections or IEnumerables) allows us to make them settable without warnings appearing.

BrianMaki (approved):

PR #5913, opened 2023-12-22
Refactor WebClient code to throw JavaScript Errors for AB#16548

BronzBierd:
# Fixes [AB#16548](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16548)

## Description

1. Refactor Error interfaces into Error subclasses.
2. Remove ResultMessage property from the ResultError class. Moved all consumers of `resultMessage` property to the standard `message` property found on all Error classes and subclasses.
3. Updated RequestResultUtil to correctly map API based ResultError's to the new Error subclass in the front end.
4. Introduced a new ResultErrorDetails to house API resultError model.

## Testing

- [ ] Unit Tests Updated
- [X] Not Required

image

image

https://cloud.cypress.io/projects/ofnepc/runs/3483/overview?roarHideRunsWithDiffGroupsAndTags=1
Final commit fixes the three breaking 429 tests:
image

## UI Changes
NO Changes all behind the scenes changes.

## Notes
Address SonarCloud findings related to throwing JS Errors.

image

No longer has issue with the throwing of non-JS Errors.

## Items to Review:


BronzBierd:
image

after cleaning up the Regex with the recommended fixes this reduces the warnings to 1 which I'll handle on SonarCloud directly as the `// nosonar` comment hasn't seemed to work for me

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -28,6 +33,31 @@ export interface ResultError {
...
+        traceId: string = "",
+        statusCode?: number
+    ) {
+        super(resultMessage);
+        this.errorCode = errorCode;
+        this.traceId = traceId;
+        this.statusCode = statusCode;
+    }
+
+    public static fromApiResultError(apiResultError: ResultError): ResultError {
```
MikeLyttle, line 49: It's a little confusing to have apiResultError typed as `ResultError`.  I'd suggest introducing an `ApiResultError` interface and referencing that interface in the `RequestResult` model in place of `ResultError`.

We could also consider reserving the "Error" name for our JS Error classes and giving the back-end error models (and banner error models) a different moniker.

MikeLyttle (dismissed):

Apps/WebClient/src/ClientApp/src/models/errors.ts
BronzBierd: I like this suggestion, I'll do that it'll be much cleaner 👍 Not sure what other moniker would fit but I'll think of some synonyms or the likes.

BronzBierd: image

I like `ResultBooBoo`

MikeLyttle: You could also append a word, so you'd get something like ErrorDetails.

BrianMaki (dismissed):

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -12,14 +12,31 @@ export interface CustomBannerError {
...
+    constructor(message: string, statusCode?: number) {
+        super(message);
+        this.statusCode = statusCode;
+    }
+}
+
+// HG API binding interface for ErrorResult model
+export interface ResultErrorDetails {
+    // API's ResultMessage mapping property. This should be treated as private and the message should be mapped to and preferred.
+    resultMessage?: string;
```
MikeLyttle, line 28:
```suggestion
    // Message that will always be populated when ResultType is Error.
    resultMessage?: string;
```

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -12,14 +12,31 @@ export interface CustomBannerError {
...
+    // The error code associated with the request. Will always be populated when ResultType is Error.
+    errorCode: string;
```
MikeLyttle, line 30:
```suggestion
    // Code that will always be populated when ResultType is Error.
    errorCode: string;
```

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -12,14 +12,31 @@ export interface CustomBannerError {
...
+    // The HTTP status code returned by the request.
     statusCode?: number;
 }
 
-export interface ResultError {
-    // The result message associated with the request. Will always be populated when ResultType is Error.
-    resultMessage: string;
+export class ResultError extends Error {
     // The error code associated with the request. Will always be populated when ResultType is Error.
     errorCode: string;
```
MikeLyttle, line 41:
```suggestion
    // Code associated with the error.
    errorCode: string;
```

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -28,15 +45,36 @@ export interface ResultError {
     actionCode?: ActionType;
```
MikeLyttle, line 45: Suggest changing the documentation here to just "Action code." since the ResultError class is divorced from the RequestResult.

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -28,15 +45,36 @@ export interface ResultError {
...
+    ) {
+        super(resultMessage);
+        this.errorCode = errorCode;
+        this.traceId = traceId;
+        this.statusCode = statusCode;
+    }
+
+    public static fromResultErrorDetails(
+        apiResultError: ResultErrorDetails
+    ): ResultError {
```
MikeLyttle, line 63: Nit: name might flow better as `fromModel`, e.g. `throw ResultError.fromModel(result.resultError);`.

Apps/WebClient/src/ClientApp/src/models/errors.ts
```diff
@@ -12,14 +12,31 @@ export interface CustomBannerError {
...
+    statusCode?: number;
+
+    constructor(message: string, statusCode?: number) {
+        super(message);
+        this.statusCode = statusCode;
+    }
+}
+
+// HG API binding interface for ErrorResult model
+export interface ResultErrorDetails {
```
MikeLyttle, line 26: It might make sense to move this into `requestResult.ts`.

Apps/WebClient/src/ClientApp/src/models/errors.ts
BronzBierd: Left over from my previous implementation thanks for catching this.

MikeLyttle (approved):

BrianMaki (approved):

PR #5965, opened 2024-01-16
Ensure User Profile is saved or deferred properly for AB#16610

BrianMaki:
# Fixes [AB#16610](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16610)

## Description

- For async this.dbContext.UserProfile.Update(userProfile) needs to be explicit
- Move result.Payload = userProfile before commit to ensure Payload is populated if status is Deferred

Note: DbProfileDelegate will probably be refactored when implementing [AB#16606](https://dev.azure.com/qslvic/Health%20Gateway/_workitems/edit/16606)

Screenshot 2024-01-16 at 4 21 53 AM

## Testing

- [ ] Unit Tests Updated
- [x] Not Required

## UI Changes

No

## Items to Review:


BronzBierd (approved):
Judging by the following [Stack Thread](https://stackoverflow.com/a/73715118/3910655), I'm assuming that when the profile is returned by `GetuserProfileAsync`, the context tracking no longer has knowledge of the model and that's why the straight call to SaveChangesAsync did not save the record.

On that note, I didn't find any documentation that concretely stated you must update because of it being async, if you have a link to those docs I would appreciate reading about this phenomenon to concretely understand what was going.

That being said, I don't see any harm in accepting this as is as it will solve the problem.

BrianMaki:
> Judging by the following [Stack Thread](https://stackoverflow.com/a/73715118/3910655), I'm assuming that when the profile is returned by `GetuserProfileAsync`, the context tracking no longer has knowledge of the model and that's why the straight call to SaveChangesAsync did not save the record.
> 
> On that note, I didn't find any documentation that concretely stated you must update because of it being async, if you have a link to those docs I would appreciate reading about this phenomenon to concretely understand what was going.
> 
> That being said, I don't see any harm in accepting this as is as it will solve the problem.

My concern is when the scenario is Deferred i..e, commit is false, the save is not called immediately.  I wanted to make sure UserProfile was updated in the DbContext.  Also, since the UserProfile was being returned in the payload, I moved the assignment of UserProifle to Payload before the commit check, to ensure UserPayload was in the DbResult as per code before the async changes.

MikeLyttle (approved):

PR #6159, opened 2024-05-03
Fix sonar code smells for AB#16718

BrianMaki:
# Implements [AB#16718](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16718)

## Description

- Removed unused code for Beta Access
- Created constant for duplicated values for dark theme
- Created constant for duplicated values for light theme
- Removed not required code in admin server
- Added unit tests for string type, int type and object type to cache provider for GetAndSet and GetAndSetAsysnc calls; address issues here: https://sonarcloud.io/project/issues?resolved=false&id=bcgov_healthgateway&open=AY4bfpuwP_aouHfxgSg-
- Removed role 'img' from html files
- Removed role 'presentation' from vue files
- Replaced Controller with ControllerBase where needed as per sonar
- Added [Route("/")] to controller where needed as per sonar.  Note: for JobSchedulerController had to replace AuthorizationConstants.LogoutPath with"Logout" (minus prefix /) and AuthorizationConstants.LoginPath with "login" (minus prefix /).  If not, sonar would complain.

## Testing

- [x] Unit Tests Updated
- [] Not Required

## UI Changes

None

## Items to Review:


Apps/Admin/Server/Admin.Server.csproj
```diff
@@ -32,4 +32,8 @@
       <GlobalPropertiesToRemove></GlobalPropertiesToRemove>
     </ProjectReference>
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="Properties\ServiceDependencies\hgadmin - Web Deploy\profile.arm.json" />
+  </ItemGroup>
```
MikeLyttle, line 38: This change to the .csproj can be reverted.

Apps/Database/src/Assets/docs/EmailValidationTemplate.html
```diff
@@ -21,7 +21,7 @@
                     scope="col"
                     style="text-align: left; width:350px;"
                 >
-                    <div role="img" aria - label="Health Gateway Logo">
```
MikeLyttle, line 24: Attribute should be `aria-label`, not `arialabel`.

Apps/Common/test/unit/CacheProviders/DistributedCacheProviderTests.cs
```diff
@@ -237,5 +269,28 @@ private static string GenerateRandomString()
         {
             return Guid.NewGuid().ToString()[..5];
         }
+
+        private static Task<object?> GetGeneratedItemValue(CacheType cacheType)
```
MikeLyttle, line 273: This doesn't need to return a Task.  One of the tests that call this method puts the returned value inside a task: `Task.FromResult(value)` (line 243) while the other does not (line 146).  It would be simpler if this method returned just `object?` and the `Task.FromResult` was in both calling methods.

I'd also suggest updating the new tests' names to indicate they are testing the default value behaviour for cache misses.

Apps/Database/src/Assets/docs/EmailAccountClosed.html
```diff
@@ -21,7 +21,7 @@
                     scope="col"
                     style="text-align: left; width:350px;"
                 >
-                    <div role="img" aria-label="Health Gateway Logo">
+                    <div aria-label="Health Gateway Logo">
```
MikeLyttle, line 24: Changes to these .html files will show up on the next DB migration; do we want to create a migration now just to deal with them?

Apps/Database/src/Assets/docs/EmailValidationTemplate.html
BrianMaki: It is 'aria-label' isn't it?

Screenshot 2024-05-03 at 12 09 34 PM

Apps/Admin/Server/Admin.Server.csproj
BrianMaki: Reverted.

Apps/Database/src/Assets/docs/EmailValidationTemplate.html
MikeLyttle: Yes, you're right.  Somehow I missed seeing the corrected line.

Apps/Database/src/Assets/docs/EmailAccountClosed.html
BrianMaki: Migration created.

MikeLyttle (approved):

PR #6251, opened 2024-07-04
Add intercept and wait on admin endpoint calls for AB#16782

BrianMaki:
# Fixes [AB#16782](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16782)

## Description

To reduce the flakiness of some tests, implemented intercept and waits on endpoint calls.  This will ensure that the endpoints are finished before assertions are done on the UI.

Add intercept and wait to:

- agent access
- beta access
- communications
- delegation
- feedback review
- reports

Updated helpers to reduce the amount of code duplication:

- commands
- intercept
- support utilities

**Cypress Admin:** https://cloud.cypress.io/projects/rccf87/runs/2114/overview?roarHideRunsWithDiffGroupsAndTags=1

## Testing

- [ ] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:

### DIFF
--- a/Apps/Admin/Tests/Functional/cypress/support/functions/intercept.js
+++ b/Apps/Admin/Tests/Functional/cypress/support/functions/intercept.js
@@ -25,19 +25,19 @@ export function setupStandardAliases(page) {
 
 export function waitForInitialDataLoad(page) {
     switch (page) {
-        case "dashboard":
+        case "/dashboard":
             waitForDashboard();
             break;
-        case "communications":
+        case "/communications":
             waitForCommunication();
             break;
-        case "feedback":
+        case "/feedback":
             waitForFeedback();
             break;
-        case "beta-access":
+        case "/beta-access":
             waitForBetaAccess();
             break;
-        case "reports":
+        case "/reports":
             waitForReport();
             break;
         default:
@@ -69,7 +69,7 @@ function setupCommunicationAliases() {
     cy.log("Setting up communication aliases.");
 
     cy.intercept("GET", "**/Broadcast/").as("getBroadcast");
-    cy.intercept("GET", `**/Communication/*`).as("getCommunication");
+    cy.intercept("GET", `**/Communication/`).as("getCommunication");
 }
 
 function setupFeedbackAliases() {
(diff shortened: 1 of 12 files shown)


Apps/Admin/Tests/Functional/cypress/support/functions/intercept.js
```diff
@@ -0,0 +1,128 @@
...
+}
+
+function waitForDashboard() {
+    cy.log("Wait on dashboard");
+    cy.wait("@getAllTimeCounts", { timeout: defaultTimeout });
+    cy.wait("@getAppLoginCounts", { timeout: defaultTimeout });
+    cy.wait("@getRatingsSummary", { timeout: defaultTimeout });
+    cy.wait("@getAgeCounts", { timeout: defaultTimeout });
+    cy.wait("@getDailyUsageCounts", { timeout: defaultTimeout });
+    cy.wait("@getRecurringUserCount", { timeout: defaultTimeout });
```
MikeLyttle, line 104: I wanted to figure out exactly how the `wait()` commands work here.

Our front-end makes these requests in parallel, so it's extremely unlikely we would receive a response from the back-end for the first request before the front-end issues the second request (and so on).  My understanding is that, like other commands, each `wait()` bottlenecks the test so it won't proceed to the next command until the current one finishes.

The documentation states there are 2 different timeout values used by `wait()`:

- `requestTimeout`: Time, in milliseconds, to wait for a request to go out
- `responseTimeout`: Time, in milliseconds, to wait until a response

By passing a 60000ms value to the `timeout` property, we are overriding both of these.

I would expect the test to hit line 99, start a timer for 60000ms (the overridden `requestTimeout`), pause the test until the site issues a matching HTTP request for `@getAllTimeCounts`, then start another timer for 60000ms (the overridden `responseTimeout`), pausing until the server sends an HTTP response.  If the request or response do not happen in the given time frames, the test should fail.

However, we are not issuing our requests in between each of these `wait()` commands.  Therefore, if the tests are passing, the window for making the request must not start when the `wait()` is issued - it must start when the `intercept()` is issued.  I couldn't find this described in the documentation when I looked.  It may be worth playing around with timeouts to be absolutely certain of this.

image

Putting all that aside, if we are issuing requests simultaneously, I think it would make more sense to use a single `wait()`, passing in each of the aliases in an array.  That way, it ensures we wait a maximum of 60s for all of the responses instead of waiting 60s for the first, waiting another 60s for the second, etc.

BrianMaki: Changed to array.

Screenshot 2024-07-05 at 9 30 03 AM
Screenshot 2024-07-05 at 9 30 19 AM
Screenshot 2024-07-05 at 9 30 47 AM

MikeLyttle (approved):

PR #6285, opened 2024-08-01
Create UserValidationService and RegistrationService from UserProfileServiceV2 for AB#16811

BrianMaki:
# Implements [AB#16811](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/16811)

## Description

- Move CreateUserProfileAsync from UserProfileServiceV2 to RegistrationService
- Move IsPhoneNunberValidAsync from UserProfileServiceV2 to UserValidationService
- Move ValidateEligibilityAsync from UserProfileServiceV2 to UserValidationService
- Removed no longer applicable unit tests from UserProfileServiceV2Tests
- Created unit tests for RegistrationService
- Created unit tests for UserValidationService
- Updated UserProfileServiceV2 to use primary constructor
- Removed disabling of pragma warning for DI parameters in UserProfileServiceV2

**Unit Tests:**

Screenshot 2024-08-01 at 11 44 08 AM

**Cypress:**

Note: Organ Donor failures expected due to PHSA data issues.

https://cloud.cypress.io/projects/ofnepc/runs/4561/overview?roarHideRunsWithDiffGroupsAndTags=1

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:

### DIFF
--- a/Apps/GatewayApi/src/Services/RegistrationService.cs
+++ b/Apps/GatewayApi/src/Services/RegistrationService.cs
@@ -38,9 +38,6 @@ namespace HealthGateway.GatewayApi.Services
     using Microsoft.Extensions.Logging;
 
     /// <inheritdoc/>
-    /// <summary>
-    /// Initializes a new instance of the <see cref="RegistrationService"/> class.
-    /// </summary>
     /// <param name="configuration">The injected configuration.</param>
     /// <param name="logger">The injected logger.</param>
     /// <param name="authenticationDelegate">The injected authentication delegate.</param>
--- a/Apps/GatewayApi/src/Services/RegistrationService.cs
+++ b/Apps/GatewayApi/src/Services/RegistrationService.cs
@@ -1,0 +1,182 @@
+// -------------------------------------------------------------------------
+//  Copyright © 2019 Province of British Columbia
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// -------------------------------------------------------------------------
+namespace HealthGateway.GatewayApi.Services
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using FluentValidation;
+    using HealthGateway.Common.AccessManagement.Authentication;
+    using HealthGateway.Common.Constants;
+    using HealthGateway.Common.Data.Validations;
+    using HealthGateway.Common.Delegates;
+    using HealthGateway.Common.ErrorHandling.Exceptions;
+    using HealthGateway.Common.Messaging;
+    using HealthGateway.Common.Models;
+    using HealthGateway.Common.Models.Events;
+    using HealthGateway.Common.Services;
+    using HealthGateway.Database.Constants;
+    using HealthGateway.Database.Delegates;
+    using HealthGateway.Database.Models;
+    using HealthGateway.Database.Wrapper;
+    using HealthGateway.GatewayApi.Models;
+    using HealthGateway.GatewayApi.Validations;
+    using Microsoft.Extensions.Configuration;
+    using Microsoft.Extensions.Logging;
+
+    /// <inheritdoc/>
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RegistrationService"/> class.
+    /// </summary>
+    /// <param name="configuration">The injected configuration.</param>
+    /// <param name="logger">The injected logger.</param>
+    /// <param name="authenticationDelegate">The injected authentication delegate.</param>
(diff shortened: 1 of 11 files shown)


Apps/GatewayApi/src/Services/RegistrationService.cs
```diff
@@ -0,0 +1,182 @@
...
+    using HealthGateway.Database.Wrapper;
+    using HealthGateway.GatewayApi.Models;
+    using HealthGateway.GatewayApi.Validations;
+    using Microsoft.Extensions.Configuration;
+    using Microsoft.Extensions.Logging;
+
+    /// <inheritdoc/>
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RegistrationService"/> class.
+    /// </summary>
```
MikeLyttle, line 43:
```suggestion
```

Documentation for the constructor no longer needed.

Apps/GatewayApi/src/Services/UserProfileServiceV2.cs
```diff
@@ -20,126 +20,61 @@ namespace HealthGateway.GatewayApi.Services
...
     using HealthGateway.GatewayApi.Models;
-    using HealthGateway.GatewayApi.Validations;
     using Microsoft.Extensions.Configuration;
     using Microsoft.Extensions.Logging;
 
     /// <inheritdoc/>
-    public class UserProfileServiceV2 : IUserProfileServiceV2
+    /// <summary>
+    /// Initializes a new instance of the <see cref="UserProfileServiceV2"/> class.
+    /// </summary>
```
MikeLyttle, line 40:
```suggestion
```

Documentation for the constructor no longer needed.

Apps/GatewayApi/src/Services/UserValidationService.cs
```diff
@@ -0,0 +1,57 @@
...
+    using System.Threading.Tasks;
+    using HealthGateway.Common.Data.Validations;
+    using HealthGateway.GatewayApi.Models;
+    using HealthGateway.GatewayApi.Validations;
+    using Microsoft.Extensions.Configuration;
+
+    /// <inheritdoc/>
+    /// <summary>
+    /// Initializes a new instance of the <see cref="UserValidationService"/> class.
+    /// </summary>
```
MikeLyttle, line 29:
```suggestion
```

Documentation for the constructor no longer needed.

Apps/GatewayApi/test/unit/Services.Test/UserValidationServiceTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// </summary>
+        /// <param name="phoneNumber">The phone number to validate.</param>
+        /// <param name="expected">The expected value of the validation result.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [InlineData("3345678901", true)]
+        [InlineData("2507001000", true)]
+        [InlineData("xxx3277465", false)]
+        [InlineData("abc", false)]
+        [Theory]
+        public async Task PhoneNumberIsNotValid(string phoneNumber, bool expected)
```
MikeLyttle, line 48:
```suggestion
        /// IsPhoneNumberValidAsync.
        /// </summary>
        /// <param name="phoneNumber">The phone number to validate.</param>
        /// <param name="expected">The expected value of the validation result.</param>
        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
        [InlineData("3345678901", true)]
        [InlineData("2507001000", true)]
        [InlineData("xxx3277465", false)]
        [InlineData("abc", false)]
        [Theory]
        public async Task ShouldValidatePhoneNumber(string phoneNumber, bool expected)
```

Documentation is out of date, since the test also checks that it correctly identifies valid numbers.

Apps/GatewayApi/test/unit/Services.Test/UserValidationServiceTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// <param name="minPatientAge">The minimum patient age to validate against.</param>
+        /// <param name="patientAge">The patient age to validate.</param>
+        /// <param name="expected">The expected eligibility result.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous unit test.</returns>
+        [InlineData(0, 0, true)]
+        [InlineData(19, 19, true)]
+        [InlineData(19, 20, true)]
+        [InlineData(19, 18, false)]
+        [Theory]
+        public async Task ShouldValidateEligibilityAsync(int minPatientAge, int patientAge, bool expected)
```
MikeLyttle, line 72:
```suggestion
        public async Task ShouldValidateEligibility(int minPatientAge, int patientAge, bool expected)
```

Nit: I don't think we need to append "Async" to the names of methods in our unit tests, even though that's what we do in the rest of the project.  We don't really need to differentiate between async and non-async versions of a test, because we will only ever have one or the other.

Note that a unit test might be named with a reference to another method that has an Async suffix, and that wouldn't be extraneous because it helps us identify the specific method.

For example:

```
ValidateEligibilityAsyncShouldThrowException()
```
^ This is fine.

```
ValidateEligibilityAsyncShouldThrowExceptionAsync()
```
^ This is extra.

As a further example, with a method named `CreateProfileAsync`, a test could be named `ShouldCreateProfile` or `ShouldCreateProfileAsync` and be fine either way, though I would lean to the former as it describes expected behaviour.

MikeLyttle (approved):

PR #6657, opened 2026-04-01
Add BcCancerScreening email backfill job (DHSOHG-3405) AB#17044

BrianMaki:
# Implements [AB#17044](https://dev.azure.com/qslvic/304a1f8c-dace-4f85-adf3-bf563d5b3a39/_workitems/edit/17044)

## Description

**Summary**

Implements a Notification Backfill Job to pre-populate UserProfileNotificationSetting for eligible users without overriding existing user-defined preferences.

**Changes**

1. Database

- Added new table UserJobState
  - Tracks processed users per job using ProcessedHdid and JobName
  - Prevents reprocessing of the same user across job runs
  - Linked to UserProfile via UserProfileId (HDID)

2. Outbox

- Updated IOutboxStore accessibility to public
- Exposed DispatchOutboxItemsAsync to allow job-triggered dispatch after batch processing

3. NotificationBackfillJob

- Added new Hangfire job to backfill notification settings

**Behavior**

- Job execution controlled via configuration (Enabled)
- Processes users in batches (default: 5000)

**Selection Criteria**

- User must have a valid contact channel:
   - Email (default) or SMS (configurable via UseSmsChannel)
- User is eligible when:
   - No existing UserProfileNotificationSetting, OR
   - Existing setting has null values (i.e., not yet processed)
- Excludes users already processed (tracked via UserJobState)

**Processing Logic**

For each user:

- Insert new UserProfileNotificationSetting if none exists
- Update only null fields (does NOT override existing true/false values)
- Create corresponding UserJobState record
- Generate notification event and store in Outbox

Channel Value Assignment:

- EmailEnabled and SmsEnabled are set based on configuration:
   - If a value is provided (true or false) → it is applied
   - If a value is null → it is not set (remains unprocessed)
- Existing values are only updated when currently null
- Explicit user-defined values (true/false) are never overridden

**Transaction + Dispatch**

- Each batch is processed within a single transaction
- After commit:
   - Outbox events are dispatched asynchronously via Hangfire

**Scheduling**

- Configured to run every minute
- Supports delayed start via Delay

**Configuration**

### Configuration
```json
{
  "NotificationBackfill": {
    "Id": "NotificationBackfill-BcCancerScreening-Email",
    "Schedule": "* * * * *",
    "Immediate": "false",
    "Delay": 60,
    "Options": {
      "Enabled": true,
      "JobName": "NotificationBackfill-BcCancerScreening-Email",
      "NotificationType": "BcCancerScreening",
      "BatchSize": 5000,
      "UseSmsChannel": false,
      "EmailEnabled": true,
      "SmsEnabled": null
    }
  }
}
```

**Key Notes**

- Null = not processed, not false
- Existing user preferences are never overridden
- Changing JobName will result in all users being treated as unprocessed

**Job Scheduler:**

Screenshot 2026-04-01 at 11 00 19 AM

**SQL Explain Plan on main query:**

Screenshot 2026-04-01 at 2 12 43 PM

**SQL Explain Plan on secondary query:**

Screenshot 2026-04-01 at 2 35 55 PM

## Testing

- [x] Unit Tests Updated
- [ ] Not Required

## UI Changes

None

## Items to Review:


Apps/Database/src/Models/UserJobState.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// </summary>
+    [Index(nameof(ProcessedHdid), nameof(JobName), IsUnique = true)]
+    public class UserJobState : AuditableEntity
+    {
+        /// <summary>
+        /// Gets the generated identifier.
+        /// </summary>
+        [Key]
+        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
+        public int Id { get; init; }
```
sslaws: Don't we typically use UUIDs for identifiers?

Apps/Database/src/Models/UserJobState.cs
```diff
@@ -0,0 +1,62 @@
...
+        [Required]
+        [MaxLength(50)]
+        public string JobName { get; init; } = string.Empty;
+
+        /// <summary>
+        /// Gets the processed HDID for this run.
+        /// </summary>
+        [Column("UserProfileId")]
+        [MaxLength(52)]
+        public string ProcessedHdid { get; init; } = null!;
```
sslaws: I think this should probably be called UserProfileId or just hdid

Apps/JobScheduler/src/appsettings.json
```diff
@@ -196,6 +196,21 @@
         "UserCount": 100000,
         "BetaFeature": "Salesforce"
     },
+    "NotificationBackfill": {
+        "Id": "NotificationBackfill-BcCancerScreening-Email",
+        "Schedule": "* * * * *",
+        "Immediate": "false",
+        "Delay": 60,
```
sslaws, line 203: Where is this used?

sslaws (changes requested):

Apps/Database/src/Models/UserJobState.cs
BrianMaki: Changed.

Apps/JobScheduler/src/appsettings.json
BrianMaki: That is used in Job Scheduler Dashboard.  This can be any name i.e., NotificationBackfill-BcCancerScreening-Email.  Using same delay implementation as other jobs in configuration.

Apps/Database/src/Models/UserJobState.cs
BrianMaki: Changed C# property to Hdid to match how UserProfile has implemented the same column

sslaws: Yes, we have it defined in most places as 

        [MaxLength(52)]
        [Required]
        public string UserProfileId { get; set; } = null!;

or

        [Key]
        [Column("UserProfileId")]
        [MaxLength(52)]
        public string HdId { get; set; } = null!;

sslaws (approved):

Apps/Database/src/Models/UserJobState.cs
BrianMaki: But everywhere else it is Hdid.

    public class PatientModel
    {
        /// <summary>
        /// Gets or sets the health directed identifier.
        /// </summary>
        public string Hdid { get; set; } = string.Empty;

